PR #2, opened 2020-12-06
Vnfs refactor

gmkurtzer:
Lots of changes around VNFS refactoring and renaming to `container`, this also includes some kernel refactoring as well.


gmkurtzer:
I've been doing lots of testing with this, and it is working well. Going ahead and merging.

PR #3, opened 2020-12-08
Node discovery feature additions

gmkurtzer:
Basic implementation of node discovery by iPXE response service.

Warewulf will automatically search and find nodes that have configured IP addresses but no configured HW Addr. One potential additional option would be to make it so that nodes also have to have a `Discoverable` tag set which would probably be smart.


svdavidson (commented):
Hey Greg, Before I fully review I wanted to put this question/suggestion out there. Is there an option to automatically add the HW MAC address to the WW config?  Just wondering if it works (or could work) like this?
1. Admin puts WW into "auto-node-discovery" mode
2. Someone goes out and manually boots new "undiscovered" nodes
3. Nodes PXE boot and pick up an unassigned dynamic IP address from DHCP range
4. Nodes boot the discovery image which sends the HW MAC address directly to WW via gRPC or equivalent. (The BMC could also be configured at this time with an IP address)
5. Existing node entry with IP address in WW is updated with new HW MAC address -or- (possibly) new node entry is automatically created with the IP address and HW MAC address
6. WW reports wew nodes to the Admin via some notification mechanism (email, text, logs, ...)
7. Admin takes WW out of "auto-node-discovery" mode
8. Admin can review the updates and rearrange / reassign node attributes as needed
I'm not sure if this is how the typical large scale HPC site might do things, but this might be an easy way for small to medium HPC sites to do node discovery and populate their WW configuration with real node information.

etc/ipxe/unconfigured.ipxe
```diff
@@ -0,0 +1,12 @@
...
+
+echo
+echo ================================================================================
+echo Warewulf v4
+echo
+echo MESSAGE: This node is unconfigured. Please have your system administrator add a
+echo          configuration for this node with HW address: {{$.Hwaddr}}
+echo
+echo Rebooting in 1 minute...
+sleep 60
```
svdavidson, line 11: Will this create an infinite booting loop?  Or is there something limiting the number of boots?

gmkurtzer: Yep, the idea is that a system admin can take that HW Address and manually configure, and the node will reboot until it is configured.

gmkurtzer:
> Hey Greg, Before I fully review I wanted to put this question/suggestion out there. Is there an option to automatically add the HW MAC address to the WW config? Just wondering if it works (or could work) like this?

This will kinda do that at present.

> 
> 1. Admin puts WW into "auto-node-discovery" mode

The concept of auto-node discovery is kinda a misnomer here... That is because every node will boot and request a PXE configuration. When it makes that request, the Warewulf server will get the node's HW address and check the node configuration for that HW address.

If that HW address is unknown, it can easily look to see if there are any "discoverable" nodes, and assign that HW address to a given node.

So the idea "autodiscover" is really just a feature of the daemon process.

> 2. Someone goes out and manually boots new "undiscovered" nodes

check.

> 3. Nodes PXE boot and pick up an unassigned dynamic IP address from DHCP range

check.

> 4. Nodes boot the discovery image which sends the HW MAC address directly to WW via gRPC or equivalent. (The BMC could also be configured at this time with an IP address)

So in this case, the discovery image is literally just the iPXE boot binary as it makes a rest request.

I'd rather not have a different image as there is no need for it as all of the pieces are already present.

> 5. Existing node entry with IP address in WW is updated with new HW MAC address -or- (possibly) new node entry is automatically created with the IP address and HW MAC address

Yep, this happens via the Warewulf daemon in this PR.

> 6. WW reports wew nodes to the Admin via some notification mechanism (email, text, logs, ...)

we can change this behavior instead such that Warewulf server doesn't do the autoconfig itself, but instead logs the hwaddress to a file or a socket that wwctl can ingest and configure.

> 7. Admin takes WW out of "auto-node-discovery" mode

check.

> 8. Admin can review the updates and rearrange / reassign node attributes as needed

The more I'm thinking about this, the more I think we should do it like in #6.

>    I'm not sure if this is how the typical large scale HPC site might do things, but this might be an easy way for small to medium HPC sites to do node discovery and populate their WW configuration with real node information.

Most HPC centers will have a HW address manifest from the vendor or assign interns.

etc/ipxe/unconfigured.ipxe
svdavidson: Seems like notifying the admin in some way might be better than a reboot loop.  I'm imagining the case where there might be a bunch of nodes rebooting over and over for days. That might not be so good on the hardware.  If there was a list of these nodes and if the BMC could be configured with a temporary IP address, it might be easier for the admin to find the nodes, by doing an Ipmitool identify on them.

PR #4, opened 2020-12-14
IPMI Support Changes

svdavidson:
This PR is for basic changes to support IPMI based on testing with real BMC hardware.  Supported features include:
1. Calling IPMI configure script in background to avoid adding to boot time
2. Avoid excess calls to change flash by only updating the BMC if the configuration has changed
3. Don't configure BMC if no IPMI IP address is set
4. Log IPMI configure script to /var/log/ipmi.log


gmkurtzer (approved):
LGTM!

PR #5, opened 2020-12-19
Added parallel batching, IPMI power cycle, sensors, and console

svdavidson:
The batch stuff is pretty basic but does allow you to specify how many threads to run in parallel and for how long.  The idea is to issue say 50 IPMI poweron commands every 30 seconds.  I added a powercycle command a sensors command.  The IPMI sensors can be very slow on some servers.  I also added an IPMI console command.


gmkurtzer (approved):
This looks fantastic to me. Merging.

PR #6, opened 2020-12-19
Rework batch package

ikaneshiro:
Use a buffered channel to limit the number of active goroutines without needing
to wait for all active goroutines to finish


gmkurtzer (approved):
This looks good to me!

svdavidson (approved):
Looks good to me

PR #13, opened 2020-12-21
Init fixes

gmkurtzer:
Significant cleanups to init subsystem, added Root file system parameter (takes `initramfs` and `tmpfs` at present), changed default handling of KernelArgs, moved init template variables "upstream", etc...


svdavidson:
Looks good to me

PR #21, opened 2021-01-27
Overlays

kivel:
When SystemOverlay name is _not_ RuntimeOverlay name, the wrong system overlay was build for the node

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,8 @@ files: all
 	mkdir -p $(DESTDIR)/var/warewulf/overlays/system/default/warewulf/bin/
 	cp wwclient $(DESTDIR)/var/warewulf/overlays/system/default/warewulf/bin/
 	cp wwctl $(DESTDIR)/usr/bin/
+	mkdir -p $(DESTDIR)/usr/lib/systemd/system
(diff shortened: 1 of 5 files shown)


brianclemens:
Thanks!

kivel:
Hi Brian


There's more to come. Yesterday I found a bug in 'hosts' as well.


Do you prefer the small/tiny fix PRs, or like to have a mighty PR with many fixes?

I ... obviously ... like the first better than the latter. But, I'm also OK to go with the flow of your project.


BTW, I'm impressed by v4! I still need to learn a lot though, but it's less painful than v3 to learn 😉

If I may, I'd like to ask about a small section of the documentation to cover the boot process. So, basically what's happening from the PXE chain loader to 'init' and then into your image/container. I can't do this myself, as I'm more of a 'user'.


best

Niko

--
Paul Scherrer Institute
Dr. Niko Kivel
Forschungsstrasse 111
WBGB/017
5232 Villigen PSI
Switzerland

phone: +41 56 310 2226
email: niko.kivel@psi.ch


________________________________

brianclemens:
Small PRs are easier to review and merge quickly, so your current tactic works fine.

That's a great suggestion, I'll draft a page for that (docs are in https://github.com/ctrliq/warewulf-docs and changing rapidly)

Thanks again Niko!

PR #49, opened 2021-03-26
NIC type

kivel:
added feature to set the NIC type. So far the type was present, but couldn't be changed.


brianclemens:
Thanks!

PR #50, opened 2021-03-26
additional functions for templates

kivel:
I added two functions for in- and decrementing variables in templates.
Also added some very basic documentation with rudimentary examples.


brianclemens:
Very handy, thanks!

Issue #56, opened 2021-04-08
Add non-interactive support to all configuration commands to support automated build and testing


tcooper:
I'm using Vagrant to provision a small ww4 test cluster simply to learn the interface/features of warewulf.

My config steps follow an early version of the EL8 QuickStart Guide. I'm not sure if the behaviors have changed or I didn't notice in early provisioning attempts but `wwctl` commands to setup and/or modify profiles seem to prefer/require interactive response when making changes even when the `--force` option is used.

Here is a brief excerpt from my Vagrant provisioning script which generates the output below...

```
    14
    15	# Pull and build the VNFS container and kernel
    16	sudo wwctl container import docker://warewulf/centos-8 centos8 --setdefault
    17	sudo wwctl kernel import "$(uname -r)" --setdefault
    18
    19	# Setup the default node profile
    20	#sudo wwctl profile set default -K $(uname -r) -C centos-8
    21	sudo wwctl profile set --force default --netdev eth1 -M 255.255.255.0 -G 10.0.2.1
    22	sudo wwctl profile set --force default --netdev eth0 -M 255.255.255.0 -G 192.168.15.15
    23	sudo wwctl profile set --force default --netdev enp0s3 -M 255.255.255.0 -G 10.0.2.1
    24	sudo wwctl profile set --force default --netdev enp0s8 -M 255.255.255.0 -G 192.168.15.15
    25
    26	sudo wwctl profile list
    27
```

Here is a brief excerpt from the Vagrant provisioning log to demonstrate...

```
  1570      mgr: + sudo wwctl container import docker://warewulf/centos-8 centos8 --setdefault
  1571      mgr: Getting image source signatures
  1572      mgr: Copying blob sha256:33e4c05dc1b693c06eceb4dc9b5fb8484c27f9b099cf07be9fda16a07d75ffd4
  1573      mgr: Copying config sha256:f2997960e742caf41ef7be6e288b09b1f97a4308881d7df642ccea1d2e6c90dd
  1574      mgr: Writing manifest to image destination
  1575      mgr: Storing signatures
  1576      mgr: [LOG]       info unpack layer: sha256:33e4c05dc1b693c06eceb4dc9b5fb8484c27f9b099cf07be9fda16a07d75ffd4
  1577      mgr: Building container: centos8
  1578      mgr: Set default profile to container: centos8
  1579      mgr: [INFO]     Warewulf daemon process not running (/tmp/warewulfd.pid)
  1580      mgr: ++ uname -r
[...]

PR #64, opened 2021-04-16
Add --yes option to some profile and node commands and ensure warewulfd is reloaded

tcooper:
Adds `if SetYes == true` logic to `wwctl profile set...`, `wwctl profile delete...` and `wwctl node delete...` commands and ensures `warewulfd` is reloaded after `wwctl {node,profile} set...` commands to resolve issues reported in #56.

There are other `wwctl` commands which call `nodeDB.Persist()` and do not include `warewulfd.DaemonReload()`. Not sure if those should also be modified so I didn't touch them.


gmkurtzer:
You got to this before I did Trevor, thank you!

PR #74, opened 2021-04-25
When available, use "pigz" to speedup compression.

yqin:


gmkurtzer:
We should probably check for pigz before just attempting the command. Probably best to do a `which` function with a boolean return to test for the necessary dependencies, and then do the command.

gmkurtzer:
Merged and I slightly changed the logic to try to find the `pigz` binary and incorporate that into the generated command. Thanks Yong!

PR #75, opened 2021-04-28
Cluster name

kivel:
changed `GroupName` to `ClusterName`.
Now `wwctl node set -c <clustername>` can be used on overlays.


kivel:
uhh, this is weird.  Why are there so many differences in the PR?
is there a way to force yout repo into my fork? I don't care about my history.

tcooper:
@kivel ... are you [keeping your fork in sync](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) with upstream?

kivel:
yep, I do this all the time with others as well, but got a strange message about wrong timelines.
I guess the solution I just copy pasted from the internet didn't work that well.
I'll just dump my fork and fork again ... never had this before.

tcooper:
Maybe _somebody_ force pushed new history to upstream breaking the cardinal rule?

PR #80, opened 2021-05-10
custom Parameters

kivel:
Feature to add custom parameters to the node configuration and profiles.
It's intended to increase the flexibility of the config.

I allows for instance to import profile-specific files.
```
wwctl profile set -p dummy --value "/var/warewulf/opt/dummyImportProfile.txt" default
```
by using
```
{{Include (printf "%v" $.Params.dummy)}}
```
in a foobar.ww overlay-template, the `dummy` parameter is resolved and the respective file is imported.
As `dummy` is can be specified in the profile, a generic overlay can be used and files can still be managed by profiles.

### DIFF
--- a/internal/app/wwctl/node/set/root.go
+++ b/internal/app/wwctl/node/set/root.go
@@ -41,9 +41,9 @@ var (
 	SetDiscoverable   bool
 	SetUndiscoverable bool
 	SetRoot           string
-	SetParam          string
+	SetKey            string
 	SetValue          string
-	SetParamDel       bool
+	SetKeyDel         bool
 )
 
 func init() {
@@ -77,9 +77,10 @@ func init() {
 	baseCmd.PersistentFlags().BoolVar(&SetNetDevDel, "netdel", false, "Delete the node's network device")
 	baseCmd.PersistentFlags().BoolVar(&SetNetDevDefault, "netdefault", false, "Set this network to be default")
 
-	baseCmd.PersistentFlags().StringVarP(&SetParam, "param", "p", "", "Define custom parameter")
-	baseCmd.PersistentFlags().StringVarP(&SetValue, "value", "", "", "Set custom parameter value")
-	baseCmd.PersistentFlags().BoolVar(&SetParamDel, "paramdel", false, "Delete custom parameter")
+	baseCmd.PersistentFlags().StringVarP(&SetKey, "key", "k", "", "Define custom key")
+	baseCmd.PersistentFlags().BoolVar(&SetKeyDel, "keydel", false, "Delete custom key")
+
+	baseCmd.PersistentFlags().StringVarP(&SetValue, "value", "", "", "Set value")
 
 	baseCmd.PersistentFlags().BoolVarP(&SetNodeAll, "all", "a", false, "Set all nodes")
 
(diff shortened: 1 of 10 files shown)


internal/pkg/overlay/overlay.go
```diff
@@ -159,6 +161,11 @@ func buildOverlay(nodeList []node.NodeInfo, overlayType string) error {
 			t.NetDevs[devname].Type = netdev.Type.Get()
 			t.NetDevs[devname].Default = netdev.Default.GetB()
 		}
+		for paramname, param := range n.Params {
+			// var nd node.Params
```
kivel, line 165: argh, very bad job on my side cleaning up before committing, sorry.

internal/app/wwctl/node/set/root.go
```diff
@@ -74,6 +77,10 @@ func init() {
 	baseCmd.PersistentFlags().BoolVar(&SetNetDevDel, "netdel", false, "Delete the node's network device")
 	baseCmd.PersistentFlags().BoolVar(&SetNetDevDefault, "netdefault", false, "Set this network to be default")
 
+	baseCmd.PersistentFlags().StringVarP(&SetParam, "param", "p", "", "Define custom parameter")
```
gmkurtzer, line 80: I'd like to use the option name "key" instead of "param" here (`--key/-k` and `--keydel`). If you concur, will you make that minor mod?

internal/app/wwctl/profile/set/main.go
```diff
@@ -233,14 +233,43 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 				p.NetDevs[SetNetDev] = &nd
 			}
 
-			wwlog.Printf(wwlog.VERBOSE, "Node: %s:%s, Setting device as default\n", p.Id.Get(), SetNetDev)
+			wwlog.Printf(wwlog.VERBOSE, "Profile: %s:%s, Setting device as default\n", p.Id.Get(), SetNetDev)
```
gmkurtzer, line 236: Ooops, nice catch!

internal/app/wwctl/profile/set/root.go
```diff
@@ -63,6 +66,10 @@ func init() {
 	baseCmd.PersistentFlags().BoolVar(&SetNetDevDel, "netdel", false, "Delete the node's network device")
 	baseCmd.PersistentFlags().BoolVar(&SetNetDevDefault, "netdefault", false, "Set this network to be default")
 
+	baseCmd.PersistentFlags().StringVarP(&SetParam, "param", "p", "", "Define custom parameter")
```
gmkurtzer, line 69: Same thing here with regards to `--key/-k`.

gmkurtzer (changes requested):
Everything looks great. What is your take on param->key?

internal/app/wwctl/node/set/root.go
kivel: sure, no problem.
`param` was based on my limited creativity, I like `key` much better.
I'm also going to change the rest of the code from `Params` to `Keys`, otherwise it's confusing when reading the code.

internal/app/wwctl/profile/set/root.go
kivel: see `nod/set`

gmkurtzer (approved):

PR #81, opened 2021-05-14
Node cleanup

kivel:
There were a few inconstant parameters in `wwctl node {add, del}`
I removed `group` and `controller`, both didn't do anything. I added `clusterName` to `wwctl node add`.
Also, the printouts in debug-mode were broken, so I fixed them too.


gmkurtzer (approved):
This looks great, thank you Nikko!

PR #83, opened 2021-07-15
added ifcfg-eth0 for SUSE based system 

mslacken:
removed network.cc due to conflicting name as the ifcfg-eth0 for SUSE systems reside in `/etc/sysconfig/network/ifcg-eth0`
Is `network,ww` needed for centos/rocky systems? 
Or would it be better to have OS specific overlays?

### DIFF
--- a/overlays/system/default/etc/sysconfig/network.ww
+++ b/overlays/system/default/etc/sysconfig/network.ww
@@ -1,0 +1,2 @@
+NETWORKING=yes
+HOSTNAME={{$.Id}}
--- a/overlays/system/default/etc/sysconfig/network.ww
+++ b/overlays/system/default/etc/sysconfig/network.ww
@@ -1,2 +1,0 @@
-NETWORKING=yes
-HOSTNAME={{$.Id}}
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,9 @@ files: all
 	cp -r etc/dhcp $(DESTDIR)/etc/warewulf/
 	cp -r etc/ipxe $(DESTDIR)/etc/warewulf/
 	cp -r overlays $(DESTDIR)/var/warewulf/
+	# remove conflicting files from RedHat or SUSE distros
+	test -d /etc/sysconfig/network/ && rm $(DESTDIR)/var/warewulf/overlays//etc/sysconfig/network.ww \
+		|| rm -r $(DESTDIR)/var/warewulf/overlays//etc/sysconfig/network
 	chmod +x $(DESTDIR)/var/warewulf/overlays/system/default/init
 	chmod 600 $(DESTDIR)/var/warewulf/overlays/system/default/etc/ssh/ssh*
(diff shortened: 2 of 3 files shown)


overlays/system/default/etc/sysconfig/network.ww
```diff
@@ -1,2 +0,0 @@
-NETWORKING=yes
-HOSTNAME={{$.Id}}
```
gmkurtzer, line 2: RHEL systems require this. Can you add it back?

gmkurtzer (changes requested):

mslacken:
The problem is that SUSE distros need the directory `/etc/sysconifg/network` but RedHat systems the file `/etc/sysconfig/network`.
There can be several solutions, as
1. Try do configure SUSE systems via `interfaces.xml` which can be read by `wicked` the SUSE network config tool
2. Have conditionals in the .ww files
3. Install only the right files when running `make install`, which I the solution I choose

gmkurtzer:
When did SuSE start using `/etc/sysconfig/network` as a directory? I wasn't aware of that.

Yeah, this will cause a problem. I'm thinking through your suggestions and what the best way to do it would be. Not sure we can do this with a conditional template as best case scenario we would end up with an empty file there.

I was really hoping not to do this, but maybe the best option would be to create a separate overlay for each distro type (e.g. EL, SuSE, Debian, etc..). So we have our base default which is responsible for the base provisioning capabilities and then on top of that we will need to add the appropriate overlay for the distro we are booting. This will ensure that a single WW4 control system is agnostic to what node operating systems are being booted. It does however mean an extra step for configuring the provisioning overlays for nodes but that can be simplified within the default profile.

mslacken:
> When did SuSE start using `/etc/sysconfig/network` as a directory? I wasn't aware of that.
> 
I don't know, but I think for some years.
I have opened a bug against `wicked` which causes the problems https://bugzilla.suse.com/show_bug.cgi?id=1188442, perhaps this can be solved 'upstream'. I think at the moment the distribution aware `Makefile` is the best soltution.

gmkurtzer:
A distribution aware `Makefile` would make it that a system installed on RHEL wouldn't be able to provision SuSE nodes and vise versa. I will update the overlays such that there are distro specific overlays that someone can choose for any node.

mslacken:
Yes, having distro specfic overlays is a good idea, especially as there might be conflicts we do not see now. So will it possible in the future to have several system overlays in a profile or node?

gmkurtzer:
Yep, we can overlay overlays now. ;-)

I will create the separation of overlays today hopefully before the meeting and merge this in.

Thanks!

gmkurtzer:
I don't know what I was thinking,... obviously, pre-coffee. No, we can not overlay overlays, but we can overlay profiles.

Apologies for the misinformation and I don't like the alternative here which is to duplicate the entire system overlay for each distro type. Let's talk about this in the Warewulf meeting today.

mslacken:
Found another way, see pr#90

PR #85, opened 2021-07-16
Bash completion

mslacken:
Added bash-completion for wwctl


mslacken:
Man pages could also be created in an similar way. Should I implement this?

PR #86, opened 2021-07-21
Configureable LocalSateDir

mslacken:
`/var/warewulf` is not compatible with FHS, so make this directory configureable in `/etc/warewulf.conf`


mslacken:
Closing as I committed to much.

PR #87, opened 2021-07-21
LocalStateDir is configureable to meet FHS

mslacken:


gmkurtzer:
Thank you for starting this process. I'd really like to see this also applied as a `prefix`, `sysconfdir`, `bindir`, etc.. 

Because `sysconfdir` would be included in that, we can't rely on the configuration file to define these things, so it would probably need to be a build option or build configuration which statically defines that for the build.

Do you want to take a stab at that?

gmkurtzer:
We will come back to this for 4.2

gmkurtzer:
nevermind... 4.3 :)

jcsiadal:
Recommend replacing LOCALSTATE with something else, since this is a shared state directory. I also think a YAML key like "root share" or something similar would be more descriptive.

For the OpenHPC build, I patched the latest 4.2.0 using WWSHAREDIR as a variable name. I also created WWTFTPDIR and moved /var/tftpboot/warewulf to /srv/tftp/warewulf. I picked the Debian name only because that distro has already made /srv the default. While I did use variables, the directories are hardcoded by the patch.

PR #93, opened 2021-08-05
preserve ownership for overlays

mslacken:
added preservation of  uid and gid  in the overlays
When e.g. the `munge.key` is propagated via an overlay, it gets the wrong ownership (root instead of munge)
This pr adds the preservation of ownerships.

### DIFF
--- a/internal/pkg/overlay/overlay.go
+++ b/internal/pkg/overlay/overlay.go
@@ -11,7 +11,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-  "syscall"
 	"text/template"
 
 	"github.com/hpcng/warewulf/internal/pkg/config"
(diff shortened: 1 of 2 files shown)


internal/pkg/overlay/overlay.go
```diff
@@ -206,15 +207,44 @@ func buildOverlay(nodeList []node.NodeInfo, overlayType string) error {
 
 			if info.IsDir() {
 				wwlog.Printf(wwlog.DEBUG, "Found directory: %s\n", location)
-
-				err := os.MkdirAll(path.Join(tmpDir, location), info.Mode())
+        info, err := os.Stat(location)
```
gmkurtzer, line 210: Why not put this into a function instead of reused?

gmkurtzer (changes requested):

mslacken:
Closing this one, as github messed up forced push, retrying with a new #106

PR #94, opened 2021-08-16
Duplicate entries in passwd file

cmoore147:
A Mention in the documentation that this file needs to be modified for different setups should be added. By leaving the overlay this way the compute nodes will pull from two sources for users and have duplicate entries in passwd file.


gmkurtzer:
The second entry is actually by design and is required for proper function (the first is just for convenience and debugging). Luckily PAM allows for this and stops at the first match for a particular UID.

Thanks for the PR, but because this is by design, I'm going to close the PR without accepting. But we should absolutely fix the documentation for this so it is better understood why this exists as it does.

Thanks!

PR #98, opened 2021-08-21
Cleanups

gmkurtzer:
A WIP pull request to push cleanups too as they are found.


gmkurtzer:
@mslacken I see you had execute bit set for the following files:

```
overlays/system/default/etc/network/interfaces.d/eth0.ww
overlays/system/default/etc/systemd/network/10-persistent-net.link.ww
```

I've removed that in this PR, but wanted to confirm that this won't break SuSE network setup. If it does need to be `755`, let's do that in the Makefile instead of counting on Git's mode.

Thanks!

PR #101, opened 2021-08-24
Show IPMI Interface When Listing All Profiles

elguero:
We added the ability to configure an IPMI interface. We missed
displaying the configuration when listing all profiles.


gmkurtzer:
Ha, great catch!

PR #103, opened 2021-08-24
Add IPMI Port Configuration

elguero:
This commit adds the ability to configure the default IPMI port for a
profile or per node. If the port is not set, we default to 623.

Updated some areas to default to 'lan' for IPMI interface if it has not
been set.


gmkurtzer:
This looks great, thanks!

PR #106, opened 2021-08-31
added function CopyUIDGID for file file ownerships

mslacken:
file ownerships are preserved now in overlays

Cleaner implementation of #93


gmkurtzer (changes requested):
This looks really great.

My first observation is that there are a number of indention misalignments. Would you mind fixing the formatting of sources (`gofmt`) and updating this branch?

Thanks!

mslacken:
now with the correct format

internal/pkg/util/util.go
```diff
@@ -97,8 +102,16 @@ func CopyFiles(source string, dest string) error {
...
+			if err != nil {
+				return err
+			}
 
-			err := os.MkdirAll(path.Join(dest, location), info.Mode())
+			err = os.MkdirAll(path.Join(dest, location), info.Mode())
+			if err != nil {
+				return err
+			}
+			err = CopyUIDGID(source,dest)
```
gmkurtzer, line 114: Did you run this through `gofmt` or fix the formatting by hand?

While this is obviously not a big deal, `gofmt` would have fixed this so there was an appropriate space between `(source, dest)`.

gmkurtzer (approved):
For future PRs, please run all changes through `gofmt` just so we can have the same formatting across the project.

note: Never, in a million years, would I have thought I would become pedantic about space formatting... Oh, how people change. LOL

Issue #88, opened 2021-07-22
wwctl container file operation


mslacken:
There is no proper way to copy files inside a existing container. It can be achieved by copy files directly to `/var/warewulf/container/chroots/FOO/rootfs`, so this is not pending issue.
Perhaps this could be possible by having a bind mount option `wwctl container exec --bind BAAR FOO /bin/bash`

gmkurtzer:
Great idea. Do you want to take a stab at this, or shall I?

gmkurtzer:
I'm playing with this now. Will report back shortly.

gmkurtzer:
Done. Please test #97.

gmkurtzer:
Do we want to escape the colon in the bind path?

gmkurtzer:
This is done.

PR #109, opened 2021-09-02
Container exec bind escape

gmkurtzer:
This PR provides the ability to be smarter about splitting the `--bind` path around the `:` character to match and align with local file system paths.

note: It is not smart enough to catch the logic of:

```
# mkdir /tmp/foo /tmp/foo\:bar
# wwctl container exec --bind /tmp/foo\:bar:/mnt alpine /bin/sh
```

This is because the shell won't pass along the escape character to `wwctl` and as a result, it will only see the `:` separator and because `/tmp/foo` will exist, it will assume that is what you are binding. In that case, you must add an extra escape character so `wwctl` receives it as follows:

```
# wwctl container exec --bind /tmp/foo\\:bar:/mnt alpine /bin/sh
```

Reference to #88


internal/pkg/util/util.go
```diff
@@ -315,3 +315,45 @@ func CopyUIDGID(source string, dest string) error {
...
+
+	return (ret)
+}
+
+func SplitPaths(input, delim string) []string {
+	var ret []string
+	str := ""
+	for i := 1; i < len(input); i++ {
+		str += string(input[i-1])
+		if (string(input[i]) == delim && string(input[i-1]) != "\\") && (IsDir(str) || IsFile(str)) {
```
elguero, line 345: This is doing a little bit more than just string manipulation. It is validating the path.

Maybe rename the function "SplitValidPaths"?  Just a thought.

gmkurtzer: Done. Thanks!

elguero (approved):

PR #115, opened 2021-09-07
MS_REC added to MS_PRIVATE mount on root

gmkurtzer:
Hoping this fixes what @mslacken found on SLES Leap with regards to mounts not going away when the mount namespace shuts down.


mslacken:
Fixes the mount problem, thanks.

PR #116, opened 2021-09-08
Added container option for kernel import

mslacken:


gmkurtzer:
Fantastic feature addition! Looks good to me.

PR #120, opened 2021-09-09
allow compressed kernels with gz suffix

mslacken:
handy for aarch64

### DIFF
--- a/internal/pkg/kernel/kernel.go
+++ b/internal/pkg/kernel/kernel.go
@@ -6,12 +6,12 @@ import (
 	"os"
 	"os/exec"
 	"path"
-
-	"github.com/pkg/errors"
+	"strings"
 
 	"github.com/hpcng/warewulf/internal/pkg/config"
 	"github.com/hpcng/warewulf/internal/pkg/util"
 	"github.com/hpcng/warewulf/internal/pkg/wwlog"
+	"github.com/pkg/errors"
 )
 
 func ParentDir() string {
@@ -72,28 +72,30 @@ func ListKernels() ([]string, error) {
 }
 
 func Build(kernelVersion string, root string) (string, error) {
+
 	kernelImage := path.Join(root, "/boot/vmlinuz-"+kernelVersion)
 	kernelDrivers := path.Join(root, "/lib/modules/"+kernelVersion)
 	kernelDestination := KernelImage(kernelVersion)
 	driversDestination := KmodsImage(kernelVersion)
 
 	// Create the destination paths just in case it doesn't exist
-	err := os.MkdirAll(path.Dir(kernelDestination), 0755)
-	if err != nil {
-		return "", fmt.Errorf("failed to create kernal dest: %s", err)
-	}
-
-	err = os.MkdirAll(path.Dir(driversDestination), 0755)
-	if err != nil {
-		return "", fmt.Errorf("failed to create driver dest: %s", err)
-	}
+	os.MkdirAll(path.Dir(kernelDestination), 0755)
+	os.MkdirAll(path.Dir(driversDestination), 0755)
 
 	if !util.IsFile(kernelImage) {
-		return "", errors.New("Could not locate kernel image")
+		// allow uncompressed kernels (can be handy for aarch64)
+		kernelImage = path.Join(root, "/boot/vmlinux-"+kernelVersion)
+		if !util.IsFile(kernelImage) {
+			return "", errors.New("Could not locate kernel image")
+		}
 	}
 
 	if !util.IsDir(kernelDrivers) {
-		return "", errors.New("Could not locate kernel drivers")
+		// strip a gz suffix
+		kernelDrivers = strings.TrimRight(kernelDrivers, ".gz")
+		if !util.IsDir(kernelDrivers) {
+			return "", errors.New("Could not locate kernel drivers")
+		}
 	}
 
 	wwlog.Printf(wwlog.VERBOSE, "Setting up Kernel\n")
--- a/internal/pkg/kernel/kernel.go
+++ b/internal/pkg/kernel/kernel.go
@@ -83,7 +83,7 @@ func Build(kernelVersion string, root string) (string, error) {
 		return "", errors.Wrap(err, "failed to create kernal dest")
   }
 	if err := os.MkdirAll(path.Dir(driversDestination), 0755) ; err != nil {
-    return "", err
+		return "", errors.Wrap(err, "failed to create driver dest")
   }
 
 	if !util.IsFile(kernelImage) {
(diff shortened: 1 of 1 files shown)


gmkurtzer (changes requested):
Please fix linting errors.

mslacken:
Fixed linting error, although the linting must have existed before.
Still the handling of adding kernels is not optimal right now. I think it would good to have:
1. individual names for kernel, not just using their version as name, as you can have the same kernel version for different arches
2. resolving the module path from the kernel version is consequent and simple, but begs problems when the kernel have different naming schemes
So wouldn't it be better to have additional someting like
`wwctl kernel import -k $(REL_PATH_TO_KERNEL) -m $(REL_PATH_TO_MODULES) -n $(KERNEL_NAME)`

internal/pkg/kernel/kernel.go
```diff
@@ -72,28 +72,34 @@ func ListKernels() ([]string, error) {
...
-	if err != nil {
-		return "", fmt.Errorf("failed to create kernal dest: %s", err)
-	}
-
-	err = os.MkdirAll(path.Dir(driversDestination), 0755)
-	if err != nil {
-		return "", fmt.Errorf("failed to create driver dest: %s", err)
-	}
+  if  err := os.MkdirAll(path.Dir(kernelDestination), 0755) ; err != nil {
+    return "", err
```
WestleyK, line 83: I would suggest to keep the error wrapping. This will be helpful for the end-user, and debugging.

```suggestion
		return "", errors.Wrap(err, "failed to create kernal dest")
```

_`errors.Wrap()` is just like `fmt.Errorf()`, but `errors.Wrap()` also includes a stack trace._

internal/pkg/kernel/kernel.go
```diff
@@ -72,28 +72,34 @@ func ListKernels() ([]string, error) {
...
+  }
+	if err := os.MkdirAll(path.Dir(driversDestination), 0755) ; err != nil {
+    return "", err
```
WestleyK, line 86: Same here:

```suggestion
		return "", errors.Wrap(err, "failed to create driver dest")
```

internal/pkg/kernel/kernel.go
```diff
@@ -72,28 +72,34 @@ func ListKernels() ([]string, error) {
...
+  }
 
 	if !util.IsFile(kernelImage) {
-		return "", errors.New("Could not locate kernel image")
+		// allow uncompressed kernels (can be handy for aarch64)
+		kernelImage = path.Join(root, "/boot/vmlinux-"+kernelVersion)
+		if !util.IsFile(kernelImage) {
+			return "", errors.New("Could not locate kernel image")
```
WestleyK, line 93: This is okay for now, but for the future error message should not contain caps, or newlines.

WestleyK (changes requested):
Looks good, just a couple suggestions about error handling. Otherwise looks good.

gmkurtzer:
I just merged @WestleyK's lint timeout increase. Please merge with the main branch and pull that in which should get this past the lint CI.

Thanks!

gmkurtzer:
Are ARM kernels always suffixed with `.gz` or is that a SuSE thing?

mslacken:
> Are ARM kernels always suffixed with `.gz` or is that a SuSE thing?
Sorry don't know this.,

gmkurtzer:
Right after the copy of the kernel, let's also do an uncompress of the kernel.

mslacken:
I am feeling a bit uneasy implementing uncompressing the kernel without a rename, as at this would mean I would add the kernel with 
```
wwctl kernel import -r $FOO vmlinux-$VERSION.gz
```
but the in reality we would deliver the uncompressed kernel `vmlinux-$VERSION` but the kernel would be known to warewulf with '.gz' suffix, what is misleading.
So my idea right now is to change the `wwctl kernel import` in such a way, that it behaves similar to `wwctl container import` so that you can have an arbitrary name for the kernel. This would mean that the kernel can be imported with
```
wwctl kernel import -r $FOO vmlinux-$VERSION.gz $NAME
```
This solves the problem, as when the kernel is compressed and a $NAME is given and would have .gz suffix, it was the will of the user. If no $NAME is given and the kernel is compressed and has .gz suffix, the kernel would uncompressed and stored with NAME=$VERSION
This would also solve the problem, that at the moment it is not possible to have the same kernel version for two different arches.

rhenwood-arm:
@mslacken  - in my case, I like the idea of copying and rename to provide some sort of 'audatibility'. For example, I want to remind myself that the kernel I'm booting is the one I downloaded or fished out of an ISO - so I'll md5 the known good kernel and check what is in my directory that warewulf is using... so I manually gunzip the kernel to the name 'vmlinuz' and keep the known good long-named kernel beside it in the directory.

I would appreciate a quick example of how the command could be used to import two same kernels for different platforms?

Is it something like:
```
wwctl kernel import -r ~/pxe vmlinux-5.12.1-aarch64.gz suse_aarch64
wwctl kernel import -r ~/pxe vmlinux-5.12.1-x86_64.gz suse_x86_64
```
?


(and while I'm at it, I would also like something like: `wwwctl kernel import -f path_to_vmlinuz_file` :) )

gmkurtzer:
@mslacken:
If you type `uname -r` on your ARM SuSE system, what do you get? I ask because the command you reference here confuses me as I would have guessed it to be just `$VERSION`:

```
wwctl kernel import -r $FOO vmlinux-$VERSION.gz
```

@rhenwood-arm:
> (and while I'm at it, I would also like something like: wwwctl kernel import -f path_to_vmlinuz_file :) )

It is not trivial to get the kernel module version path from a kernel file directly.

mslacken:
As far I see the kernel and its modules are stored at the moment as `/var/warewulf/provision/kernel/$VERSION/kmods.img` and `vmlinuz`. 
I would suggest to use '/var/warewulf/provision/kenrl/$NAME/` as base directory and have an additional `version` file in this directory which stores the version. The version is anyway only needed, when 'building/imporing' the kernel, so storing it is not necessary, but still nice to have.

mslacken:
> @mslacken:
> If you type `uname -r` on your ARM SuSE system, what do you get? I ask because the command you reference here confuses me as I would have guessed it to be just `$VERSION`:
> 
On an ARM SUSE system 'uname -r` gives the version without the .gz suffix. In `/boot/` the lkernels have a `vmlinux` prefix and a `.gz` suffix.

gmkurtzer:
Will you check out #137, I think I got this all integated, including gzip decompression automatically for ARM kernels to support iPXE.

gmkurtzer:
I'm going to merge this because we have blockers on getting 4.2.0 out and this is holding other things up. If there are any bugs with this, let's fix those after this is merged.

gmkurtzer:
Apologies, I meant to merge the other version of the PR that includes the gunzip. Leaving this one closed for now.

Issue #121, opened 2021-09-09
log should have timestamps


mslacken:
The log file `/var/log/warewulfd.log` has no timestamps. This makes debugging more complicated than it could be.

gmkurtzer:
Please see PR #122.

gmkurtzer:
This has been merged.

PR #122, opened 2021-09-10
Adding timestamps to log output

gmkurtzer:
Adding timestamps per issue request #121.

### DIFF
--- a/internal/app/wwctl/server/status/main.go
+++ b/internal/app/wwctl/server/status/main.go
@@ -2,10 +2,9 @@ package status
 
 import (
 	"github.com/hpcng/warewulf/internal/pkg/warewulfd"
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 func CobraRunE(cmd *cobra.Command, args []string) error {
-	return errors.Wrap(warewulfd.DaemonStatus(), "failed to start daemon")
+	return warewulfd.DaemonStatus()
 }
--- a/internal/pkg/warewulfd/daemon.go
+++ b/internal/pkg/warewulfd/daemon.go
@@ -25,6 +25,10 @@ func DaemonStart() error {
 		}
 
 	} else {
+		if util.IsFile(WAREWULFD_PIDFILE) {
+			return errors.New("process is already running")
+		}
+
 		os.Setenv("WAREWULFD_BACKGROUND", "1")
 
 		f, err := os.OpenFile(WAREWULFD_LOGFILE, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
@@ -49,6 +53,8 @@ func DaemonStart() error {
 
 		fmt.Fprintf(p, "%d", pid)
 
+		fmt.Printf("Started Warewulf server at PID: %d\n", pid)
+
 	}
 
 	return nil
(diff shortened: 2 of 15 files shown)


internal/pkg/warewulfd/daemon.go
```diff
@@ -48,16 +49,14 @@ func DaemonStart() error {
...
-
 	}
 
 	return nil
 }
 
 func DaemonStatus() error {
 	if !util.IsFile(WAREWULFD_PIDFILE) {
-		wwlog.Printf(wwlog.INFO, "Warewulf daemon process not running (%s)\n", WAREWULFD_PIDFILE)
+		fmt.Printf("Warewulf daemon process not running (%s)\n", WAREWULFD_PIDFILE)
```
WestleyK, line 59: This should be returning the error. I'v actually already fixed this in #125.

internal/pkg/warewulfd/daemon.go
```diff
@@ -86,7 +85,7 @@ func DaemonStatus() error {
 
 func DaemonReload() error {
 	if !util.IsFile(WAREWULFD_PIDFILE) {
-		wwlog.Printf(wwlog.INFO, "Warewulf daemon process not running (%s)\n", WAREWULFD_PIDFILE)
+		fmt.Printf("Warewulf daemon process not running (%s)\n", WAREWULFD_PIDFILE)
 		return nil
```
WestleyK, line 89: Again, should return the error.

```suggestion
		return errors.New(fmt.Sprintf("warewulf daemon process is not runnin (%s)", WAREWULFD_PIDFILE))
```

internal/pkg/warewulfd/daemon.go
```diff
@@ -114,7 +113,7 @@ func DaemonReload() error {
 
 func DaemonStop() error {
 	if !util.IsFile(WAREWULFD_PIDFILE) {
-		wwlog.Printf(wwlog.INFO, "Warewulf daemon process not running (%s)\n", WAREWULFD_PIDFILE)
+		fmt.Printf("Warewulf daemon process not running (%s)\n", WAREWULFD_PIDFILE)
 		return nil
```
WestleyK, line 117: Here too:

```suggestion
		return errors.New(fmt.Sprintf("warewulf daemon process is not runnin (%s)", WAREWULFD_PIDFILE))
```

internal/pkg/warewulfd/warewulfd.go
```diff
@@ -21,30 +20,28 @@ func RunServer() error {
...
-				wwlog.Printf(wwlog.WARN, "Could not load database: %s\n", err)
+				fmt.Printf("ERROR: Could not load database: %s\n", err)
 			}
 		}
 	}()
 
 	err := LoadNodeDB()
 	if err != nil {
-		wwlog.Printf(wwlog.WARN, "Could not load database: %s\n", err)
+		fmt.Printf("ERROR: Could not load database: %s\n", err)
```
WestleyK, line 30: Why not return the error here?

internal/pkg/warewulfd/warewulfd.go
```diff
@@ -21,30 +20,28 @@ func RunServer() error {
...
 	http.HandleFunc("/overlay-runtime", RuntimeOverlaySend)
 
-	wwlog.Printf(wwlog.VERBOSE, "Starting HTTPD REST service\n")
+	daemonLogf("Starting HTTPD REST service\n")
 
 	err = http.ListenAndServe(":9873", nil)
 	if err != nil {
-		wwlog.Printf(wwlog.ERROR, "Could not start listening service: %s\n", err)
+		fmt.Printf("ERROR: Could not start listening service: %s\n", err)
 		os.Exit(1)
```
WestleyK, line 45: Avoid using `os.Exit()`. This should be replaced with: 

```suggestion
		return errors.Wrap(err, "could not start listening service")
```

WestleyK (changes requested):
Looks ok. Some suggestions about error handling.

internal/pkg/warewulfd/daemon.go
gmkurtzer: It's not an error though, just a status message.

Thoughts on how to best handle this?

gmkurtzer: Fixed in commit update

gmkurtzer:
@WestleyK I think I got all of the suggestions, plus a lot more in terms of how to properly be handling the error conditions and returns. Please check when you get a chance, thanks!

gmkurtzer:
I am merging this to resolve conflicts with Westley's other PRs.

internal/app/wwctl/server/status/main.go
```diff
@@ -2,10 +2,9 @@ package status
 
 import (
 	"github.com/hpcng/warewulf/internal/pkg/warewulfd"
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 func CobraRunE(cmd *cobra.Command, args []string) error {
-	return errors.Wrap(warewulfd.DaemonStatus(), "failed to start daemon")
+	return warewulfd.DaemonStatus()
```
WestleyK, line 9: Why did you remove the error wrap? Unless you are using predefined errors (which should be avoided), it is good practice to wrap every error we return.

internal/pkg/warewulfd/daemon.go
```diff
@@ -86,26 +88,22 @@ func DaemonStatus() error {
...
 	}
 
 	pid, _ := strconv.Atoi(string(dat))
 	process, err := os.FindProcess(pid)
-
 	if err != nil {
-		fmt.Printf("Failed to find process: %s\n", err)
-		return err
+		return errors.Wrap(err, "failed to find running PID")
 	} else {
```
WestleyK, line 103: No need for this `else` statement. _This `else` after return occurs several times here._

WestleyK (commented):
I know its already merged, but this can be fixed in another pr.

PR #128, opened 2021-09-13
Overlay UI updates

gmkurtzer:
These updates (still in testing) change how overlays are addressed by putting `system` or `runtime` as a mandatory option on the CLI.

Please test.

### DIFF
--- a/internal/app/wwctl/overlay/chown/main.go
+++ b/internal/app/wwctl/overlay/chown/main.go
@@ -35,7 +35,7 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 		os.Exit(1)
 	}
 
-	if len(args) > 3 {
+	if len(args) > 4 {
 		gid, err = strconv.Atoi(args[4])
 		if err != nil {
 			wwlog.Printf(wwlog.ERROR, "GID is not an integer: %s\n", args[4])
--- a/internal/app/wwctl/overlay/chown/main.go
+++ b/internal/app/wwctl/overlay/chown/main.go
@@ -11,36 +11,43 @@ import (
 	"github.com/hpcng/warewulf/internal/pkg/util"
 	"github.com/hpcng/warewulf/internal/pkg/wwlog"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 func CobraRunE(cmd *cobra.Command, args []string) error {
 	var overlaySourceDir string
-	overlayName := args[0]
-	fileName := args[1]
 	var uid int
 	var gid int
 	var err error
 
-	uid, err = strconv.Atoi(args[2])
+	overlayKind := args[0]
+	overlayName := args[1]
+	fileName := args[2]
+
+	if overlayKind != "system" && overlayKind != "runtime" {
+		return errors.New("overlay kind must be of type 'system' or 'runtime'")
+	}
+
+	uid, err = strconv.Atoi(args[3])
 	if err != nil {
-		wwlog.Printf(wwlog.ERROR, "UID is not an integer: %s\n", args[2])
+		wwlog.Printf(wwlog.ERROR, "UID is not an integer: %s\n", args[3])
 		os.Exit(1)
 	}
 
 	if len(args) > 3 {
-		gid, err = strconv.Atoi(args[3])
+		gid, err = strconv.Atoi(args[4])
 		if err != nil {
-			wwlog.Printf(wwlog.ERROR, "GID is not an integer: %s\n", args[3])
+			wwlog.Printf(wwlog.ERROR, "GID is not an integer: %s\n", args[4])
 			os.Exit(1)
 		}
 	} else {
 		gid = 0
 	}
 
-	if SystemOverlay {
+	if overlayKind == "system" {
 		overlaySourceDir = config.SystemOverlaySource(overlayName)
-	} else {
+	} else if overlayKind == "runtime" {
 		overlaySourceDir = config.RuntimeOverlaySource(overlayName)
 	}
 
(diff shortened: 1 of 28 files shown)


elguero (commented):
Found a minor bug.

internal/app/wwctl/overlay/chown/main.go
```diff
@@ -11,36 +11,43 @@ import (
...
+	}
+
+	uid, err = strconv.Atoi(args[3])
 	if err != nil {
-		wwlog.Printf(wwlog.ERROR, "UID is not an integer: %s\n", args[2])
+		wwlog.Printf(wwlog.ERROR, "UID is not an integer: %s\n", args[3])
 		os.Exit(1)
 	}
 
 	if len(args) > 3 {
```
elguero, line 38: Missed increasing the length here to 4.

elguero, line 38: Missed increasing the len of arguments to 4 here.

internal/app/wwctl/overlay/list/main.go
```diff
@@ -18,21 +19,32 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 	var o []string
 	var err error
 	var nodeList []node.NodeInfo
+	var overlayName string
+
+	overlayKind := args[0]
+
+	if len(args) > 1 {
+		overlayName = args[1]
+	}
```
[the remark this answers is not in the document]
gmkurtzer: I just pushed a minimum arg count for Cobra on `overlay list` which should cover this.

internal/app/wwctl/overlay/chown/main.go
gmkurtzer: Good catch, fixed!

internal/app/wwctl/overlay/delete/main.go
```diff
@@ -10,88 +10,94 @@ import (
...
+				return errors.Wrap(err, "failed deleting overlay")
 			}
 		}
 		fmt.Printf("Deleted overlay: %s\n", args[0])
 
-	} else if len(args) > 1 {
-		for i := 1; i < len(args); i++ {
-			removePath := path.Join(overlayPath, args[i])
+	} else {
+		removePath := path.Join(overlayPath, overlayPath)
```
elguero, line 63: I think we need `removePath := path.Join(overlayPath, fileName)` here.

I am getting this for a path:
```
[ERROR]           Path to remove doesn't exist in overlay: /var/warewulf/overlays/runtime/myOverlay1/var/warewulf/overlays/runtime/myOverlay1
```

elguero (changes requested):
Found a bug with path name as noted.

When deleting overlay files or directories, the node is not being updated. It works when adding a file but does not remove the file or directory.

When deleting an overlay, this warning might be a little misleading about skipping an overlay template not found.
```
$ sudo wwctl overlay delete runtime myOverlay1
Deleted overlay: runtime
[INFO]     Updating Runtime Overlays...
[WARN]     n0001.cluster                      : Skipped (runtime overlay template not found)
```


Is it expected to be able to set in a node(s) profile an overlay that does not exist?


Related to above question, upon deleting an overlay, the node(s) that have that overlay in their profile will still have it there?

```
NODE                 FIELD              PROFILE      VALUE
n0001.cluster        RuntimeOverlay     --           myOverlay1
n0001.cluster        SystemOverlay      --           (default)
```

internal/app/wwctl/overlay/imprt/main.go
```diff
@@ -9,39 +9,44 @@ import (
...
+		wwlog.Printf(wwlog.ERROR, "Overlay does not exist: %s:%s\n", overlayKind, overlayName)
 		os.Exit(1)
 	}
 
 	err := util.CopyFile(source, path.Join(overlaySource, dest))
 	if err != nil {
-		wwlog.Printf(wwlog.ERROR, "Failed copying file into overlay sourcedir:\n")
-		wwlog.Printf(wwlog.ERROR, "%s\n", err)
-		os.Exit(1)
+		return errors.Wrap(err, "could not copy file into overlay")
```
elguero, line 49: For user friendliness, if a user only provides a directory as a destination, could we take the base filename of the source and use that as the target filename?

Just tested this out:
```
if util.IsDir(path.Join(overlaySource, dest)) {
    dest = path.Join(dest, path.Base(source))
}
```
One other thing, if there is a file in the dest with the same name, should we warn the user about it and exit?
Just tested this below the code I suggested above.
```
 if util.IsFile(path.Join(overlaySource, dest)) {
    wwlog.Printf(wwlog.ERROR, "A file with that name already exists in the %s overlay %s\n:", overlayKind, overlayName)
    os.Exit(1)
}
```

internal/app/wwctl/overlay/delete/main.go
gmkurtzer: I concur. Fixing.

internal/app/wwctl/overlay/imprt/main.go
gmkurtzer: Import should indeed take the source path as the destination as this is "supposed" to do:
```
	if len(args) == 4 {
		dest = args[3]
	} else {
		dest = source
	}
```

Is your first suggestion a replacement to this or another conditional?

Regarding the second bit of code, totally agree. Will throw that in there and commit what I have.

elguero: That block takes care of when the user does not specify a destination at all.

Example:
`sudo wwctl overlay import runtime myOverlay1 /home/myoung/warewulf-4.0.0-1.el8.x86_64.rpm`

The resulting path passed to util.CopyFile() is, "/var/warewulf/overlays/runtime/myOverlay1/home/myoung/warewulf-4.0.0-1.el8.x86_64.rpm".

I am suggesting adding another conditional to handle when a user supplies only a directory name as the destination, without specifying the file name.

Example
`sudo wwctl overlay import runtime myOverlay1 /home/myoung/warewulf-4.0.0-1.el8.x86_64.rpm /testing09142021/`

Right now, util.CopyFile() errors out saying that the file could not be copied because 'testing09142021' is a directory.

The path.Join() on line 47, `err := util.CopyFile(source, path.Join(overlaySource, dest))`,  results in the path being "/var/warewulf/overlays/runtime/myOverlay1/testing09142021" which is indeed a directory.

My initial flow when I was testing was to use mkdir first and then import a file into that directory without having to type out or copy and paste the file name. If a user were to use the alias, cp, instead of import, it feels like a natural thing to do.

gmkurtzer: Ahhhhh, very good call, and thank you for the explanation!

PR #132, opened 2021-09-15
Adding a 'build' target to Makefile and other adjustments

elguero:
This target will allow for the a build to occur without calling the
vendor target.

Remove the .tools directory when running the 'clean' target.

Up the version number.


gmkurtzer:
After speaking with @ikaneshiro I went another way to fix this issue. Thanks!

PR #136, opened 2021-09-17
Fixes #135

### LINKED ISSUE #135: warewulfd port in config has no effect
the port setting in the config file does not have any effect on the daemon.

kivel:
in line 42 of warewulfd.go, the port number was hard coded.

### DIFF
--- a/internal/pkg/warewulfd/warewulfd.go
+++ b/internal/pkg/warewulfd/warewulfd.go
@@ -9,7 +9,7 @@ import (
 	"strconv"
 
 	"github.com/hpcng/warewulf/internal/pkg/warewulfconf"
-	"github.com/hpcng/warewulf/internal/pkg/wwlog"
+	"github.com/pkg/errors"
 )
 
 // TODO: https://github.com/danderson/netboot/blob/master/pixiecore/dhcp.go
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Hah, the issues of going from POC to production... Great catch, thanks!

internal/pkg/warewulfd/warewulfd.go
```diff
@@ -37,9 +41,22 @@ func RunServer() error {
 	http.HandleFunc("/overlay-system/", SystemOverlaySend)
 	http.HandleFunc("/overlay-runtime", RuntimeOverlaySend)
 
+	conf, err := warewulfconf.New()
+	if err != nil {
+		wwlog.Printf(wwlog.ERROR, "Could not get Warewulf configuration: %s\n", err)
+		os.Exit(1)
```
gmkurtzer, line 47: Instead of the wwlog error and then exit, can you do this instead:

`return errors.Wrap(err, "could not get Warewulf configuration")`

gmkurtzer (changes requested):

internal/pkg/warewulfd/warewulfd.go
kivel: sure, didn't do that on purpose. I didn't want to mix old and new error handling.
Might take till Monday, though.

gmkurtzer: Yeah, I'm on a push to update all error handling, every one helps!

gmkurtzer (approved):
Looks great, thanks!

Issue #119, opened 2021-09-09
arm support


mslacken:
I have tried to add boot an aarch64 node form x86 host running warewulf v4. For this I added an aarch64 container as an docker image. I also added the kernel from this container (see #120). After a restart of tftp did only boot to ipxe and showed following error

```
iPXE 1.21.1+ (g56f7d) -- Open Source Network Boot Firmware -- http://ipxe.org
Features: DNS HTTP iSCSI TFTP AoE EFI Menu

net0: 52:54:00:0f:75:a0 using SNP on SNP-0000:01:00.0 (Ethernet) [open]
  [Link:up, TX:0 TXE:1 RX:0 RXE:0]
  [TXE: 1 x "Network unreachable (http://ipxe.org/28086090)"]
Configuring (net0 52:54:00:0f:75:a0)...... ok
net0: 192.168.100.103/255.255.255.0
net0: fe80::5054:ff:fe0f:75a0/64
Next server: 192.168.100.250
Filename: http://192.168.100.250:9873/ipxe/52-54-00-0f-75-a0
http://192.168.100.250:9873/ipxe/52-54-00-0f-75-a0... ok
[Security] 3rd party image[0] can be loaded after EndOfDxe: .
InstallProtocolInterface: 5B1B31A1-9562-11D2-8E3F-00A0C969723B 43A5AEC40
Unloading driver at 0x00000000000
52-54-00-0f-75-a0 : 784 bytes [script]

================================================================================
Warewulf v4 now booting: arm01

Container: leap-15.3-arm-hpc:latest
Kernel: 5.3.18-59.19-default.gz
KernelArgs: crashkernel=no

http://192.168.100.250:9873/kernel/52-54-00-0f-75-a0... ok
[Security] 3rd party image[0] can be loaded after EndOfDxe: .
InstallProtocolInterface: 5B1B31A1-9562-11D2-8E3F-00A0C969723B 43A5AE1C0
Unloading driver at 0x00000000000
Could not select: Exec format error (http://ipxe.org/2e008081)

There was an error, rebooting in 15s...

```

Even this error is fixed, I would need a executable of `wwclient` for arch64. I can achieve this by creating a `default-aarch64` system profile and adding a aarch54 executable to this profile. Is this the best practice way to crossboot in warewulf, or should there be an architectur flag added?

mslacken:
I also see following messages in `/var/log/wareulfd.log`
```
[...]

PR #137, opened 2021-09-18
Arm kernel support

gmkurtzer:
An implementation of PR #120 for issue #119.

### DIFF
--- a/internal/pkg/kernel/kernel.go
+++ b/internal/pkg/kernel/kernel.go
@@ -1,7 +1,9 @@
 package kernel
 
 import (
+	"compress/gzip"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -80,16 +82,20 @@ func Build(kernelVersion string, root string) (string, error) {
 	// Create the destination paths just in case it doesn't exist
 	err := os.MkdirAll(path.Dir(kernelDestination), 0755)
 	if err != nil {
-		return "", fmt.Errorf("failed to create kernal dest: %s", err)
+		return "", errors.Wrap(err, "failed to create kernel dest")
 	}
 
 	err = os.MkdirAll(path.Dir(driversDestination), 0755)
 	if err != nil {
-		return "", fmt.Errorf("failed to create driver dest: %s", err)
+		return "", errors.Wrap(err, "failed to create driver dest")
 	}
 
 	if !util.IsFile(kernelImage) {
-		return "", errors.New("Could not locate kernel image")
+		if !util.IsFile(kernelImage + ".gz") {
+			return "", errors.New("Could not locate kernel image")
+		} else {
+			kernelImage = kernelImage + ".gz"
+		}
 	}
 
 	if !util.IsDir(kernelDrivers) {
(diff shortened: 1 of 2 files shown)


mslacken:
I could import the kernel and it got uncompressed, still I could not boot the it. This could also be a issue with my underlying virtual hardware. 
Still this patch keeps trying to search for a `vmlinuz` which is a x86_64 specific name.

rhenwood-arm:
@mslacken I had success with virtual hardware on AArch64... as well as a x86 virtual host running with QEMU. I unziped my kernel and named it `vmlinuz` for convenience.

I followed (something like) these instructions: https://warewulf.hpcng.org/docs/contributing/development-environment-kvm

What error are you seeing?

mslacken:
i always see the error 
```
Could not select: Exec format error (http://ipxe.org/2e008081)
```
As far as I see it from the logs, the kernel is delivered, see
```
[Tue Sep 21 16:08:06 CEST 2021] SEND:            arm01: /var/lib/warewulf/provision/kernel/5.3.18-59.19-default/vmlinuz
```
and it has the right format
```
/var/lib/warewulf/provision/kernel/5.3.18-59.19-default/vmlinuz: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), statically linked, BuildID[sha1]=7f1094f51de7ab74a2ee5096f88c7615a0dbf16b, not stripped
```

gmkurtzer:
@mslacken You mentioned that `vmlinuz` is an x86 specific name. My bad. What is the ARM specific name?

Can you copy your system kernel to a temporary name, decompress it, and see what is different between that and what Warewulf is dropping in `/var/lib/provision/kernel/*/vmlinuz`?

gmkurtzer:
> @mslacken I had success with virtual hardware on AArch64... as well as a x86 virtual host running with QEMU. I unziped my kernel and named it `vmlinuz` for convenience.

@rhenwood-arm I'd like to ensure that you don't have to uncompress or move your vmlinuz. It should be uncompressing the kernel automatically now. what kernel name are you seeing?

mslacken:
Still can't boot the arm kerel with following error
```
================================================================================
Warewulf v4 now booting: arm01

Container: leap15.3-arm
Kernel: 5.3.18-59.19-default
KernelArgs: crashkernel=no vga=791 rootfstype=rootfs

http://192.168.100.250:9873/kernel/52-54-00-0f-75-a0... ok
[Security] 3rd party image[0] can be loaded after EndOfDxe: .
InstallProtocolInterface: 5B1B31A1-9562-11D2-8E3F-00A0C969723B 43A5AB140
Unloading driver at 0x00000000000
Could not select: Exec format error (http://ipxe.org/2e008081)

There was an error, rebooting in 15s...
```

gmkurtzer:
@mslacken What is the differences between the kernel that Warewulf cached and t he system kernel (once gunzipped)?

mslacken:
@gmk I can't use the system kernel as the host running warewulf is on x86_64

gmkurtzer:
@mslacken There could be other factors going on then in a heterogenous environment.

How are you importing a kernel that isn't the same architecture as your host?

gmkurtzer:
I am merging *THIS* PR as it is blocking other things. If there are any bugs with this release, let's please patch and fix them as necessary. Thanks!

PR #139, opened 2021-09-20
handling kernels by name

mslacken:
version is also stored, kernel name defaults to its version

### DIFF
--- a/internal/app/wwctl/kernel/imprt/main.go
+++ b/internal/app/wwctl/kernel/imprt/main.go
@@ -17,9 +17,9 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 
 	// Checking if container flag was set, then overwriting OptRoot
 	kernelVersion := args[0]
-	kernelName := args[1]
-	if kernelName == "" {
-		kernelName = kernelVersion
+	kernelName := kernelVersion
+  if len(args) > 1 {
+		kernelName = args[1]
 	}
 	if OptContainer != "" {
 		if container.ValidSource(OptContainer) {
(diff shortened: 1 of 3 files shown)


internal/app/wwctl/kernel/imprt/main.go
```diff
@@ -14,58 +14,57 @@ import (
...
-			}
-		}
-		output, err := kernel.Build(arg, OptRoot)
-		if err != nil {
-			wwlog.Printf(wwlog.ERROR, "Failed building kernel: %s\n", err)
-			os.Exit(1)
+
+	// Checking if container flag was set, then overwriting OptRoot
+	kernelVersion := args[0]
+	kernelName := args[1]
```
gmkurtzer, line 20: This is going to cause a panic if args[1] is undefined.

gmkurtzer:
I started doing a review on this, but I think there are some areas that will cause some pretty major issues, and if I am understanding the intention behind this properly, I believe this can be done much simpler. 

Pushing this off until milestone 4.3.0.

internal/app/wwctl/kernel/imprt/main.go
mslacken: Should be fixed now.

mslacken:
I don't see were major issues can occur, as the kernels and their drivers are stored by name, which can with this patch set to an arbitrary name. If no name is given, the import falls back to the old behavior.

gmkurtzer:
Is the goal here just to be able to use an arbitrary kernel binary but still use the modules at `/var/modules/$(uname -r)`?

PR #140, opened 2021-09-20
Fix panic when failing to include the command to be run by the

elguero:
'container exec' command.

Update documentation for the 'container exec' command.


gmkurtzer:
Hah, great catch, thank you!

PR #142, opened 2021-09-21
resolves #141

### LINKED ISSUE #141: wwctl overlay list, generic error
### improvement

`wwctl overlay list`, creates a generic error
`ERROR: requires at least 1 arg(s), only received 0`

To improve user experience, a more meaningful error should be returned.
`ERROR: overlay kind must be specified. Use -h or --help for additional information.`

kivel:
nicer, more helpful, error message for `wwctl overlay list`

### DIFF
--- a/internal/app/wwctl/overlay/list/main.go
+++ b/internal/app/wwctl/overlay/list/main.go
@@ -21,14 +21,18 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 	var nodeList []node.NodeInfo
 	var overlayName string
 
-	overlayKind := args[0]
-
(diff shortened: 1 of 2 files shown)


gmkurtzer:
I agree but we should do this through all of the commands so the UI is consistent. Can I suggest we push this off to 4.3.0?

kivel:
sure, we can even postpone it to a later version. There is no rush, I just stumbled across this, because I was still used to the old behaviour of the command.
If I find the time, I'll check which other commands should be changed.

gmkurtzer:
This has been fixed and integrated.

PR #144, opened 2021-09-21
resolves #143

### LINKED ISSUE #143: kernels can only be set to default at import
### new feature
Kernels can only be defined as default during import.
Propose to add command
` wwctl kernel set --setdefault <kernel>`

kivel:
feature to set kernel as default added.


gmkurtzer:
there are a bunch of assumptions we have to make if we just allow `--setdefault`, and technically this can't "unset" it (so the help has to be tweaked a bit to match).

But it seems as though this is a node or profile configuration change, not a kernel configuration (as there are no configuration params for kernels).

kivel:
@gmkurtzer you're absolutely right. The issue of 'UNSET' only occurred to me later. 
Maybe it's better to close this PR all together.

PR #148, opened 2021-09-23
Make sure that ifcfg-XXXX files work properly

gmkurtzer:
A offering from Bill Comisky in Slack to ensure that nodes boot network predictably for EL8 varients.

### DIFF
--- a/containers/Singularity/rocky-8.def
+++ b/containers/Singularity/rocky-8.def
@@ -14,11 +14,8 @@ Include: dnf
     shadow-utils rsyslog tzdata util-linux words zlib tar less \
     gzip which util-linux openssh-clients openssh-server dhclient \
     pciutils vim-minimal shadow-utils strace cronie crontabs cpio \
-    wget rocky-release ipmitool yum NetworkManager network-scripts
+    wget rocky-release ipmitool yum network-scripts
 
-    systemctl disable NetworkManager
-    systemctl disable NetworkManager-dispatcher
-    systemctl disable NetworkManager-wait-online
     systemctl enable network-scripts
-    
+
     touch /etc/sysconfig/disable-deprecation-warnings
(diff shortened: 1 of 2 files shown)


containers/Singularity/rocky-8.def
```diff
@@ -14,4 +14,11 @@ Include: dnf
...
     gzip which util-linux openssh-clients openssh-server dhclient \
     pciutils vim-minimal shadow-utils strace cronie crontabs cpio \
-    wget rocky-release ipmitool yum NetworkManager
\ No newline at end of file
+    wget rocky-release ipmitool yum NetworkManager network-scripts
+
+    systemctl disable NetworkManager
+    systemctl disable NetworkManager-dispatcher
+    systemctl disable NetworkManager-wait-online
+    systemctl enable network-scripts
```
bcomisky, line 22: oops, I mistyped this originally, should be the following:
`systemctl enable network`

Not a native systemd service, this creates SysV style init script links.

containers/Singularity/rocky-8.def
```diff
@@ -14,4 +14,11 @@ Include: dnf
     shadow-utils rsyslog tzdata util-linux words zlib tar less \
     gzip which util-linux openssh-clients openssh-server dhclient \
     pciutils vim-minimal shadow-utils strace cronie crontabs cpio \
-    wget rocky-release ipmitool yum NetworkManager
\ No newline at end of file
+    wget rocky-release ipmitool yum NetworkManager network-scripts
```
bcomisky, line 17: Can we just leave off NetworkManager?  If so we can skip the disables, below too.

gmkurtzer: Yep, good thinking. I'll do that.

Thanks!

gmkurtzer:
@bcomisky How does that look?

containers/Singularity/rocky-8.def
bcomisky: Had some issues this morning with inconsistent eth0/eth1 device naming using this setup.  See warewulf slack thread.. I ended up adding back NetworkManager and excluding all devices except warewulf manged ones.. and still managed to have issues until I changed names from eth0 to ww0.   So not sure if I would've ended up in the same place using network-scripts.  Just not sure it's the way to go.

PR #149, opened 2021-09-24
Bash-completion dynamic nouns

mslacken:
Adds dynamic nouns, so that e.g.
```
wwctl kernel delete <TAB><TAB>
```
will give you a list of the installed kernels. Works now for various subcommands.

### DIFF
--- a/internal/app/wwctl/container/build/root.go
+++ b/internal/app/wwctl/container/build/root.go
@@ -1,6 +1,9 @@
 package build
 
-import "github.com/spf13/cobra"
+import (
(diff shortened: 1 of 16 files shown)


mslacken:
@gmkurtzer Checked most of the commands and they work as expected.
Note: This pr does not add additional syntax, but only completion

gmkurtzer:
Per request from @mslacken, I've added this to the 4.2.0 milestone.

gmkurtzer:
This is really slick!! Well done. I'm merging this now, but now I want all of the command working like this (e.g. `list`). LOL

PR #151, opened 2021-09-25
Fix sh syntax for equal operator

### LINKED ISSUE #150: Fail to boot node due to incorrect shell syntax
When trying to boot a debian container on a ww node, I get the following error:

```
Warewulf v4 is now booting: node1

Mounting up kernel file systems
Checking Rootfs type
/init: 32: test: initramfs: unexpected operator
/init: 36: test: initramfs: unexpected operator
ERROR: Unknown Warewulf Root file system: initramfs

There was a problem with the provisioning process, rebooting in 1 minute...
```

The problem is that the `init` script is using invalid syntax  in 3 places here:

https://github.com/hpcng/warewulf/blob/c097c15b1321a9dd58bde67c65cac789b8a30c58/overlays/system/default/init#L32-L37

`==` is invalid in `sh` but valid in `bash`.

This fix is to use `=` instead [since `=` is valid in both `sh` and `bash`](https://stackoverflow.com/questions/20449543/shell-equality-operators-eq).

robgjansen:
Fixes #150

### DIFF
--- a/overlays/system/default/init
+++ b/overlays/system/default/init
@@ -29,12 +29,12 @@ chmod 755 /warewulf/wwinit
 echo "Checking Rootfs type"
 ROOTFSTYPE=`stat -f -c "%T" /`
 
-if test "$WWROOT" == "initramfs"; then
+if test "$WWROOT" = "initramfs"; then
     echo "Provisioned to default initramfs file system: $ROOTFSTYPE"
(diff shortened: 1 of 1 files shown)


gmkurtzer (approved):
Hah, great catch, sorry about that!

PR #161, opened 2021-10-01
Update spec file and systemd unit file

elguero:
These commits do the following:

- Fix package reference for EL7
- Update spec file to use systemd macros
- Use macros to refer to file paths
- Update syntax

- Fix link to documentation in systemd unit file


gmkurtzer:
Much needed fixes, thank you!

PR #163, opened 2021-10-01
Initial refactor and cleanup of wwlog

gmkurtzer:
Rework `wwlog()` to use `fmt` instead of `log`.

### DIFF
--- a/internal/pkg/wwlog/wwlog.go
+++ b/internal/pkg/wwlog/wwlog.go
@@ -44,7 +44,7 @@ func prefixGen(level int) string {
 
 func printlog(level int, message string) {
 	if level == INFO && logLevel <= INFO {
-		fmt.Printf(message)
+		fmt.Print(message)
 	} else if level <= logLevel {
 		if level < INFO {
 			fmt.Fprintf(os.Stderr, prefixGen(level)+message)
--- a/internal/pkg/wwlog/wwlog.go
+++ b/internal/pkg/wwlog/wwlog.go
@@ -1,8 +1,8 @@
 package wwlog
 
 import (
-	"log"
-	"strings"
+	"fmt"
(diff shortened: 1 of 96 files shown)


internal/pkg/wwlog/wwlog.go
```diff
@@ -16,55 +16,39 @@ const (
...
+		return "[CRITICAL]"
 	}
+	return "[UNDEF]"
 }
 
 func Println(level int, message string) {
 	if level <= logLevel {
-		prefixLevel(level)
-		log.Println(message)
+		fmt.Fprintln(os.Stderr, prefixGen(level)+" "+message)
```
PaulCharlton, line 46: @gmkurtzer are you sure you want all logs going to stderr?

internal/pkg/wwlog/wwlog.go
```diff
@@ -16,55 +16,39 @@ const (
...
-
 	Printf(DEBUG, "Set log level to: %d\n", logLevel)
 }
 
-func SetIndent(i int) {
-	Indent = strings.Repeat(" ", i)
-}
-
-func prefixLevel(level int) {
+func prefixGen(level int) string {
```
PaulCharlton, line 27: @gmkurtzer assuming that this gets call A LOT.    a golang "switch" on "level" would be guaranteed faster and more maintainable code over time

internal/pkg/wwlog/wwlog.go
gmkurtzer: Good point, fixed.

internal/pkg/wwlog/wwlog.go
gmkurtzer: Fixed, thanks for the suggestion!

Issue #156, opened 2021-09-29
'kernel import --container' typically won't find a kernel


rhenwood-arm:
After installing a kernel in RockyLinux container `dnf install kernel-modules-extra` the kernel isn't in `/boot`. It is available in `/lib/modules/4.18.0-305.19.1.el8_4.aarch64/vmlinuz`. 

After installing a kernel in Alpine: `apk add linux-lts`, `/boot` is populated, but the kernel has the name: `vmlinuz-lts`.

I looks like the assumption that the kernel will be in `/boot` and is called `vmlinuz-$kernelVersion` is made here:
https://github.com/hpcng/warewulf/blob/a71c060b15a7f520d48d82b23dbe110d07840658/internal/pkg/kernel/kernel.go#L75

My first thought is that it might be good provide a way to override this assumption with arbitrary kernel location, in the case of a kernel import from a kernel?

gmkurtzer:
Did you also install `kernel-core` package? That is what installs the file in `/boot/vmlinuz*`.

rhenwood-arm:
yep: `kernel-modules-extra` pulls in `kernel-modules`, `kernel-core`, and `linux-firmware`.

gmkurtzer:
@rhenwood-arm That is very interesting... hrmm... What is the full path to the kernel and the modules directory for that kernel?

rhenwood-arm:
In the case of Rocky: `/lib/modules/4.18.0-305.19.1.el8_4.aarch64/vmlinuz`
and for Alpine: `/lib/modules/5.10.61-0-lts/`

gmkurtzer:
You mentioned there was a vmlinuz-lts named kernel?

rhenwood-arm:
yes: for the Alpine container example, installing `apk add linux-lts` gives me:

path to kernel: `/boot/vmlinuz-lts`
path to modules: `/lib/modules/5.10.61-0-lts/`

for Rocky:

path to kernel: `/lib/modules/4.18.0-305.19.1.el8_4.aarch64/vmlinuz`
path to modules: `/lib/modules/4.18.0-305.19.1.el8_4.aarch64/`

gmkurtzer:
ok, I wasn't expecting the path to kernel on rocky side to be there, but that's an easy enough thing to fail over to look for.

Regarding the Alpine location, that is interesting indeed. Does it also exist at `/lib/modules/$(uname -r)/vmlinuz`? If so, that would help a lot.

[...]

PR #164, opened 2021-10-02
Better logic to locate kernels

gmkurtzer:
(ref #156)

### DIFF
--- a/internal/pkg/kernel/kernel.go
+++ b/internal/pkg/kernel/kernel.go
@@ -16,6 +16,16 @@ import (
 	"github.com/hpcng/warewulf/internal/pkg/wwlog"
 )
 
+var (
+	kernelSearchPaths = []string{
+		// This is a printf format where the %s will be the kernel version
+		"/boot/vmlinuz-%s",
+		"/boot/vmlinuz-%s.gz",
+		"/lib/mmodules/%s/vmlinuz",
+		"/lib/mmodules/%s/vmlinuz.gz",
+	}
+)
+
 func ParentDir() string {
 	return path.Join(config.LocalStateDir, "provision/kernel")
 }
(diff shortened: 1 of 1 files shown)


internal/pkg/kernel/kernel.go
```diff
@@ -122,21 +122,25 @@ func Build(kernelVersion string, kernelName string, root string) (string, error)
 		return "", fmt.Errorf("failed to create version dest: %s", err)
 	}
 
-	if !util.IsFile(kernelImage) {
-		if !util.IsFile(kernelImage + ".gz") {
-			return "", errors.New("Could not locate kernel image")
-		} else {
-			kernelImage = kernelImage + ".gz"
-		}
+	if util.IsFile(path.Join(root, "/boot/vmlinuz-"+kernelVersion)) {
```
PaulCharlton, line 125: @gmkurtzer this would read cleaner and be more extensible to additional locations if the locations were placed in a slice of strings, then use the "for ... range" to iterate locations until found.    also, can use the variable "kernelDrivers" instead of retyping the path...

gmkurtzer: I got the slice integrated, but I don't get your second point about `kernelDrivers` and retyping the path.

Thanks!

internal/pkg/kernel/kernel.go
```diff
@@ -16,6 +16,16 @@ import (
 	"github.com/hpcng/warewulf/internal/pkg/wwlog"
 )
 
+var (
+	kernelSearchPaths = []string{
+		// This is a printf format where the %s will be the kernel version
+		"/boot/vmlinuz-%s",
+		"/boot/vmlinuz-%s.gz",
+		"/lib/mmodules/%s/vmlinuz",
```
PaulCharlton, line 24: @gmkurtzer should this be `/lib/modules` ?  (single `m`)

PR #165, opened 2021-10-04
Version

mslacken:
Added subcommand `wwctl version`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ debfiles: debian
 	cp wwclient $(DESTDIR)/var/warewulf/overlays/system/debian/warewulf/bin/
 
 wwctl:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,6 +1,10 @@
 .PHONY: all
-
 VERSION := 4.2.0
+VERSION_FULL ?= $(shell test -e .git && git describe --tags --long)
```
PaulCharlton, line 3: @mslacken is it possible that "git describe" should only have the "--first-parent" option set?  see: https://git-scm.com/docs/git-describe/2.33.0

mslacken:
Right `--first-parent` is sensible, I have updated the PR

gmkurtzer:
merge conflicts... let's blame @jcsiadal  instead of me. ;-)

mslacken:
Rebased it now, so that it can be merged

Issue #169, opened 2021-10-13
Remove "<>" from cobra documentation


jcsiadal:
See https://github.com/spf13/cobra/issues/1137.

Man conversion is generating lots of "WARNING: go-md2man does not handle node type HTMLSpan". The conversion is dropping text delimited by <>, so the final pages have missing options. For example:
      Use:   "chown [flags] <overlay kind> <overlay name> <path> <UID> [<GID>]",
   becomes in the man page
      wwctl overlay chown [flags]     []

I'll attach the master list. If we want to delimit user-text, we need something else.

jcsiadal:
Patched in PR #195 . Merged to main.

Issue #170, opened 2021-10-13
wwctl should generate default nodes.conf file in proper context


jcsiadal:
During RPM creation, wwctl is called by both man_page and bash_completion. By default, it will attempt to write a new nodes.conf file into /etc. As a normal user, this generates build errors, but as root it's modifying the filesystem (although you shouldn't generate RPMs as root).

My thoughts to fix this:
- Eliminate conf file generation when these functions are called.
- Have wwctl respect operation in alternate roots, such as when $DESTDIR is set (this would be preferred). Basically, add a command line option for root dir and then update Makefile to call wwctl with $DESTDIR as root.

jcsiadal:
I suspect the problem lies with app/wwctl/set/root.go. As I understand cobra, any packages referenced in an init() will have their init() called. If we want bash to include profile name completion, it makes sense to read nodes.conf in init(), but we need to avoid creating anything on the filesystem.

I think we should include the default nodes.conf during the build process and remove the automatic generation.

jcsiadal:
Fixed in PR #195 . Merged to main.

PR #173, opened 2021-10-15
Updates for generating RPMs

jcsiadal:
Fixes: #169 and #170.

This is an update with changes for cleanly generating RPM packages from source. I made changes to the spec.in package and now have a working OpenHPC RPM package in testing.

This is a summary of changes:
1) Updated cobra commander docs to remove '<>' around optional text. Updated some of the content and there's an obvious project here to align the WIki text and help text to match perfectly.
2) Updated help template. I used typical manpage formatting as the guide.
3) Removed automatic node config file generation. The default template is now included with "make install". I left the config generator as an unused (and untested) stub function in case we want to enable it later. I also changed it to use WriteFile.
4) Changed the Makefile so that manpages and bash_completion files are created during build instead of install. This keeps it better aligned to RPM structure.
5) Backported current work on the OpenHPC RPM to the warewulf RPM template.

### DIFF
--- a/internal/pkg/node/datastructure.go
+++ b/internal/pkg/node/datastructure.go
@@ -102,6 +102,7 @@ type NetDevEntry struct {
 	Gateway Entry `yaml:"gateway,omitempty"`
 }
 
+//nolint
 func newConfig() int {
 	//TODO: Add interface option to call this function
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Really great updates here @jcsiadal, well done!

PR #174, opened 2021-10-16
Bracket expansion for node ranges

gmkurtzer:
First feature add for bracket expansion to support node ranges like `n[0000-99].cluster[0-1]` to reference 100 nodes in two clusters (0 and 1).

This patch also provides the capability to increment IP addresses for new node creation:

```
$ sudo wwctl node add n00[00-99] --ipaddr 192.168.2.1 --netmask 255.255.255.0 --netdev eth0
```

### DIFF
--- a/internal/app/wwctl/node/set/main.go
+++ b/internal/app/wwctl/node/set/main.go
@@ -17,6 +17,7 @@ import (
 
 func CobraRunE(cmd *cobra.Command, args []string) error {
 	var err error
+	var count uint
 	var SetProfiles []string
 
 	nodeDB, err := node.New()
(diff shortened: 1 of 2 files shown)


gmkurtzer:
still need to add ability to increment IP addresses for `node set` command.

mslacken:
Great job, but I think the syntax you suggested is unclear, would following form not be better
```
$ sudo wwctl node add n00[00-99] --ipaddr 192.168.2.[001-100] --netmask 255.255.255.0 --netdev eth0
```

gmkurtzer:
That's a good point on the IP addr incrementing, but then there has to be alignment as well there to keep in alignment. The current way is that there is only a single source of count, and we don't have to maintain alignment.

Issue #176, opened 2021-10-21
Netdev Abstraction


scarycall:
I was looking to make my network config overlay work easier for multiple different network devices:

I would like to abstract the netdev name itself to make a single system overlay network configuration file.  Either add the netdev name as a key value in the node network device definition, or allow indexed list/array addressing of the $.NetDevs list:

This is a sample nodes network configuration.  We have different network devices depending on deployment and type.
cn01                 ens192:HWADDR      --           00:0c:29:0e:44:93
cn01                 ens192:IPADDR      --           192.168.0.51
cn01                 ens192:NETMASK     --           255.255.255.0
cn01                 ens192:GATEWAY     --           192.168.0.1
cn01                 ens192:TYPE        --           --
cn01                 ens192:DEFAULT     --           true

is there a way to do:
{{$.NetDevs.default.Ipaddr}}
Or 
{{$.NetDevs(0).Ipaddr}}
For use in the ifcfg-ww.ww network overlay:
  IPADDR={{$.NetDevs.default.Ipaddr}}

And,
{{$.NetDevs(0).netdev}}
or
{{$.NetDevs.vlan1.netdev}
For use in ifcfg-ww.ww network overlay:
  DEVICE="{{$.NetDevs.vlan1.netdev}"

Sample default system overlay file /etc/sysconfig/network-scripts/ifcfg-ww.ww as it exists before abstraction.

  TYPE="Ethernet"
  PROXY_METHOD="none"
  BROWSER_ONLY="no"
  BOOTPROTO="static"
  DEFROUTE="yes"
  IPV4_FAILURE_FATAL="no"
  IPV6INIT="no"
  IPV6_AUTOCONF="no"
  IPV6_DEFROUTE="no"
  IPV6_FAILURE_FATAL="no"
  IPV6_PRIVACY="no"
  NAME="ens192"
  DEVICE="ens192"
  ONBOOT="yes"
  IPADDR={{$.NetDevs.ens192.Ipaddr}}
  NETMASK={{$.NetDevs.ens192.Netmask}}
  GATEWAY={{$.NetDevs.ens192.Gateway}}
  HWADDR={{$.NetDevs.ens192.Hwaddr}}

gmkurtzer:
Hi @scarycall,

I just merged in #177, which I believes supports what you are asking for... There are now separate constructs for network configuration names and device names: `--netname` and `--netdevice` (note that `--netdev` is now an alias to `--netname`).

[...]

PR #177, opened 2021-10-23
Refactor netdev interfaces

gmkurtzer:
This is still a work in progress as a solution to #176.

The feature change separates out the network device from the network device name, so the name can be arbitary and unrelated to the network device itself (e.g. eth0).


PaulCharlton:
@gmkurtzer re - "files changed: 19";   seems like some refactoring could reduce the overall complexity of what on the surface appears to be a fairly straightforward change.

gmkurtzer:
> @gmkurtzer re - "files changed: 19"; seems like some refactoring could reduce the overall complexity of what on the surface appears to be a fairly straightforward change.

Yeah, there are a number of places that this will have an affect. I'm going to go through this again, but I don't think it is possible to simplify much more.

PR #180, opened 2021-10-25
Add node list capability for users

jcsiadal:

Assuming that wwctl remains locked down for users, this will provide the capability to warewaulfd to proxy pre-determined calls to wwctl. This adds the ability to print "node list" and "node ready" through HTTP. Users on any node could query using curl or wget. This feature could be easily expanded to include other unprivileged output.

I added a YAML element to turn off this feature. I used a system call instead of embedded functions to avoid the need for code duplication.

### DIFF
--- a/etc/warewulf.conf
+++ b/etc/warewulf.conf
@@ -3,6 +3,7 @@ netmask: 255.255.255.0
 warewulf:
   port: 9873
   secure: true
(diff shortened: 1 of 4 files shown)


jcsiadal:
Will be updating and resubmitting.

Issue #183, opened 2021-10-25
Extensions for handling NFS and fstab files


robgjansen:
I've noticed some issues with NFS handling:

- [ ] `wwctl configure nfs` does not honor the `enabled` option (it always overwrites /etc/exports)
- [ ] Can't specify options for nfs mounts
- [ ] Can't add exports on controller without also mounting them on nodes
- [ ] Can't easily add node-specific mounts (e.g. local hdds or ssds) to fstab

More rationale below.

---

Currently, we specify NFS exports in the `warewulf.conf` file like this:

```
exports:
  - /home
  - /var/warewulf
```

Because wwctl manages the `/etc/exports` file, it is difficult to add custom options when mounting these on the nodes, and it is difficult to export the shares without mounting them statically on the nodes (e.g., I want to use autofs instead of static mount for some of the shares).

---

The `/etc/fstab` file is managed by `wwctl configure nfs` but it doesn't offer a convenient way to specify local node-specific mounts. I wonder if we might want to write a `/etc/fstab.ww` template instead, and then expose a way for each node to add a mount string to an array that will then get added to the end of `/etc/fstab.ww`.

mslacken:
solved in #240 , please reopen if not all was implemented

PR #184, opened 2021-10-25
Some improvements for handling NFS mounts

robgjansen:
This fixes the bug where wwctl ignores the `enabled` option from `/etc/warewulf/warewulf.conf`, and also adds a new `NfsExtended` namespace to expose configuration of NFS mount options and static-fstab mounts.

Refs #183

### DIFF
--- a/internal/app/wwctl/configure/nfs/main.go
+++ b/internal/app/wwctl/configure/nfs/main.go
@@ -63,6 +63,13 @@ func Configure(show bool) error {
 				fmt.Fprintf(fstab, "%s:%s %s nfs defaults 0 0\n", controller.Ipaddr, export, export)
 			}
 
+			for _, export := range controller.Nfs.ExportsExtended {
+				fmt.Fprintf(exports, "%s %s/%s(%s)\n", export.Path, controller.Network, controller.Netmask, export.Options)
+				if export.Mount {
+					fmt.Fprintf(fstab, "%s:%s %s nfs defaults 0 0\n", controller.Ipaddr, export.Path, export.Path)
+				}
+			}
+
 			fmt.Printf("Enabling and restarting the NFS services\n")
 			if controller.Nfs.SystemdName == "" {
 				err := util.SystemdStart("nfs-server")
@@ -81,12 +88,17 @@ func Configure(show bool) error {
 		for _, export := range controller.Nfs.Exports {
 			fmt.Printf("%s %s/%s\n", export, controller.Network, controller.Netmask)
 		}
+		for _, export := range controller.Nfs.ExportsExtended {
+			fmt.Printf("%s %s/%s\n", export.Path, controller.Network, controller.Netmask)
+		}
 		fmt.Printf("\n")
 		fmt.Printf("SYSTEM OVERLAY: default/etc/fstab:\n")
 		for _, export := range controller.Nfs.Exports {
 			fmt.Printf("%s:%s %s nfs defaults 0 0\n", controller.Ipaddr, export, export)
 		}
-
+		for _, export := range controller.Nfs.ExportsExtended {
+			fmt.Printf("%s:%s %s nfs defaults 0 0\n", controller.Ipaddr, export.Path, export.Path)
+		}
 	}
 
 	return nil
--- a/internal/app/wwctl/configure/nfs/main.go
+++ b/internal/app/wwctl/configure/nfs/main.go
@@ -33,13 +33,6 @@ func Configure(show bool) error {
 	}
 
 	if !SetShow {
-		exports, err := os.OpenFile("/etc/exports", os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
-		if err != nil {
-			wwlog.Printf(wwlog.ERROR, "%s\n", err)
-			os.Exit(1)
-		}
-		defer exports.Close()
-
 		fstab, err := os.OpenFile("/var/warewulf/overlays/system/default/etc/fstab", os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
 		if err != nil {
 			wwlog.Printf(wwlog.ERROR, "%s\n", err)
@@ -47,7 +40,6 @@ func Configure(show bool) error {
 		}
 		defer fstab.Close()
 
-		fmt.Fprintf(exports, "# This file was written by Warewulf (wwctl configure nfs)\n")
 		fmt.Fprintf(fstab, "# This file was written by Warewulf (wwctl configure nfs)\n")
 
 		fmt.Fprintf(fstab, "rootfs / tmpfs defaults 0 0\n")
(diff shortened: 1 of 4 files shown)


internal/app/wwctl/configure/nfs/main.go
```diff
@@ -56,35 +48,57 @@ func Configure(show bool) error {
...
 
-		for _, export := range controller.Nfs.Exports {
-			fmt.Fprintf(exports, "%s %s/%s(sync)\n", export, controller.Network, controller.Netmask)
-			fmt.Fprintf(fstab, "%s:%s %s nfs defaults 0 0\n", controller.Ipaddr, export, export)
-		}
-
-		fmt.Printf("Enabling and restarting the NFS services\n")
-		if controller.Nfs.SystemdName == "" {
-			err := util.SystemdStart("nfs-server")
+		if controller.Nfs.Enabled {
```
PaulCharlton, line 51: what happens when then key "Nfs.Enabled" does not exist?
are there unit tests for any of this logic?

robgjansen: I think we probably want `Nfs.Enabled` to default to true in the case that it is not specified in the config file.

I'm not familiar enough with the code to understand how/where to properly test this. Could you provide some pointers? I've also made the PR editable by the maintainers in case it's more desirable to push a test than to explain how to implement it.

Note that I don't think the logic was tested before, or we would have caught the case where setting `Nfs.Enabled` to false had no effect.

gmkurtzer: Unit tests??? ;-)

panda1100: I'm testing this code now and I got `line 25: cannot unmarshal !!map into []*warewulfconf.NfsExportConf` error with config bellow. How should I write a config??

```
 19 nfs:
 20   systemd name: nfs-server
 21   exports:
 22   - /home
 23   - /var/warewulf
 24   exports extended:
 25     path: /opt
 26     options: rw
 27     mount: true
``

Comment (panda1100) [reply internal/app/wwctl/configure/nfs/main.go:51]: Sorry, my bad. The following works:

```
 19 nfs:
 20   enabled: true
 21   systemd name: nfs-server
 22   exports:
 23   - /home
 24   - /var/warewulf
 25   exports extended:
 26     - path: /opt
 27       options:
 28         - rw
 29         - sync
 30       mount: true
```

Comment (panda1100) [reply internal/app/wwctl/configure/nfs/main.go:51]: Sorry, my bad. Above error disappear with the following config.
but now I have error in `/etc/exports`.
Options(string array) are convert to `[rw sync]` this form.

```
 19 nfs:
 20   enabled: true
 21   systemd name: nfs-server
 22   exports:
 23   - /home
 24   - /var/warewulf
 25   exports extended:
 26     - path: /opt
 27       options:
 28         - rw
 29         - sync
 30       mount: true
 ```
 
 ```
# This file was written by Warewulf (wwctl configure nfs)
/home 10.0.0.0/255.255.255.0(sync)
/var/warewulf 10.0.0.0/255.255.255.0(sync)
/opt 10.0.0.0/255.255.255.0([rw sync])
 ```

gmkurtzer: We should be able to handle the default being enabled in the YAML unmarshal parsing. In theory, I think this should work, but I didn't test to confirm:

```
	Enabled     bool     `yaml:"enabled"  default:"true"`
```

panda1100: It looks like `gopkg.in/yaml.v2` doesn't support **default value**. --> https://github.com/go-yaml/yaml/issues/165
With this module, https://github.com/creasty/defaults , easy to support **default value**.

The following is git diff from this PR:

```
diff --git a/internal/pkg/warewulfconf/datastructure.go b/internal/pkg/warewulfconf/datastructure.go
index 620665f..0cfac05 100644
--- a/internal/pkg/warewulfconf/datastructure.go
+++ b/internal/pkg/warewulfconf/datastructure.go
@@ -3,6 +3,7 @@ package warewulfconf
 import (
        "github.com/hpcng/warewulf/internal/pkg/util"
+       "github.com/creasty/defaults"
 )
 
 const ConfigFile = "/etc/warewulf/warewulf.conf"
@@ -43,12 +44,23 @@ type TftpConf struct {
 }
 
 type NfsConf struct {
-       Enabled     bool     `yaml:"enabled"`
+       Enabled     bool     `default: "true" yaml:"enabled"`
        Exports     []string `yaml:"exports"`
        ExportsExtended []*NfsExportConf `yaml:"exports extended"`
        SystemdName string   `yaml:"systemd name"`
 }
 
+func (s *NfsConf) UnmarshalYAML(unmarshal func(interface{}) error) error {
+    defaults.Set(s)
+
+    type plain NfsConf
+    if err := unmarshal((*plain)(s)); err != nil {
+        return err
+    }
+
+    return nil
+}
+
 type NfsExportConf struct {
        Path     string  `yaml:"path"`
        Options  string  `yaml:"options"`
```

and configuration file looks like this

```
nfs:
  enabled: true
  systemd name: nfs-server
  exports:
  - /var/warewulf
  exports extended:
    - path: /home
      options: rw,sync
      mount: true
    - path: /opt
      options: sync
      mount: false
```

and I set to nfs enabled to false, and execute `sudo wwctl configure nfs -d -v`,
then  I get following debug print produced by following code:

```
2021/11/08 06:41:41.604538 [ERROR] 	  NFS ENABLED OPTION: false
```

```
diff --git a/internal/app/wwctl/configure/nfs/main.go b/internal/app/wwctl/configure/nfs/main.go
index b9d5ec0..cb98917 100644
--- a/internal/app/wwctl/configure/nfs/main.go
+++ b/internal/app/wwctl/configure/nfs/main.go
@@ -47,7 +47,7 @@ func Configure(show bool) error {
                fmt.Fprintf(fstab, "tmpfs /run/shm tmpfs defaults 0 0\n")
                fmt.Fprintf(fstab, "sysfs /sys sysfs defaults 0 0\n")
                fmt.Fprintf(fstab, "proc /proc proc defaults 0 0\n")
-
+               wwlog.Printf(wwlog.ERROR, "NFS ENABLED OPTION: %t\n",controller.Nfs.Enabled)
                if controller.Nfs.Enabled {
                        exports, err := os.OpenFile("/etc/exports", os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
                        if err != nil {
```

robgjansen: What is printed when you _remove_ the `enabled` key from the config file, i.e., did you verify that the default value is true when not explicitly set?

panda1100: @robgjansen
Yes, I verified.
If I remove `enabled` key from config file, this PR sets `true`(default value) to `enabled`.

```
nfs:
  systemd name: nfs-server
  exports:
  - /var/warewulf
```

panda1100:
Apologize for lint error.
I,ll send you a PR again...following is a patch for this error.

```
diff --git a/internal/pkg/warewulfconf/datastructure.go b/internal/pkg/warewulfconf/datastructure.go
index 4d0caf7..91ca4be 100644
--- a/internal/pkg/warewulfconf/datastructure.go
+++ b/internal/pkg/warewulfconf/datastructure.go
@@ -51,14 +51,16 @@ type NfsConf struct {
 }
 
 func (s *NfsConf) UnmarshalYAML(unmarshal func(interface{}) error) error {
-    defaults.Set(s)
+       if err := defaults.Set(s); err != nil {
+               return err
+       }
 
-    type plain NfsConf
-    if err := unmarshal((*plain)(s)); err != nil {
-        return err
-    }
+       type plain NfsConf
+       if err := unmarshal((*plain)(s)); err != nil {
+               return err
+       }
 
-    return nil
+       return nil
 }
 
 type NfsExportConf struct {
```

internal/app/wwctl/configure/nfs/main.go
gmkurtzer: @panda1100 I've never used `github.com/creasty/defaults` before, but very cool implementation to set YAML/JSON defaults. But I also think it might be a tad bit overkill, as it would be pretty easy just to evaluate the enabled member. Right?

gmkurtzer:
I really like the `exports extended` format, and I'm wondering if we don't just make that default, and just failover to the previous `exports` for backwards compatibility.

internal/app/wwctl/configure/nfs/main.go
```diff
@@ -56,35 +48,57 @@ func Configure(show bool) error {
...
+			for _, export := range controller.Nfs.Exports {
+				fmt.Fprintf(exports, "%s %s/%s(sync)\n", export, controller.Network, controller.Netmask)
+				fmt.Fprintf(fstab, "%s:%s %s nfs defaults 0 0\n", controller.Ipaddr, export, export)
 			}
-		}
 
+			for _, export := range controller.Nfs.ExportsExtended {
+				fmt.Fprintf(exports, "%s %s/%s(%s)\n", export.Path, controller.Network, controller.Netmask, export.Options)
+				if export.Mount {
+					fmt.Fprintf(fstab, "%s:%s %s nfs defaults 0 0\n", controller.Ipaddr, export.Path, export.Path)
```
gmkurtzer, line 69: I wonder if there would be useful to extend options to the client side mounts as well... Thoughts?

jcsiadal:
I wish to second the ability to support NFS options by default, instead of of a separate extended section. I would find two sections to be an odd layout. I'm writing my new recipe and have to manually create cluster exports with the right values.

Also, I added defaults to the warewulf.conf in my #229 PR. The yaml.v2 package has the fancy ability to pre-load values in the struct and only changes them if those keys are unmarshalled.

internal/app/wwctl/configure/nfs/main.go
jcsiadal: I agree, but warewulf.conf should do the service stuff and nodes.conf should do the client stuff. I don't like mixing functions. I would remove the "mount" key here, since that configuration should reside under nodes.conf.

I'm noticing all of this seems NFS3. In 2021, shouldn't we be defaulting to NFSv4 standards.

PR #186, opened 2021-10-27
Add GitHub action to build new release assets and upload them

elguero:
When a new release is published, the GH workflow will:
- Build the distribution tarball
- Use the tarball to build el7 and el8 RPMS using mock on Rocky Linux
- Upload the tarball and RPMs to the release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,7 @@
 VERSION ?= 4.2.0
 RELEASE ?= 1
 
-ifndef SRC
-	SRC := main
-endif
+SRC ?= main
 
 # auto installed tooling
 TOOLS_DIR := .tools
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 .PHONY: all
 
 VERSION := 4.2.0
+RELEASE := 1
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,6 +1,11 @@
 .PHONY: all
 
 VERSION := 4.2.0
+RELEASE := 1
```
PaulCharlton, line 4: curious if "?=" would be a better pattern in general than ":=", since it allows for the variables to be overridden by makefile command line arguments and inherited environment variables

gmkurtzer:
@elguero Thank you for doing this. We have a merge conflict and a pending question from @PaulCharlton for you to take a look at. Thanks!

Makefile
elguero: Thank you for the suggestion. I have changed the pattern used for both VERSION and RELEASE.

elguero:
@gmkurtzer The merge conflict and the pending question have been addressed. Thanks!

gmkurtzer:
Excellent, thank you @elguero!

PR #194, opened 2021-10-29
use ctime instead of mtime to detect file owner/permission change

panda1100:
`wwctl container exec` doesn't detect file owner/permission change when exit and decide not to build VNFS(https://github.com/hpcng/warewulf/issues/193).

This request implemented to use 'ctime' instead of 'mtime' to detect owner/permission changes.

### DIFF
--- a/internal/pkg/util/util.go
+++ b/internal/pkg/util/util.go
@@ -27,9 +27,14 @@ func DirModTime(path string) (time.Time, error) {
 			return err
 		}
 
-		cur := info.ModTime()
+		fi, err := os.Stat(path)
+		if err != nil {
+			return err
+		}
+		stat := fi.Sys().(*syscall.Stat_t)
+		cur := time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec))
 		if cur.After(lastTime) {
-			lastTime = info.ModTime()
+			lastTime = time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec))
 		}
 
 		return nil
--- a/internal/pkg/util/util.go
+++ b/internal/pkg/util/util.go
@@ -29,7 +29,7 @@ func DirModTime(path string) (time.Time, error) {
 
 		fi, err := os.Stat(path)
 		if err != nil {
-			return err
+			return nil
 		}
 		stat := fi.Sys().(*syscall.Stat_t)
 		cur := time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec))


PaulCharlton:
@panda1100 can you explain how the scope of this change works with various filesystems?   On most filesystems, CTIME is immutable, even when owner and/or permissions change.  On the other hand, MTIME has many possibilities for change -- any change to inode, metadata, or file contents will change MTIME

panda1100:
@PaulCharlton
I tested on ext4 and xfs filesystem.
If you modify file permission and/or modify file owner and /or modify file contents, it updates ctime.
If you modify file permission and/or file owner, it updates ctime but not updates mtime.
I tested with following code(let's say ctime.go).

```go
package main

import (
        "fmt"
        "os"
        "time"
        "syscall"
)

func main() {

        fileName := os.Args[1]
        fmt.Printf("fileName is %s\n",fileName)
        fi, _ := os.Stat(fileName)
        stat := fi.Sys().(*syscall.Stat_t)
        ctime := time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec)).UTC()
        atime := time.Unix(int64(stat.Atim.Sec), int64(stat.Atim.Nsec)).UTC()
        mtime := time.Unix(int64(stat.Mtim.Sec), int64(stat.Mtim.Nsec)).UTC()
        fmt.Printf("ctime: %s\n",ctime)
        fmt.Printf("atime: %s\n",atime)
        fmt.Printf("mtime: %s\n",mtime)
}

```

`go run ctime.go /path/to/file`

it doesn't works at least with symbolic link.

panda1100:
@PaulCharlton
The aim of this change is that to cover permission change, owner change detection (add to current modification detection).

Because, during I built Warewulf cluster,
I found owner of some file (at that time munge.key, that is used for slurm setup) isn't correct, 
so I modified ownership of that file in the Warewulf container.

I was assumed container image is automatically rebuild by owner/permission change,
but It skipped rebuilding all the time.That leads me a little confusion.

The workaround is create dummy file on somewhere, but I believe It's not ideal solution.
This is why I wrote this pull request.

panda1100:
@PaulCharlton 
> This also needs to work when there are hard-links to the same inode, where the directory entries referring to the inode may have concurrently different owners and permissions such as ACL extension records.

Could you provide me a more detailed example of this?
In the context of "container" and "overlay", what kind of scenario do you imagine?

Hard link & ACL
=============

I tested following procdure:

+ create 3 directories with 3 different acl settings on ext4 file system.
+ create hard link refer to same inode on each "acl"ed directories
+ change owner of hard link
+ PR detects ctime change of the link.

Does this answer to  your question?

gmkurtzer (approved):
This looks good to me.

Issue #131, opened 2021-09-15
wwclient update on HUP


gmkurtzer:
wwclient should properly catch the HUP signal and pull/update files on that signal.

PR #196, opened 2021-11-02
wwclient with node identification by hardware adresses

### LINKED ISSUE #192: runtime overlay client identification is based on IP addr only
#### change client identifying entity
The client identification for runtime overlays is based on the IPv4 address only.
This will fail for IPv6 clients, and for clients configured w/o static IP addresses.

https://github.com/hpcng/warewulf/blob/4272e13fa3f17394bed8fa8dfb05ecd5196837b1/internal/pkg/warewulfd/runtime.go#L30

**suggestion:**
change client identification key to a unique entity, i.e. MAC. The MAC is needed anyway as the ipxe processes is based on it.
However, since the whole idea behind this is security, the new identification method should be carefully evaluated in this regard.

kivel:
closes #192 
### new feature

- adds the ability to identify nodes by hardware address, rather than sender IP only.

### mechanism

- the wwclient send Mac addresses encoded via http query.

### fix
The splitting of `host` and `port` is now IPv6 save, at least I think so :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ files: all
 	install -d -m 0755 $(DESTDIR)/usr/share/man/man1
 	test -f $(DESTDIR)/etc/warewulf/warewulf.conf || install -m 644 etc/warewulf.conf $(DESTDIR)/etc/warewulf/
 	test -f $(DESTDIR)/etc/warewulf/hosts.tmpl || install -m 644 etc/hosts.tmpl $(DESTDIR)/etc/warewulf/
+	test -f $(DESTDIR)/etc/warewulf/nodes.conf || install -m 640 etc/nodes.conf $(DESTDIR)/etc/warewulf/
 	cp -r etc/dhcp $(DESTDIR)/etc/warewulf/
 	cp -r etc/ipxe $(DESTDIR)/etc/warewulf/
 	cp -r overlays $(DESTDIR)/var/warewulf/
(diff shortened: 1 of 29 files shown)


gmkurtzer:
This is good thinking, and something I thought of too. The reason I didn't do this before is so no other system could get a particular runtime-overlay then the one it is intended for (for security purposes).

What is a use-case where this would be necessary (and please don't say IPv6! LOL).

kivel:
In our case, the nodes obtain an IP from the central DHCP service. Hence, the IP is _not_ known to the master/daemon.
I know, this is an edge case, as HPC clusters are less likely to have this kind of setup.
In terms of security I'd argue that the respective subnet should be isolated anyway. All members of said network could sniff the traffic, and intercept the runtime overlay that way. 
Imho, there is no harm in having this feature. A statement in the documentation should clearly state that there is a down side in terms of security.

kivel:
@gmkurtzer I made the changes to `wwclient` regarding the logging you proposed.
I also added the SIGHUP handling, @mslacken added in #131 . I made the signal buffered, otherwise `go vet ./...` kept complaining. In an effort to make the code easier to read, the methods where added.

gmkurtzer:
> In our case, the nodes obtain an IP from the central DHCP service. Hence, the IP is _not_ known to the master/daemon. I know, this is an edge case, as HPC clusters are less likely to have this kind of setup. 

Understood and makes sense.

> In terms of security I'd argue that the respective subnet should be isolated anyway. All members of said network could sniff the traffic, and intercept the runtime overlay that way. Imho, there is no harm in having this feature.

Because the runtime overlay isn't being broadcasted, a malicious user would only be able to sniff the traffic if they were root on one of the endpoints, routers, or switches that is passing the data.

But I do concur, there is minimal risk for sniffing. The greater risk is spoofing, but if/when warewulfd is set to `secure`, only root connections would be allowed anyway, which means it would have to be spoofed by a system, on the private cluster network, where a user has root (which might be possible if I'm following your network configuration where the nodes exist on the main network).

> A statement in the documentation should clearly state that there is a down side in terms of security.

Agreed and the mitigation is not to put any secure keys or secrets into the overlays.

gmkurtzer (commented):
I haven't had time to go over or test all of the changes. Have you confirmed that this doesn't change the default usage?

kivel:
> I haven't had time to go over or test all of the changes. Have you confirmed that this doesn't change the default usage?

I did confirm the default behavior at the time I made the changes.
However, the effort by @mslacken to have the `wwclient` as systemd service should take precedence over my changes.
If we agree that this feature is worth adding to the code base. I add the changes on top of #212. Otherwise it'll be a nightmare to deal with the merge conflicts.

gmkurtzer:
This feature is now integrated.

PR #199, opened 2021-11-03
fix FindByHwaddr

### LINKED ISSUE #198: FindByHwaddr fails on mixed character cases
### bug
node `FindByHwaddr` fails when comparing mixed cases strings

### solution

- [ ] cast to lower case before comparing stings
- [ ] add tests

kivel:
fixes #198 
- [x] added tests for FindByHwaddr and FindByIpaddr
- [x] fixed  FindByHwaddr with `strings.EqualFold`

added `early fail` for FindByHwaddr and FindByIpaddr to catch non-MAC and non-IP strings as early as possible


gmkurtzer:
This looks great, thanks!

PR #200, opened 2021-11-04
time since last seen based on runtime overlay request

kivel:
@gmkurtzer you wanted a feature to quickly see which nodes failed to check in, or did so a long time ago.
I implemented such a feature. There are no arguments to filter the output so far.
`wwctl node list -L` gives a list of nodes with the time in seconds since they last requested a runtime overlay.


kivel:
@gmkurtzer 
I moved from writing the data into the config file to holding it in memory as requested. I hope this is what you had in mind.

The build still fails because the go version is too old.
Is there a reason to stay < 1.17?
This feature is only available for >=1.17
https://pkg.go.dev/net/url#Values.Has

gmkurtzer:
Hi @kivel,

Let's not store the time update in the primary nodeDB structure, instead, can you make it a simple map within the daemon itself?

That would be simpler, and not pollute the nodeDB structure with unnecessary complexity.

Thanks!

gmkurtzer:
Oh, and the point you made about Go version... I need to go back and verify, but I believe it is at that Golang version to be aligned with what RHEL/CentOS/Rocky default to.

kivel:
OK, I'll look into moving it to a separate map/structure.
Is the daemon really the right place? Because than it's adding the complexity there instead of the nodedb.

mslacken:
@kivel I accidentally marked this request as ready for review, as this looks good for me. Is this OK for you?

kivel:
To me it is, but I'm afraid @gmkurtzer might want an even less complicated implementation.
I personally like to have the _full_ nodeInfo available, as in the future this feature might grow. If I explicitly cheery pick `Id` and `Clustername` for this reporting, it might fire back along the road.

gmkurtzer:
This PR can be closed as #237 seems to cover this functionality. If there is anything missing, please reopen this, or merge this PR against #237.

Thank you!

Issue #181, opened 2021-10-25
warewulf.conf: ipaddr cannot be FQDN


kivel:
### This is a potential feature request.
for load balancing purposes, it would be advantageous to specify the hostname, rather than the IP address.
Currently, the string is converted to int at the '.', which fails of course.
config:
```
ipaddr: gfa-boot.psi.ch
netmask: 255.255.255.0
```
error:
```
panic: strconv.Atoi: parsing "gfa-boot": invalid syntax

goroutine 1 [running]:
github.com/brotherpowers/ipsubnet.convertQuardsToInt({0xc000035080, 0x3, 0xcf4ef0})
	/go/src/warewulf/vendor/github.com/brotherpowers/ipsubnet/ip.go:28 +0x11a
github.com/brotherpowers/ipsubnet.(*Ip).GetIPAddressQuads(0x0)
	/go/src/warewulf/vendor/github.com/brotherpowers/ipsubnet/ipPortion.go:23 +0x3d
github.com/brotherpowers/ipsubnet.(*Ip).networkCalculation(0xc00059fb90, {0xcf4f3f, 0x2}, {0xcf4ef0, 0x1})
	/go/src/warewulf/vendor/github.com/brotherpowers/ipsubnet/networkPortion.go:25 +0x53
github.com/brotherpowers/ipsubnet.(*Ip).GetNetworkPortion(...)
	/go/src/warewulf/vendor/github.com/brotherpowers/ipsubnet/networkPortion.go:9
github.com/hpcng/warewulf/internal/pkg/warewulfconf.New()
	/go/src/warewulf/internal/pkg/warewulfconf/constructors.go:45 +0x3b1
github.com/hpcng/warewulf/internal/app/wwctl/server/start.CobraRunE(0x13e1f80, {0xcf5ad4, 0x1, 0x1})
	/go/src/warewulf/internal/app/wwctl/server/start/main.go:13 +0x5a
github.com/spf13/cobra.(*Command).execute(0x13e1f80, {0xc0004c1f40, 0x1, 0x1})
	/go/src/warewulf/vendor/github.com/spf13/cobra/command.go:850 +0x60e
github.com/spf13/cobra.(*Command).ExecuteC(0x13d9700)
	/go/src/warewulf/vendor/github.com/spf13/cobra/command.go:958 +0x3ad
github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/warewulf/vendor/github.com/spf13/cobra/command.go:895
main.main()
[...]

PR #202, opened 2021-11-05
Added a check that parses FQDNs to IPv4 format

carterpeel:
Fix for #181

### DIFF
--- a/internal/pkg/util/util.go
+++ b/internal/pkg/util/util.go
@@ -398,3 +398,23 @@ func HostnameToV4(hostname string) (ipv4 string, err error) {
     }
     return ipv4, fmt.Errorf("could not get ipv4 from hostname")
 }
+
+func GetLocalAddresses() (addresses map[string]struct{}, err error) {
+    addresses = make(map[string]struct{})
+
+    ifaces, err := net.Interfaces()
+    if err != nil {
+        return nil, err
+    }
+    for _, v := range ifaces {
+        addrs, err := v.Addrs()
+        if err != nil {
+            return nil, err
+        }
+        for _, v := range addrs {
+            addresses[v.String()] = struct{}{}
+        }
+    }
+
+    return
+}
--- a/internal/pkg/util/util.go
+++ b/internal/pkg/util/util.go
@@ -417,3 +417,22 @@ func AddressExists(addr net.IP) bool {
 	}
 	return false
 }
+
+func GetFirstIPv4(iface *net.Interface) (ipv4 net.IP, err error) {
+	if iface == nil {
+		return nil, fmt.Errorf("iface cannot be nil")
+	}
+	addrs, err := iface.Addrs()
+	if err != nil {
+		return nil, err
+	}
+	for _, v := range addrs {
+		switch ip := v.(type) {
+		case *net.IPAddr:
+			if ipv4 = ip.IP.To4(); ipv4 != nil {
+				return ipv4, nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("could not find any ipv4 addresses")
+}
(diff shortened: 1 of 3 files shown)


gmkurtzer:
I had a conversation with @carterpeel about another possible way to implement this using a device name instead of IP/hostname and netmask.

This should be pretty easy to implement in the constructor in the warewulfconf package.

internal/pkg/util/util.go
```diff
@@ -382,3 +382,49 @@ func IncrementIPv4(start string, inc uint) string {
...
+	}
+	for _, ip := range ips {
+		if v4 := ip.To4(); v4 != nil {
+			return v4, nil
+		}
+	}
+	return ipv4, fmt.Errorf("could not get ipv4 from hostname")
+}
+
+func GetFirstIPv4(iface *net.Interface) (ipv4 net.IP, err error) {
```
gmkurtzer, line 402: Just to be sure I'm following, this function is returning the IPv4 address of a specified network interface configuration?

Is it "the first" just incase the device has multiple IP addresses configured for that interface (e.g. not aliases)?

If it has multiple addresses, do we want to figure that out, and then require someone to define the ipaddr instead?

carterpeel: Yes, the first IPv4 is returned in the event the device has multiple addresses configured...

I'm unsure of how `net.Interface.Addrs()` treats alias addresses, so it's definitely something to consider.

Here's the way `util.GetFirstIPv4()` works functionally:

1. Get all the addresses that exist on the inteface. (`addrs := []net.Addr`)
2. For each value in `addrs`, if the criteria below is met, the IPv4 in question is returned.

- `ip := value.(type)` must be either `*net.IPAddr` or `*net.IPNet`
- Is `r := ip.IP.To4()` `nil`? If yes, do nothing and check the next address in the slice. If no, return `r`.


...I think that prompting the user for input could be helpful, but is that what the community wants? What is the best way to prompt a user for input when `wwctl server start` sends the process to the background?

gmkurtzer: I like that flow, with one potential addition...

If someone specifies the `ipaddr` in the configuration file, what if it matches the second or third address in the `[]net.Addr` array? That should work but I believe it would fail in the current implementation. 

Thoughts?

carterpeel: It doesn't seem to fail the way you described. If `ipaddr` is specified, `device_name` is ignored. I just tested it with the config file shown [here](https://ghostbin.com/gGZCg).

`eth0` looks like [this](https://ghostbin.com/T7KIe).

gmkurtzer:
@carterpeel Just coming back to this and tested. When I remove the ipaddr and netmask from the warewulf configuration, I get the following result from a `wwctl configure nfs`:

```
$ cat /etc/exports
# This file was written by Warewulf (wwctl configure nfs)
/home 0.0.0.0/ffffff00(rw,sync)
/opt 0.0.0.0/ffffff00(ro,sync,no_root_squash)
```

So it seems we have some conversion failures. Can you recheck this?

carterpeel:
> @carterpeel Just coming back to this and tested. When I remove the ipaddr and netmask from the warewulf configuration, I get the following result from a `wwctl configure nfs`:
> 
> ```
> $ cat /etc/exports
> # This file was written by Warewulf (wwctl configure nfs)
> /home 0.0.0.0/ffffff00(rw,sync)
> /opt 0.0.0.0/ffffff00(ro,sync,no_root_squash)
> ```
> 
> So it seems we have some conversion failures. Can you recheck this?

Sure thing. On it ASAP.

gmkurtzer:
I'm going to close this. Hopefully at some point we will reopen and integrate this.

PR #204, opened 2021-11-08
Update help output

jcsiadal:

Reformats help output to common format. Help output is updated to clean up spacing and indentation and to use common format conventions (tar --help is a good example). Because LT/GT signs cannot be used, replaceable text is referenced by capitalized text, with the same capitalized text used in the descriptions.

Some updates to content to reduce filler words.


gmkurtzer (approved):
This is fantastic, thank you @jcsiadal!

Issue #203, opened 2021-11-08
Unable to uniquely reference node in wwctl


robgjansen:
I have nodes named `node1` and `node10`. Whenever I try to set an attribute on `node1` (e.g., `wwctl node set -y ipmiaddr 10.0.0.1 node1`), wwctl also applies it to `node10`. I am unable to find a way to target `node1` but not `node10`.

mslacken:
There is no way to target `node1` right now, but I fixed this in #205. 
Traditionally node names are zero padded, node would have `node01` what avoids this problem.
Thanks for bringing this up.

mslacken:
was closed #205

PR #205, opened 2021-11-09
Adding '^' and '$' to serach string to avoid greedy regexp

mslacken:
regexp.MatchString will match also substrings, what means that 
```
wwctl node list node1
```
will not only match `node1` but also `node10`, so its impossible to target `node1` which is also described in #203.
With '^' at start of the search string and '$' at the end, its possible to target `node1`. The old behavior can be retained with
```
wwctl node list node1.
```


gmkurtzer (approved):
Looks great to me.

PR #209, opened 2021-11-11
implemented dealing of SIGHUP for wwclient

mslacken:
`wwclient` does an imdiate refresh when receiving a SIGHUP. Requested in #131


mslacken:
This issue is also resolved in #212

PR #210, opened 2021-11-12
RFE: retry unconfigured instead of reboot

griznog:
Use autoboot instead of reboot to just retry without going through POST again.

Although I'm using this now and it works, this is not intended to be merged as-is, just a feature request disguised as a PR.

### DIFF
--- a/etc/ipxe/unconfigured.ipxe
+++ b/etc/ipxe/unconfigured.ipxe
@@ -7,6 +7,6 @@ echo
 echo MESSAGE: This node is unconfigured. Please have your system administrator add a
 echo          configuration for this node with HW address: {{$.Hwaddr}}
 echo
-echo Rebooting in 1 minute...
-sleep 60
-reboot
+echo Retrying provisioning in 15s...
+sleep 15
+autoboot


gmkurtzer (commented):
No reason in my mind not to just merge this as it stands if it works. I'm not familiar enough with ipxe autoboot to know if there are any negative effects.

griznog:
This works well in my testing with UEFI, have not tried it with BIOS booting, but I don't see any difference for this. Also this is just a first pass because I got tired of waiting for a full POST when bringing up new nodes. I'd like to add to this to have it display more info about the node and someday it'd be nice if an unconfigured node could make a call to warewulfd and get added to a "discovered" node list or some other holding area where an entry for it would be pre-populated and just need to be approved and renamed.

gmkurtzer:
Warewulfd currently can add nodes that are in a discoverable state on the first PXE request. Are you thinking that it should happen not through the PXE config request, but rather a separate URI endpoint request that occurs when a node is unconfigured?

griznog:
I'm deep into thought experiments here because I typically add all my nodes
to the config beforehand. But thinking about people who like to rack and
cable nodes, then boot them in a specific order to add them it'd be nice if
one could prepopulate a bunch of nodes in the config but with bogus
addresses, then when a node came up either have a menu option on the
console to select a config to assign to it or have a wwctl command to
assign a discovered node to a config. Maybe the discoverable thing does
that already and I overlooked it when bringing up nodes, but unless I had
pre-populated the MAC address all my nodes just kept rebooting.

I'd also like options for memtest, maybe a recovery boot image; I could see
some people liking to spin off a kickstart from the menu, etc. A prettier
menu/interface would be fun, which is the motivation for me playing with
console/framebuffer support in iPXE.

griznog

PR #211, opened 2021-11-15
updated .gitignore to ignore man pages

mslacken:


mslacken:
Changed Makefile so that the binaries for man_page and bash_completion creation use the warewulf.conf within the source tree. This makes the warnings during compilation go away

gmkurtzer (approved):
This looks good to me, but I'm curious... Why did you change the configuration file locations to be vars instead of constants?

mslacken:
> This looks good to me, but I'm curious... Why did you change the configuration file locations to be vars instead of constants?

The configuration locations need to be vars, so that they can't be changed via linker option for the binaries which generate the bash completion and man pages. So this vars can point to `~/warewulf/etc/nodes` what makes the ERROR messages about a missing `nodes.conf` go away, if you compile warewulf on a system without `/etc/warewulf/nodes.conf`

PR #212, opened 2021-11-15
wwclient systemd integration

### LINKED ISSUE #201: Create symlinks in overlays
It would be useful to be able to create symlinks in the overlays. This allows greater flexibility in system configuration tasks with systemd, which requires symlinks to enable systemd service and mount units at boot time.

(The less flexible workaround is to create the needed links in the container, which is not ideal.)

mslacken:
run `wwclient` as systemd service. This has following advantages
* wwclient can be restarted
* other service can depend on it, e.g. munge can now dstributed via runtime overlay
Following changed had to be integrated into `wwclient`
* listen on SIGUP, SIGTERM and SIGKILL
* have a pid file
* deal with symlinks in the overlays (fixes #201)


kivel:
@mslacken , you beat me to it. I was looking into this as well, but had it on the list for next week :)
I'd highly appreciate it if this PR would make it into the master.

mslacken:
Had to force a push, that the legacy is still available

gmkurtzer:
@kivel Did you test this, if so would you mind doing a review/approve on the PR? I haven't had a chance to test this yet.

kivel:
@gmkurtzer, sure, I'll test it next week.

kivel:
@gmkurtzer @mslacken
In general this works for me, if I run the client as __root__.
Does `wwclient` need to be executable by normal users too?
If so, we need to create a directory, `/run/wwclient/` with the respective permissions. Otherwise it'll fail to create the pid file.

- [ ] clarify which users need to run `wwclient`
- [ ] change base directory for pid file from `/var/run` to `/run` to be FHS 3.0 compliant.
- [ ] `pidfile.go` reports misleading error when user has insufficient privs to read/write. @kivel fix this.

mslacken:
> @gmkurtzer @mslacken In general this works for me, if I run the client as **root**. Does `wwclient` need to be executable by normal users too? 
No, its impossible to run `wwclient` as another user than root, as the it has to unpack the runtime overlay which will contain files form other users. So only root can do this.

kivel:
> > In general this works for me, if I run the client as **root**. Does `wwclient` need to be executable by normal users too?
> 
> No, its impossible to run `wwclient` as another user than root, as the it has to unpack the runtime overlay which will contain files form other users. So only root can do this.

OK, so than we should check the UID and exit rigt away if it's not '0'.

mslacken:
@brianclemens @elguero Can you please check how to fix the linter error

elguero:
> @brianclemens @elguero Can you please check how to fix the linter error

@mslacken It looks like the `go.sum` file was not committed that goes along with the `go.mod` change.

```
missing go.sum entry for module providing package github.com/coreos/go-systemd/daemon (imported by github.com/hpcng/warewulf/internal/app/wwclient); to add:\n\tgo get github.com/hpcng/warewulf/internal/app/wwclient)]]"
```

mslacken:
@kivel With the new commit, its now possible to send SIGHUP multiple times. This will always reset the time for the regular update. If the regular update happens always at :30 and SIGHUPO was sent at :45 the next update would also happen at :45

gmkurtzer:
@mslacken I think there are a number of other commits included in this. Can you rebase this against main, and weed out the systemd integration?

internal/pkg/pidfile/pidfile.go
```diff
@@ -0,0 +1,84 @@
+package pidfile
+
```
gmkurtzer, line 2: I wonder if we want to leverage this for the Warewulf daemon as well.

mslacken: Yes, I will put that on my TODO list

mslacken:
Rebased against main. All the changes in this PR are needed for the systemd integration.

mslacken:
@gmkurtzer Can we please merge this PR, as with more work going on in `wwclient` a rebase becomes more and more complicated.
I tested this PR with a openSUSE Leap 15.3 and rocky-8 and all things work as expected.

PR #216, opened 2021-11-23
Add fields for use in iPxe templates

griznog:
Adding two fields to make available to iPxe templates.

- Id : Already there as hostname and fqdn, adding ID to be consistent with node field naming. 
- Cluster: Adding for consistency

The motivation here is to pass values in as kernel params to avoid the need to write files into the image via overlays for trivial node/asset information and to make all values available to the iPxe templates that are available to overlay templates.

These two commits could be folded together, but I didn't know how to do that offhand.


gmkurtzer (approved):

gmkurtzer:
Looks good to me. Merging!

PR #218, opened 2021-11-25
install `wwclient` for all system overlays

### LINKED ISSUE #217: `make install` copies `wwclient` only to default system-overlay
https://github.com/hpcng/warewulf/blob/21cc3ecbb17d0ca27681f46fac0524a8e444dc8c/Makefile#L90

### expected behavior
`make install` should install the `wwclient` to all system-overlays

kivel:
By default `wwclient` will be installed for all system-overlays by `make install` or `make install_wwclient`.

A specific overlay can be targeted by providing the variable `OVERLAY_DIR`.
example for the system-overlay `test`:
```sh
sudo make OVERLAY_DIR=/var/warewulf/overlays/system/test install_wwclient
```
NOTE: the path must _not_ be suffixed with a `/`!

closes #217

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,10 @@ export GOPROXY
 # built tags needed for wwbuild binary
 WW_BUILD_GO_BUILD_TAGS := containers_image_openpgp containers_image_ostree
 
+# system-overlay directories for install_wwclient
+OVERLAY_DIR ?= $(DESTDIR)/var/warewulf/overlays/system/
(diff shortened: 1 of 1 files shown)


gmkurtzer:
This looks good to me, but as we discussed in the community meeting, longer term the overlays will change such that there will always be a `wwinit` overlay, which is where `wwclient` will land, and then the system and runtime overlays will be unified.

gmkurtzer (approved):

PR #224, opened 2021-12-06
add bash completion for node list

mslacken:


elguero (approved):
Tested it and it looks good!

Issue #222, opened 2021-12-02
Booting large container images (and associated items)


griznog:
Summarizing my recent dive into booting container images where the compressed image is > 4 GB. The main problem seems to be a bug in the kernel routines that decompress the initrd/initramfs image which causes it to fail if the image file is > 4 GB. This can be solved in at least these two ways:

1. Leave the image uncompressed (tested up to a 16 GB image on a node with 2 TB of RAM)
2. In the ipxe script use `imgextract` instead of `initrd` to fetch the image, letting ipxe do the decompression step, e.g. 
```
imgextract --name container ${base}/container/{{.Hwaddr}}   || goto reboot
```
I've been doing this with a custom `ipxe.efi` which has gzip support built in and I haven't testing with the ipxe version/build shipped with warewulf. Given an `ipxe.efi` with gzip support, the imgextract option works with no change needed on the warewulf side.  

Other requirements are that this only works with UEFI, legacy BIOS booting/PXE seems to be capped at around a 2 GB image size. 

Closing with some useful search terms for people hitting this problem so this issue shows up:

```
Initramfs unpacking failed: read error

Kernel panic - not syncing: VFS: Unable to mount root fs

Kernel panic - not syncing: No working init found.
```

griznog:
Adding an actual suggestion here and merging in another item from my `ipxe.efi` wishlist, it'd be nice if the things in `internal/pkg/staticfiles/files/tftp` had both imgextract/gzip support as well as console and framebuffer support. 

But in thinking about that it also occurs to me that I don't really know where these come from or are they secure so should these be fetched or built directly from ipxe.org or the ipxe source? I could see someone really obsessed with security possibly wanting to use the IMGTRUST features of ipxe with certs and the works, in which case they'd have to build their own.

gmkurtzer:
@griznog Great suggestions here! Can you please check that #235 is what you tested?

gmkurtzer:
Something to consider here as well is how this will affect ARM as I recall Richard and John from ARM mentioning the iPXE build didn't include zlib support. Will have to test this for ARM, and/or maybe have a secondary iPXE template for ARM.

griznog:
I'm closing this, if/when it becomes and ARM (or Power) issue can reopen.

PR #235, opened 2021-12-30
Implement fix for #222 (thanks Griznog!)

gmkurtzer:
This is a commit for @griznog's suggestion in #222.


gmkurtzer:
To use the new iPXE template do the following:
```
wwctl profile set --ipxe big_image default
```

gmkurtzer:
I am going to merge this, because it just adds an optional iPXE template. We do need to come back though to confirm that `imgextract` works properly on ARM iPXE which I don't think supports zlib.

PR #237, opened 2021-12-30
Simplified version of node status monitoring with CLI

gmkurtzer:
Tracking since last update (#200) is a fantastic idea, but we needed a more REST standards compliant version. This PR provides that in a simple to maintain implementation and includes a basic CLI for usage:

JSON response from Warewulf server after a node has checked in:
```
$ curl http://127.0.0.1:9873/status
{
  "nodes": {
    "n0000": {
      "stage": "iPXE",
      "sent": "big_image.ipxe",
      "ipaddr": "127.0.0.1",
      "last seen": 1640907034
    }
  }
}
```

Here is the response when pulled from the Warewulf CLI and merged into the rest of the configured nodes:
```
$ sudo ./wwctl node status
NODENAME             STATUS               SENT                      LASTSEEN (s)
================================================================================
n0000                iPXE                 big_image.ipxe            46        
n0001                --                   --                        --        
n0002                --                   --                        --        
n0003                --                   --                        --        
n0004                --                   --                        --        
n0005                --                   --                        --        
n0006                --                   --                        --        
n0007                --                   --                        --        
n0008                --                   --                        --        
n0009                --                   --                        --        
n0010                --                   --                        --        
n0011                --                   --                        --        
n0012                --                   --                        --        
n0013                --                   --                        --        
n0014                --                   --                        --        
n0015                --                   --                        --        
n0016                --                   --                        --        
n0017                --                   --                        --        
n0018                --                   --                        --        
n0019                --                   --                        --        
n0020                --                   --                        --        
```

Other data can be pulled in and showed here, please let me know what else would be desirable to know.


kivel:
I couldn't test this so far, but isn't it reporting the nodes in alphabetic order by node name?
When checking the status, I personally am interested in those that _didn't_ check in recently. So the nodes with the longest `lastSeen` should be on top of the list.
This at least was the idea behind #200.

mslacken:
The code also contains also the provision status for the kmods, kernel and container, but these are not shown in the output of `wwctl node status` or did I miss something?

mslacken:
I think it would also be sensible to show a checksum for the image which was sent to nodes.

gmkurtzer:
@kivel:

> I couldn't test this so far, but isn't it reporting the nodes in alphabetic order by node name? When checking the status, I personally am interested in those that _didn't_ check in recently. So the nodes with the longest `lastSeen` should be on top of the list. This at least was the idea behind #200.

Because we are exporting/ingesting JSON from the Warewulf daemon, we can sort the results easily in the CLI depending on anything you want. Right now I'm doing it based on node-name, but it also has filters already built in to support things like showing any node that hasn't checked in within a given time range.

Before we start adding lots of different features to this implementation, I want to get a thumbs up on the base implementation architecture and then make smaller merges for specific feature additions.

gmkurtzer:
> The code also contains also the provision status for the kmods, kernel and container, but these are not shown in the output of `wwctl node status` or did I miss something?

That is the "STAGE" that you see here:

https://github.com/hpcng/warewulf/pull/237/files#diff-5ac32f179ea95035d08e741c566e608bb7880c4811641e60bac3ded4b5c988b1R79

And the "Sent" here:

5ac32f179ea95035d08e741c566e608bb7880c4811641e60bac3ded4b5c988b1R111

I will clean up the member names so everything is the same assume there are no red flags or limitations with this base architecture.

mslacken (commented):
Looks great.

kivel:
> Before we start adding lots of different features to this implementation, I want to get a thumbs up on the base implementation architecture and then make smaller merges for specific feature additions.

👍

elguero:
Looks good for a start.

The end goal is to establish a REST api, correct?

gmkurtzer:
@elguero To start off with, yes, but a very simple REST API. I can see this growing over time, but we will have to make a decision at some point do we go GRPC or REST for the control surface.

gmkurtzer:
@kivel I added options around sorting by node checkin time, reversing, as well as filtering. Here is the updated usage:

```
Usage:
  wwctl node status [OPTIONS] [NODENAME...]

Options:
  -h, --help         help for status
  -l, --last         Sort by the last check-in time
  -r, --reverse      Reverse the sort order
  -t, --time int     Filter by last checkin time (seconds)
  -u, --unknown      Only show nodes of unknown status
  -U, --update int   Set the update frequency for 'watch' (ms) (default 500)
  -w, --watch        Watch the status automatically

Global Options:
  -d, --debug     Run with debugging messages enabled.
  -v, --verbose   Run with increased verbosity.
```

Let me know what you think!

kivel:
> Let me know what you think!

@gmkurtzer very cool!

PR #239, opened 2022-01-05
changelog for netdev and nfs

mslacken:


tcooper:
Not sure if this needs a new issue (likely) but on a multi-homed node how do we differentiate the provisioning interface and interface for the default route? While I don't think it's currently specified I might expect the interface named 'default' to be the one to use as the default route and that may not be the provisioning interface. I'm thinking here of a login node, a compute node with non-NAT'ed public IP.

mslacken:
There is no connection to the `default` route for the network here. As far as I know the interface with a valid gateway assigned will get the `default` route and if there are several gateways, there will also be several `default` routes, but for such complicated setup I would recommend to write a template for the route configuration.

PR #241, opened 2022-01-18
fixups for kernel import

mslacken:
- fixed kernel import from container
- need valid url or dir to import something

### DIFF
--- a/internal/pkg/kernel/kernel.go
+++ b/internal/pkg/kernel/kernel.go
@@ -111,7 +111,7 @@ func ListKernels() ([]string, error) {
 
 func Build(kernelVersion string, kernelName string, root string) (string, error) {
 	kernelDrivers := path.Join(root, "/lib/modules/"+kernelVersion)
-  kernelDriversRelative := path.Join("/lib/modules/"+kernelVersion)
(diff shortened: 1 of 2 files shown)


internal/pkg/kernel/kernel.go
```diff
@@ -111,7 +111,7 @@ func ListKernels() ([]string, error) {
 
 func Build(kernelVersion string, kernelName string, root string) (string, error) {
 	kernelDrivers := path.Join(root, "/lib/modules/"+kernelVersion)
-  kernelDriversRelative := path.Join("/lib/modules/"+kernelVersion)
+	kernelDriversRelative := path.Join("/lib/modules/" + kernelVersion)
```
jcsiadal, line 114: This should be path.Join("/lib/modules", kernelVersion). The use of Join circumvents the need to cat the filename to the main path.

gmkurtzer:
This looks good to me!

internal/pkg/kernel/kernel.go
gmkurtzer: I will fix this in another PR as I think it was my error to begin with. Great catch!

gmkurtzer: addressed in #243.

PR #252, opened 2022-01-24
Initial commit of asset key tag security check

gmkurtzer:
This PR enables a security feature using the asset tag to limit downloading any of the node data.


gmkurtzer:
note: As part of this feature enhancement, `wwclient` will obtain the provisioned `HWADDR` via `/proc/cmdline` which will be written by the iPXE bootstrap template.

gmkurtzer:
TODO: Still need to add UUID to the node data-structure, set commands, and warewulfd logic.

gmkurtzer:
TODO: Still need to add UUID to the node data-structure, set commands, and warewulfd logic.

cmd/wwclient/wwclient.go
```diff
@@ -67,14 +69,38 @@ func main() {
...
+
 	for {
 		var resp *http.Response
 		counter := 0
 
 		for {
 			var err error
 
-			getString := fmt.Sprintf("http://%s:%d/overlay-runtime", conf.Ipaddr, conf.Warewulf.Port)
+			getString := fmt.Sprintf("http://%s:%d/overlay-runtime/%s?assetkey=%s&uuid=%s", conf.Ipaddr, conf.Warewulf.Port, wwid, tag, localUUID)
```
kivel, line 103: in line 88ff. `wwid` is parsed as a generic variable. It is only now used exclusively for the Hwaddr, as the daemon is setting it that way. This might change in the future.
Wouldn't it be prudent to explicitly declare it via `?wwid=%s` in `getString`?

kivel (approved):
I like with PR.
It contains breaking changes, as the /runtime/ entry point of the http service need the hwaddr. This is certainly OK, but need to be clearly pointed out in the changelog.

gmkurtzer:
note: There is an issue with regards to upper and lower case HW addresses.

cmd/wwclient/wwclient.go
gmkurtzer: I'm not sure I follow here... I'm going to merge this as it fixes a issue in current main with provisioning the runtime overlay. Can you open this as an issue?

gmkurtzer:
Merging as this fixes some issues with the runtime overlay at provisioning time.

PR #256, opened 2022-01-25
Missing hyphen in dhcp template path. Unit tests.

MatthewHink:
This is a single character bug fix.
Broke this out into a function for testing.


jcsiadal:
Concur. I currently have to set this to make the template work.
```
dhcp:
  template: static-
```

gmkurtzer:
Great catch, thanks!

PR #257, opened 2022-01-26
Various relocation fixes

gmkurtzer:
There are a number of fixes necessary after merging the relocation features


gmkurtzer:
Merging this, because provisioning isn't working without it (due to `/bin/wwclient` which tramples the system's symlink from `/bin` to `/usr/bin`).

PR #260, opened 2022-01-26
Fix wwctl chmod arg parsing.

MatthewHink:
Tested manually by: sudo wwctl overlay chmod wwinit /warewulf/wwinit 755
Considering how to automate such testing. It may require a make install on the build machine, and I need to find out more about how that works.


MatthewHink:
Sneaking in a unit test.

gmkurtzer:
Looks good!

PR #262, opened 2022-01-27
nfs fix

mslacken:
Even if `/home/` is defined in `warewulf.conf` a second mount will show up
in `fstab` and `exports` as this line is filed via the default settings

- nfs mount option must be defaults
- remove default home mount


gmkurtzer:
I saw this and haven't gotten to it yet, thank you!

mslacken:
For the long term I would more like to see the `/fstab` handled by a `$WWROOT/overivays/wwinit/etc/fstab.ww`, but I have no clue how templates work in go.

Issue #215, opened 2021-11-18
wwctl container build --all does require a container name


mslacken:
From the logic standpoint `wwctl container build --all` should work without providing any container name.
At the moment the command `wwctl container build --all FOO` would work as FOO is in all containers, but I just don't like the syntax.

PR #264, opened 2022-01-27
Fix container build --all. #215.

MatthewHink:
We didn't need to add an arg check here because it's already there.


gmkurtzer:
Great catch!

PR #266, opened 2022-01-28
removed static embeded pxe files in wwctl

mslacken:
removed the static included ipxe in wwctl, instead using the Makefile for installation.


gmkurtzer:
@mslacken Looks great.

Please move the files in `internal/pkg/staticfiles/files/tftp/arm64.efi` to a directory off of the project root like `staticfiles/` and update the `Makefile` accordingly.

Thanks!

MatthewHink:
I can say I spent too much time trying to understand why staticfiles were implemented this way,.

gmkurtzer:
At some point, I was going to make all of Warewulf into a single command `wwctl`. So all files that needed to be installed needed to come from that single binary.

Since we aren't doing that single binary file model anymore, we can pull all of this apart now.

PR #267, opened 2022-01-28
moved fstab creation for nodes to template

mslacken:
the file `/etc/fstab` created with `wwctl configure nfs` takes bits which are baked into wwctl. Having the file as a template `fstab.ww` allows some flexibilty for other mounts.


gmkurtzer:
@mslacken This is really good thinking, awesomeness!

PR #269, opened 2022-01-31
heck if ./.ww4/exludes exist and use if its exists

mslacken:
Usage of this feature is very basic, but an example exlucde file would be
```
/boot/
./var/cache
./usr/src
```
with `./FOO` refering to FOO in the rootfs of the container.
As `wwctld -dv container build FOO` will show the exact command, the 'source' of exclude 
is exhibited to the admin.


gmkurtzer:
@mslacken Can we move this to the `/etc/warewulf/` directory in the container and also drop a default generated version for the particular distribution family if it doesn't already exist on import?

PR #270, opened 2022-01-31
will exec /usr/bin/bash if no cmd is provided

mslacken:
Adds `/usr/bin/bash` as default argument for
`wwctl container exec foo`


gmkurtzer:
Ahh, I see you already merged this. Please see my comments from #270.

PR #271, opened 2022-01-31
execute './.ww4/runExit` on exit of container exec

mslacken:
After exiting a shell with `wwctl container exec` a script/executeable `./.ww4/runExit`
will be executed.


gmkurtzer:
@mslacken This is a very interesting feature, great thinking!

I'm assuming the use-case is to do things like `dnf clean all`? 

Can you write a default file that includes reasonable OS logic to clean things appropriately, when importing a container into Warewulf if it doesn't exist already.

Lastly, can we rename the file to be more specific? Perhaps `/etc/warewulf/container_exit.sh`?

Thanks!

gmkurtzer:
Actually, in looking at this further, how does this ever get to this point in the code due to the prior `syscall.Exec()`?

PR #273, opened 2022-02-02
added option for automatic kernel detection

mslacken:
With the `-D` flag a kernel can now be detected under the directories provided by kernelSearchPaths.
This detection needs either container via the `-C` option or a chroot via the `--root` option to work.


gmkurtzer:
Great idea, and looks good!

PR #274, opened 2022-02-02
add some eyecandy

mslacken:
adds some information about the node before login


overlays/wwinit/etc/issue.ww
```diff
@@ -0,0 +1,8 @@
+ode: {{.Id}}
```
MatthewHink, line 1: Typo here, otherwise +1.

MatthewHink (commented):
This is useful.

PR #275, opened 2022-02-02
Centralize build variables to Makefile

jcsiadal:
*** Needs rebase and this is still in testing. Adding here for review and get some feedback.

The goal of the change is to eliminate the need to make changes in multiple locations when adding or modifying a variable. Secondarily, the warewulfconf package becomes the runtime authority for system and server configuration.

A makefile variable can be exported by adding it to a list (the variable VARLIST in the makefile). This will save it's value to the Defaults.mk file and add it to the update list for .in files and make it available to the warewulfconf package.

warwulfconf makes both build and runtime variables available to other packages. Most of these are fixed values at build time, but adding them as a configurable option for runtime becomes much simpler. Still more future work to do here, since other packages shouldn't need to create or read the config file directly.

Also some general cleanup encountered while updating config values.


gmkurtzer:
@jcsiadal Looking forward to testing this. I merged some PRs that were ahead of this and now we've got some merge conflicts. Will you resolve, and I will test shortly.

jcsiadal:
I started splitting this up due to the number of changes. I have one PR submitted and will add more.

PR #279, opened 2022-02-04
finaly added /usr/bin/bash as default to cnt exec

mslacken:


gmkurtzer:
@mslacken Oh, that is super helpful, but perhaps it would be better to add a `shell` command for this feature instead of `exec`? Thanks!

gmkurtzer:
Looks fantastic, thank you!

PR #291, opened 2022-02-11
Rename custom key/value entries to "tags"

gmkurtzer:
Also, changed the interface to make it easier to manage and be able to specify multiple key/value tags in a single command line.

### DIFF
--- a/internal/pkg/node/datastructure.go
+++ b/internal/pkg/node/datastructure.go
@@ -36,7 +36,7 @@ type NodeConf struct {
 	Discoverable   string              `yaml:"discoverable,omitempty"`
 	Profiles       []string            `yaml:"profiles,omitempty"`
 	NetDevs        map[string]*NetDevs `yaml:"network devices,omitempty"`
-	Keys           map[string]string   `yaml:"keys,omitempty"`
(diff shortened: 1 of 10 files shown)


internal/pkg/node/datastructure.go
```diff
@@ -36,7 +36,7 @@ type NodeConf struct {
 	Discoverable   string              `yaml:"discoverable,omitempty"`
 	Profiles       []string            `yaml:"profiles,omitempty"`
 	NetDevs        map[string]*NetDevs `yaml:"network devices,omitempty"`
-	Keys           map[string]string   `yaml:"keys,omitempty"`
+	Tags           map[string]string   `yaml:"tags,omitempty"`
```
MatthewHink, line 39: Be aware of the yaml change here.

MatthewHink (approved):

internal/pkg/node/datastructure.go
gmkurtzer: I don't know of anyone using this (aside from the one case that was using it accidentally thinking it was the assetkey). I can add a bit of code though to automatically convert it and provide backwards compatibility.

gmkurtzer: Done.

PR #280, opened 2022-02-05
Updates for RPM building and client files

jcsiadal:
Updates specfile to match distro guidelines. Suppots modifying client directory path at build time.

Verified working on Rocky 8.5.

CHANGLELOG

Adds
- Support relocating client files at build time (required for LSB compliant nodes)
- OS variable added to perform some make choices

Changes

- Saving and exporting the Makefile vars now scripted using list of variables
- Updated specfile to match RH and SUSE requirements and guidelines
- Moved /usr/local/var to /var/local; dynamic files shall not use /usr.
- State files use %sharedstatedir for path (std for SLE and RHEL)
- Release value adjusted to meet different DEB and RPM reqs.

Signed-off-by: jcsiadal <jeremy.c.siadal@intel.com>

PR #293, opened 2022-02-11
Move configuration code out of wwctl

jcsiadal:
Branched from PR #280.

Configuration functions are moved out of wwctl, so that they can be accessed without calling wwctl code. New configuration package implements a single exported function.

CHANGELOG

Changes

- Relocate non-Cobra functional code in 'configure' from wwctl to pkg.
- Convert repeated code in ssh and tftp to loops.
- Make 'show' argument work correctly on all configs


jcsiadal:
I did a basic validation test on Rocky 8.5, after installing from RPM using the generated specfile.

PR #300, opened 2022-02-13
Container exit script runner

gmkurtzer:
This is a different implementation for the runner in #271 that executes the `container_exit.sh` script when the container exits.


gmkurtzer:
Still need to drop a default file at `/etc/warewulf/container_exit.sh` when the container is imported into Warewulf.

PR #304, opened 2022-02-14
Host overlay

mslacken:
Added the creation of host overlays, what means that files under
`OVERLAYDIR/host` will be placed in the root directory where `wwctl` is executed. 
This will be helpful e.g for the `/etc/exports` which than can be created from `OVERLAYDIR/etc/exports.ww` what adds a greater freedom for the admin, and the `exports` files does not come out of the blue sky.


gmkurtzer:
I really like this idea and I'll iterate through it over the next day or so.

Nice work!

mslacken:
`/etc/hosts` and '/etc/dhpcd.conf` are also now manages via host templates

gmkurtzer:
Please coordinate and merge this against #311.

gmkurtzer:
This has been merged. Thank you!

PR #306, opened 2022-02-15
Utilize containerized kernel by default

gmkurtzer:
There have been a number of cases where incorporating a driver that has both a kernel and userspace component has been difficult due to the userspace and kernel coming from different locations (host Vs. containers). Some packages are GPUs, Infiniband, and Lustre.

This change allows for the kernel to be included and distributed inside the container. Warewulf will now find the newest kernel within the container and boot on that automatically.


kivel:
Will this remove the ability to set a kernel per node or profile, and the kernel has to be in the container?

internal/pkg/container/kernel.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+
+		if len(kernels) > 1 {
+			sort.Slice(kernels, func(i, j int) bool {
+				return kernels[i] > kernels[j]
+			})
+			wwlog.Printf(wwlog.VERBOSE, "Multiple kernels found in container: %s\n", container)
+			return kernels[0]
+		} else if len(kernels) == 1 {
+			return kernels[0]
```
MatthewHink, line 42: This seems like the success case here.

internal/pkg/container/kernel.go
```diff
@@ -0,0 +1,47 @@
...
+		// This is a printf format where the %s will be the kernel version
+		"/boot/vmlinuz-*",
+		"/lib/modules/*/vmlinuz*",
+	}
+)
+
+func KernelFind(container string) string {
+	container_path := RootFsDir(container)
+	if container_path == "" {
+		return ""
```
MatthewHink, line 22: Should we consider returning an error or (ok == false) here and below? Does the caller need to check for an empty string returned here and NOT_FOUND below?

internal/pkg/container/kernel.go
```diff
@@ -0,0 +1,47 @@
...
+		kernels, err := filepath.Glob(check_path)
+		if err != nil {
+			return ""
+		}
+
+		if len(kernels) > 1 {
+			sort.Slice(kernels, func(i, j int) bool {
+				return kernels[i] > kernels[j]
+			})
+			wwlog.Printf(wwlog.VERBOSE, "Multiple kernels found in container: %s\n", container)
```
MatthewHink, line 39: Consider if there is an API around this. The caller won't see the log message.

internal/pkg/container/kernel.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"path"
+	"path/filepath"
+	"sort"
+
+	"github.com/hpcng/warewulf/internal/pkg/wwlog"
+)
+
+var (
+	kernelSearchPaths = []string{
```
MatthewHink, line 12: If the kernel search paths are passed in as a parameter to KernelFind, it may be easier to write a functional test. The constants here are great. Consider parameterization in the function.

MatthewHink (commented):
This looks like a good PR. Have some concerns around KernelFind, but could be an overreaction on my part.

mslacken:
I do think we should leave it the way it is as it may be useful to run different containers with the same or different containers with the same kernel.
kernel import is also quite easy now, as it is just
`wwctl kernel import -C Container -D`
which will import the kernel as the container name.
Also kernel updates will become hilarious difficult in the container and confuse the users.

internal/pkg/container/kernel.go
```diff
@@ -0,0 +1,47 @@
...
+		// This is a printf format where the %s will be the kernel version
+		"/boot/vmlinuz-*",
+		"/lib/modules/*/vmlinuz*",
```
rhenwood-arm: Kernel location probably should have a commandline override so we can specify the kernel location: see the discussion here https://github.com/hpcng/warewulf/issues/156

gmkurtzer:
This PR has been updated to by default use the kernel within a container but also allow a kernel to be used as an override if the kernel isn't present or if someone wishes to test or use a different kernel (this was discussed in our community meeting).

internal/pkg/container/kernel.go
gmkurtzer: Yep, exactly.

internal/pkg/container/kernel.go
gmkurtzer: Yeah, as it stands now, if it isn't found, it returns empty string.

internal/pkg/container/kernel.go
gmkurtzer: Yep, but that's gonna be a much bigger change I think, unless you are suggesting we return verbose and/or `error` messages to the caller?

internal/pkg/container/kernel.go
gmkurtzer: Can you elaborate on this?

internal/pkg/container/kernel.go
gmkurtzer: The kernel location doesn't matter as long as it is in a standard location. Are you asking for kernels that aren't in standard locations?

rhenwood-arm: It looks like this isn't an issue any more -- so I think the defaults are fine... and if they don't work, then I can always move my kernels within a container and re-import :)

internal/pkg/container/kernel.go
MatthewHink: Cool - makes sense for the caller to look for one and only thing.

internal/pkg/container/kernel.go
MatthewHink: For now it may be best to just log. Plan is for the API to call identical code as the CLI.

griznog:
Switching from slack to here, in testing this I'm seeing some strange behavior. If I create a new node, leave it with all defaults and point it at a freshly imported container, it boots fine. If I then point that at one of my local containers, it can't find the kernel. An `strace` reveals `wwctl` is looking for the entirely wrong kernel version of that container:

```
pid 1683878] newfstatat(AT_FDCWD, "/datapool/warewulf/chroots/rocky8-x86_64/rootfs/boot/vmlinuz-4.18.0-348.el8.x86_64", 0xc0001dab98, 0) = -1 ENOENT (No such file or directory)
[pid 1683876] <... nanosleep resumed>NULL) = 0
[pid 1683878] newfstatat(AT_FDCWD, "/datapool/warewulf/chroots/rocky8-x86_64/rootfs/boot/vmlinuz-4.18.0-348.el8.x86_64.gz",  <unfinished ...>
[pid 1683876] nanosleep({tv_sec=0, tv_nsec=20000},  <unfinished ...>
[pid 1683878] <... newfstatat resumed>0xc0001dac68, 0) = -1 ENOENT (No such file or directory)
[pid 1683878] newfstatat(AT_FDCWD, "/datapool/warewulf/chroots/rocky8-x86_64/rootfs/lib/modules/4.18.0-348.el8.x86_64/vmlinuz", 0xc0001dad38, 0) = -1 ENOENT (No such file or directory)
[pid 1683878] newfstatat(AT_FDCWD, "/datapool/warewulf/chroots/rocky8-x86_64/rootfs/lib/modules/4.18.0-348.el8.x86_64/vmlinuz.gz", 0xc0001dae08, 0) = -1 ENOENT (No such file or directory)
[pid 1683878] openat(AT_FDCWD, "", O_RDONLY|O_CLOEXEC <unfinished ...>
```

That container has installed it 
```
kernel-4.18.0-348.12.2.el8_5.x86_64
kernel-core-4.18.0-348.12.2.el8_5.x86_64
kernel-devel-4.18.0-348.12.2.el8_5.x86_64
kernel-headers-4.18.0-348.12.2.el8_5.x86_64
kernel-modules-4.18.0-348.12.2.el8_5.x86_64
kernel-tools-4.18.0-348.12.2.el8_5.x86_64
kernel-tools-libs-4.18.0-348.12.2.el8_5.x86_64
```
The previously imported upstream container has the same kernel installed and works fine. Looking in the broken container, I can't find any reference to the kernel version that `wwctl` tries to open, no idea where it is getting that from.

griznog:
Upon closer inspection it turns out that because I had installed things that used `extras` in the `/usr/lib/modules/$KERNELVERSION/extras`, the removal of those kernel rpms does not clean up the `/usr/lib/modules/$KERNELVERSION` directory. It looks like the search for a kernel picks up those version strings and keeps them even if there is no associated kernel? Suggest either not looking in `/lib/modules` to get kernel versions or verifying that an associated kernel actually exists before trying to use it. 

@gmkurtzer  related to this it'd be nice if when multiple kernel versions are found, the latest version is used as the default. Although sorting that properly from distro to distro seems non-trivial. `sort -V` fails to do the right thing because the .x release kernel breaks the version numbering pattern for easy sorting.

gmkurtzer:
@griznog Great catch, you are right, my logic is faulty there for things like extra kernels. Will get to that this evening and I think there are ways to do it (e.g. glob for `/lib/modules/*/vmlinuz` and/or other tell-take files).

What is interesting is that the kenrel doesn't seem to be installed to `/boot/vmlinu*` anymore on EL8 (or am I missing an RPM?).

griznog:
> What is interesting is that the kenrel doesn't seem to be installed to `/boot/vmlinu*` anymore on EL8 (or am I missing an RPM?).

> What is interesting is that the kenrel doesn't seem to be installed to `/boot/vmlinu*` anymore on EL8 (or am I missing an RPM?).

On my Rocky8 systems `kernel-core` provides that. 
 
```
# rpm -qif /boot/vmlinuz-4.18.0-348.12.2.el8_5.x86_64
Name        : kernel-core
Version     : 4.18.0
Release     : 348.12.2.el8_5
Architecture: x86_64
Install Date: Thu 17 Feb 2022 03:27:44 PM PST
Group       : System Environment/Kernel
Size        : 70863759
License     : GPLv2 and Redistributable, no modification permitted
Signature   : RSA/SHA256, Wed 19 Jan 2022 12:36:31 PM PST, Key ID 15af5dac6d745a60
Source RPM  : kernel-4.18.0-348.12.2.el8_5.src.rpm
Build Date  : Wed 19 Jan 2022 10:18:16 AM PST
...
```

griznog:
> 

This didn't work for me:

```
# wwctl kernel import -C rocky8-x86_64 -D
Found kernel at: /datapool/warewulf/chroots/rocky8-x86_64/rootfs/boot/vmlinuz-0-rescue-738593b281574d0cb9be23cac591a56d
[ERROR]   : Failed building kernel: Could not locate kernel drivers
```
That kernel comes from dracut maybe? It's not in an rpm:

```
# rpm -qif /boot/vmlinuz-0-rescue-738593b281574d0cb9be23cac591a56d
file /boot/vmlinuz-0-rescue-738593b281574d0cb9be23cac591a56d is not owned by any package
```

mslacken:
> 
> ```
> # wwctl kernel import -C rocky8-x86_64 -D
> Found kernel at: /datapool/warewulf/chroots/rocky8-x86_64/rootfs/boot/vmlinuz-0-rescue-738593b281574d0cb9be23cac591a56d
> [ERROR]   : Failed building kernel: Could not locate kernel drivers
> ```
> 
> That kernel comes from dracut maybe? It's not in an rpm:
> 
> ```
> # rpm -qif /boot/vmlinuz-0-rescue-738593b281574d0cb9be23cac591a56d
> file /boot/vmlinuz-0-rescue-738593b281574d0cb9be23cac591a56d is not owned by any package
> ```
The problem here is that the kernel version could not be deduced from the name. `rescue-738593b281574d0cb9be23cac591a56d` is a strange name for a kernel.
This needs not to be related to dracut, but seems most likely to be the case. The only proper solution would be to parse the found kernel for its version.

mslacken:
When working on kernel and modules can you please also add the firmware, see #317

internal/pkg/container/kernel.go
gmkurtzer: This will have to be a bigger fix for API integration. For now, I'm going to leave this as it stands for compatibility with existing standard (all of which we will go back and fix at once once we've defined the best way to do it).

Thanks and great thinking!

gmkurtzer:
Now including `/lib/firmware` in kernel build, but keep in mind, it can grow big:

```
$ du -sh /lib/firmware/
687M	/lib/firmware/
```

PR #307, opened 2022-02-15
Remove unique settings from profile set. Warn if set in the node DB.

MatthewHink:
IP address
MAC address
Asset Tag

IP and MAC are unique to each node. It seems like asset tag should be unique as well.

### DIFF
--- a/internal/app/wwctl/profile/set/main.go
+++ b/internal/app/wwctl/profile/set/main.go
@@ -70,6 +70,11 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 			p.Root.Set(SetRoot)
 		}
 
+		if SetAssetKey != "" {
+			wwlog.Printf(wwlog.VERBOSE, "Profile: %s, Setting asset key to: %s\n", p.Id.Get(), SetAssetKey)
+			p.AssetKey.Set(SetAssetKey)
+		}
+
 		if SetKernel != "" {
 			wwlog.Printf(wwlog.VERBOSE, "Profile: %s, Setting Kernel to: %s\n", p.Id.Get(), SetKernel)
 			p.KernelVersion.Set(SetKernel)
--- a/internal/app/wwctl/profile/set/main.go
+++ b/internal/app/wwctl/profile/set/main.go
@@ -70,11 +70,6 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 			p.Root.Set(SetRoot)
 		}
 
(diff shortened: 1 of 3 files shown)


internal/app/wwctl/profile/set/main.go
```diff
@@ -70,11 +70,6 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 			p.Root.Set(SetRoot)
 		}
 
-		if SetAssetKey != "" {
-			wwlog.Printf(wwlog.VERBOSE, "Profile: %s, Setting asset key to: %s\n", p.Id.Get(), SetAssetKey)
-			p.AssetKey.Set(SetAssetKey)
-		}
-
```
gmkurtzer, line 77: While I wouldn't personally recommend this, we don't have to force asset keys to be unique, so let's allow this in a profile configuration.

internal/app/wwctl/profile/set/root.go
```diff
@@ -91,7 +88,6 @@ func init() {
 	baseCmd.PersistentFlags().StringVarP(&SetIpxe, "ipxe", "P", "", "Set the node's iPXE template name")
 	baseCmd.PersistentFlags().StringVarP(&SetInit, "init", "i", "", "Define the init process to boot the container")
 	baseCmd.PersistentFlags().StringVar(&SetRoot, "root", "", "Define the rootfs")
-	baseCmd.PersistentFlags().StringVar(&SetAssetKey, "assetkey", "", "Set the node's Asset tag (key)")
```
gmkurtzer, line 94: And thus leave this too.

internal/pkg/node/constructors.go
```diff
@@ -229,7 +229,9 @@ func (config *nodeYaml) FindAllProfiles() ([]NodeInfo, error) {
 		p.RuntimeOverlay.Set(profile.RuntimeOverlay)
 		p.SystemOverlay.Set(profile.SystemOverlay)
 		p.Root.Set(profile.Root)
-		p.AssetKey.Set(profile.AssetKey)
+		if profile.AssetKey != "" {
+			wwlog.Printf(wwlog.WARN, "Ignoring asset key %v in profile %v\n", profile.AssetKey, name)
+		}
```
gmkurtzer, line 234: Let's do this for ipaddr and hwaddr instead of assetkey.

gmkurtzer:
This looks good to me, thanks!

PR #312, opened 2022-02-17
Initialise Keys map for overlay templates

Prepultrue:
`tstruct.Keys` is currently uninitialised, leading to panics when custom keyvalue pairs are used.


gmkurtzer:
Great catch, thank you for the fix @Prepultrue!

gmkurtzer (approved):

PR #315, opened 2022-02-17
added kernel search path for arch compatibility

TerranWorks:
added search path for kernel to allow arch users to utilize the `` wwctl kernel import `` functionality


mslacken:
Great to see warewulf also running on arch.
Thanks for your contribution.

gmkurtzer:
This is fantastic, thank you @TerranWorks!

PR #318, opened 2022-02-23
Only write to IPMI if this is explictly wanted

mslacken:
As soon as the init script had a valid IMPI configuration, the script always set this values. This may not be wanted as IMPI is sometimes centrally managed with other tools.


gmkurtzer:
Love this idea, just needs a quick conflict fix. Thanks!

gmkurtzer:
Why was removal of the container child code included here?

Please note, this breaks all container exec and shell commands.

mslacken:
@gmkurtzer Sorry this was an error from my side, I wanted this a a separate PR. 
But are this code really needed? When working on #326 I touched the exec/shell part in order to find out, if `/etc/passwd` was modified and put this code into the child, where it was never executed?

gmkurtzer:
Yes, that code is 100% required.

I do remember you putting code into the child/main.go (#271), but it was after the `syscall.Exec()` as well as the `os.Exit()` on error, which means it would never be executed.

PR #320, opened 2022-02-24
Handle KernelOverride conditionally in ipxe scripts

griznog:
- Utilize containerized kernel and remove kernel Support
- Fixes for kernel identification within the container
- Leverage the upstream containers and add kernel to container
- rename openhpc to lowercase
- Specify full host URI
- Restored deleted kernel components for overrides
- Support kernel overrides from the host or other containers
- Updated container recipes to include kernel and use upstream base containers
- Make displaying the kernel handle the default case when an override is not specified.


griznog:
Struggling with git here, all I wanted was to add my two commits but somehow lumped it all together. Not sure how to create a PR based off a branch and not include the rest of that branch.

gmkurtzer:
You added your patches to my PR, so it included them. No worries, everything is good, merging!

PR #311, opened 2022-02-16
Move operational code for configuration out of wwctl 

jcsiadal:
Configuration functions are moved out of wwctl, so that they can be accessed without calling wwctl code. New configuration package implements a single exported function.

Should make it easier to add more configuration options.

Many of the minor code format changes were done automatically by Google's VSCode Go plugin.  

CHANGELOG

Additions

- SafeFileCopy added to util; won't overwrite destination files
- Restores (only missing) iPXE files when TFTP is configured

Changes

- Relocate non-Cobra functional code in 'configure' from wwctl to pkg.
- Convert repeated code in ssh and tftp to loops.
- Make 'show' argument work correctly on all configs

PR #323, opened 2022-02-28
Host overlays

gmkurtzer:
Merging #304 and #311 into a single PR with some added fixes.


gmkurtzer:
Merging now that all checks have passed.

PR #326, opened 2022-03-02
Sync-uids

mslacken:
Adds the `wwctl container syncuser` what will synchronize the users from `/etc/[passwd|group]` from the host to the container. File ownerships are changed accordingly. The call will fail if there is uid/gid collision, e.g if the user only exists in the container and has the same uid/gid as another user on the host. 
On success `/etc/[passwd|group]` are copied to the container and the users/gid exclusive to the container are added.
The function is also called in container import.


mslacken:
Can also someone please test it with rocky host?

gmkurtzer:
This shouldn't be necessary since we are doing the `/etc/passwd.ww` which pulls in passwd entries from all sources.

mslacken:
> This shouldn't be necessary since we are doing the `/etc/passwd.ww` which pulls in passwd entries from all sources.

It is, as the synchronization is done at the import of the container. What you forget is that the template `passwd.ww` is only active on the node, but *NOT* at the time of `wwctl container exec`. So at the moment we have an absolute incoherent state, where you have other uids/gids when doing a `wwctl container exec` as when you run the container.

griznog:
This is non-trivial to address, given the existence/use of overlays any `wwctl container exec` will always mean that there is a split-brain approach to managing images and configurations. Even if you went to the effort to set up a "virtual" node for each container, create the associated overlays, mount them with as an overlayfs and then exec in the resulting container/overlay mashup, how do you handle changes to the container/overlays and get them back into the right place and track them? The current model makes it effectively impossible to have any kind of comprehensive revision control over image configuration, which pretty much applies to any system which allows modification to the container itself, short of one that perhaps used filesystem snapshots for revision control. 

In my current setup I actually do use filesystem snapshots to manage the container chroots and `git`/`svn` to manage overlays with the long term goal to move all configuration out of the container and into overlays thus eliminating the need to modify the container beyond a simple, tractable `dnf`/`apt` change. But given current limitations of overlays (limited/fixed number of overlays, no script hooks available during overlay creation, only templates for node-specific changes...) it's just not a rich enough mechanism to be a complete configuration system. The `wwinit` sysv over systemd approach further conflates the issue. Architecturally it's really hard to manage node configurations as we currently do if you want to provide any kind of strict change control/tracking framework to whole image/node management. Especially once you move away from the case of a simple cluster with homogeneous nodes and files based `nsswitch.conf`

mslacken:
> This is non-trivial to address, given the existence/use of overlays any `wwctl container exec` will always mean that there is a split-brain approach to managing images and configurations. 
Having the uids/gids synced at import exactly avoids another split brain situation. If a package adds a new user there will be a change in `/etc/passwd` which is monitored.
At the moment at every container import the users are synchronized, would it be better, if only a warning is printed if there is a difference for uid/gid in `/etc/passwd` at container import?

griznog:
Backing up a level, syncing uid/gid from the master is inherently broken for any use case beyond the simple cluster case where the nodes are the same OS/config as the master. While that case is certainly valid as a basic/default setup, anything done to support it should be done in a way that doesn't adversely impact using WW to provision more complex environments. Almost by definition a more complex environment will not have `/etc/[passwd|group]` distributed this way, especially if provisioning images from other distributions than the master node. This should definitely be a configurable thing that is easy to disable.

gmkurtzer:
I think we need to discuss this feature in our next meeting as I have some concerns about what this might do when system account UIDs are different on host and container via packages getting installed at different times (e.g. `ssh`).

tcooper:
I tend to agree with @gmkurtzer on this. We maintain a flat namespace for users/groups at $workplace and doing so creates numerous headaches that require complicated processes in an attempt to ensure there is not unexpected UID/GID collision in/on managed systems.

mslacken:
You both exactly want what I want. Avoid the collision of uid/gid from the host and container. So this PR does following:
* Check the users of host and container at import and print a warning if there any differences
* Add the `wwctl container syncuser $CONTAINER` which will sync the files and user with the same name in the container and on the host, add *all* users from `/etc/passwd` from the host to the container and retain the users which are only in the container, but bail out if there are unresolveable conflicts.
* monitor `/etc/passwd` and `/etc/group` for changes from `wwctl container exec` commands

griznog:
There should be no requirement that these be in sync, it's entirely possible someone might provision multiple clusters from the same warewulf master, with each cluster differing in distribution, users/group info source, architecture, ???. 

Also in the case where these are intended/desired to be in sync with the master it's not just sufficient to sync `/etc/[passwd|group|shadow|gshadow]`, it's also necessary to walk the entire container filesystem(s) correcting file and directory ownership for every remapped system account. 

Keeping them in sync also requires any system accounts in the container missing from the master to be created/sync'd back to the master making this a two-way replication. Otherwise the assumption these are in sync breaks when browsing/modifying the container/chroot from the master directly (skipping `wwctl container exec`).

The goal of keeping these in sync with the master seems to be a regresssion back to earlier warewulfs where the master was the source of everything node-ish. If that is what someone wants to do, a better approach might be a command to build a container from the master, e.g. scrape installed packages, settings, user/group info, etc., and use those to build a chroot that then stays in sync with the master. Pulling in arbitrary containers and then trying to make them look like the master seems like a much more difficult path to get to the same result. Or one could just use Bright 😉

mslacken:
As discussed in the community meeting the `SyncUid` command will now run at every `wwctl container import` and `wwctl container exec` so that the uids/gids get synchronzied.
Both commands now have the `--nosyncuser` switch to avoid this behavior.

Issue #305, opened 2022-02-14
#301 correction for SLES?


truatpasteurdotfr:
Hi,

#301 is hardcoding `/var/lib/tftpboot/warewulf` whereas there is some sles specific definition at
lines 2-9 of warewulf.spec.in 
``` 
%if 0%{?sle_version}
%global tftpdir "/srv/tftpboot"
%global srvdir "/srv"
%else
%global tftpdir "/var/lib/tftpboot"
%global srvdir "%{_localstatedir}/warewulf/srv"
%endif
```

line 111
``` 
%attr(-, root, root) "%{tftpdir}/warewulf
```

gmkurtzer:
Done, thanks @truatpasteurdotfr!

PR #332, opened 2022-03-05
Fixing RPM build error caused by #301

gmkurtzer:
Thank you @truatpasteurdotfr for catching this (ref #305)


gmkurtzer:
More fixes to rpmbuild, so that RPMs build (tested on Rocky LInux).

PR #334, opened 2022-03-09
LSB Compliant installation added

jcsiadal:
Update to install Warewulf in an LSB-compliant filesystem if it can determine the OS ID; otherwise, it uses GNU defaults. If the PREFIX is set, it always uses GNU defaults for subdirectories. Outcome is closer to other OSS installs which implement autoconf or cmake.

The RPM spec now correctly installs state stuff to sharedstatedir. This is the standard for Fedora and SUSE.

The variable scan and update now uses a loop; this is just the first step to removing the need to manually update buildconfig variables in multiple locations if they're renamed or added.

The Warewulf client still needs to be updated to support relocation. I left it as-is for now since we discussed eliminating and moving some parts of it.

CHANGELOG

Additions
- License file added to install

Changes
- Default installs on recognized OSs use LSB-compliant subdirectories.
- RPM defaults to sharedstatedir instead of localstatedir
- Default.mk is dynamically created.
- Minor fixups of variable names


mslacken:
Looks good.

PR #342, opened 2022-03-21
Remove-base-template

mslacken:
* if the template provides multiple files, the base file won't be written
* add the255.255.255.0 as default netmask to every interface


mslacken:
Merging this one in, as this fixes the problems with the munge.key for @griznog and the changes of the config to hold the templates will touch the same files, so merging will become complicated otherwise.

PR #343, opened 2022-03-24
store the ovelays as yaml list

mslacken:
Changes the backend that the Entry struct holds `[]string` instead of `string`. As the Entry values are accessed with Get and Set, the conversion can be done there.


griznog:
This works well on my test setup:

```
[root@corbicula overlays]# wwctl profile add myprofile
[root@corbicula overlays]# wwctl profile set --runtime ssh,autofs,nhc,warewulf myprofile
[root@corbicula overlays]# cat /etc/warewulf/nodes.conf 
nodeprofiles:
  default:
  myprofile:
    runtime overlay:
    - ssh
    - autofs
    - nhc
    - warewulf
    network devices:
      default: {}
...
```

Booting a node with that profile gets it everything from the list.

mslacken:
I also found no bugs, so merging in.

PR #347, opened 2022-03-30
warewulf.conf option disable host overlays

mslacken:
Adds the bool option 
warewulf:
  host overlay: true
so that host overlays can be disabled. The command 
`wwctl build overlay -H` will still build the host overlay, as
this may be explicitly wanted.


griznog:
Tested and working.

PR #352, opened 2022-04-01
use secure overlay dir

mslacken:
If there is a user writeable file in the host overlay, a user can
create arbitrary files on the host root file system. Although these
files will belong to this user, this might allow users to modify
any files.
So this patch restricts the access to the host overlay to root and
produces a warning if this was changed.


mslacken (commented):
Add and use wwlog.Printf(wwlog.SECWARN,"...

Issue #251, opened 2022-01-20
Move IPMI configuration to its own key


jcsiadal:
IPMI configuration should be its own subkey in nodes.conf.

PR #354, opened 2022-04-04
YAML file update

jcsiadal:
This is the update for the YAML config files to make them less flat-file-like and to add a version control field. Providing the PR right now while I test it, so I can get input if changes are needed.

I want to use the version control field to provide seamless migration in future versions. In a nutshell: keep all old datastructures in the code; if the YAML fails to unmarshall, check the version, and update it if older. Alternately, just report that it's older and provide a command to update it to the newer format.

This will help meet requirements for seamless upgrades.

I deliberately made the version line ugly. It should be intuitive that it's not something admins should touch.

Updated YAML  hierarchy to remove anything with more than two "[prefix]name: value" lines, converting [prefix] into a parent group. Unfortunately, our code base is inefficient--instead of calling a common function, there are a lot of verbatim copies of the same code in different files. This really needs to be fixed, but for now, a minor update to a variable impacts many files.


mslacken (approved):

mslacken:
resolves #251

PR #358, opened 2022-04-06
Ignore dotfiles/hidden files when listing overlays.

griznog:
- Have unconfigured nodes retry booting without a full reset of the server to speed things up.
- Have unconfigured nodes retry booting without a full reset of the server to speed things up.
- Add check to exclude hidden files (dotfiles)


griznog:
I should really learn to use git. I'll resubmit this when I figure out how to drop all this history and the extraneous commit in the ipxe file.

Issue #366, opened 2022-04-07
Node add failure


MatthewHink:
This command line used to work. Failure may be due to the data structure changes?
It looks like Kernel isn't getting instantiated if it does not exist here:
https://github.com/hpcng/warewulf/blob/95050a44cebeb5fe90cb6ef3a2c3f7e24543b812/internal/pkg/node/modifiers.go#L63

```
[mhink@rocky warewulf]$ sudo wwctl node add n0000 --netdev eth0 --ipaddr 10.0.8.150  --hwaddr 08:00:27:39:46:70 --discoverable 
Added node: n0000
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x5c7b4f]

goroutine 1 [running]:
github.com/hpcng/warewulf/internal/pkg/node.(*nodeYaml).NodeUpdate(_, {{{0xc000495f50, 0x1, 0x1}, {0x0, 0x0, 0x0}, {0x0, 0x0}, {0x0, ...}}, ...})
	/home/mhink/go/src/github.com/hpcng/warewulf/internal/pkg/node/modifiers.go:63 +0x38f
github.com/hpcng/warewulf/internal/app/wwctl/node/add.CobraRunE(0x1587da0, {0xc000432f80, 0x1, 0x8})
	/home/mhink/go/src/github.com/hpcng/warewulf/internal/app/wwctl/node/add/main.go:148 +0x18c5
github.com/spf13/cobra.(*Command).execute(0x1587da0, {0xc000432f00, 0x8, 0x8})
	/home/mhink/go/src/github.com/hpcng/warewulf/vendor/github.com/spf13/cobra/command.go:850 +0x60e
github.com/spf13/cobra.(*Command).ExecuteC(0x1584920)
	/home/mhink/go/src/github.com/hpcng/warewulf/vendor/github.com/spf13/cobra/command.go:958 +0x3ad
github.com/spf13/cobra.(*Command).Execute(...)
	/home/mhink/go/src/github.com/hpcng/warewulf/vendor/github.com/spf13/cobra/command.go:895
main.main()
	/home/mhink/go/src/github.com/hpcng/warewulf/cmd/wwctl/main.go:13 +0x25
[mhink@rocky warewulf]$ 

```

mslacken:
Yes, the data structures are not initialized at this point. Thanks for getting this

Issue #365, opened 2022-04-07
unit tests fail


mslacken:
The unit tests fail on the actual branch in internal/pkg/node/util_test.go

MatthewHink:
Here is some more info. This looks related to the data structure changes.
- This is an access violation.
- n.Kernel is nil here: https://github.com/hpcng/warewulf/blob/95050a44cebeb5fe90cb6ef3a2c3f7e24543b812/internal/pkg/node/constructors.go#L70
- The failing test is using this configuration. Kernel does not appear to be defined: https://github.com/hpcng/warewulf/blob/95050a44cebeb5fe90cb6ef3a2c3f7e24543b812/internal/pkg/node/util_test.go#L10
```
[mhink@rocky warewulf]$ git log -n 1
commit 6c0fd068cc7118b8bf9a3f9e901b179a30474858 (HEAD -> main, origin/main, origin/HEAD)
Author: Christian Goll <cgoll@suse.de>
Date:   Thu Apr 7 17:22:38 2022 +0200

    updated Makefile for v4.3.0-rc1
[mhink@rocky warewulf]$ make test-it
go test -v ./...
?   	github.com/hpcng/warewulf/cmd/bash_completion	[no test files]
?   	github.com/hpcng/warewulf/cmd/config_defaults	[no test files]
?   	github.com/hpcng/warewulf/cmd/man_page	[no test files]
?   	github.com/hpcng/warewulf/cmd/wwclient	[no test files]
?   	github.com/hpcng/warewulf/cmd/wwctl	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwclient	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwctl	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwctl/configure	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwctl/configure/dhcp	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwctl/configure/nfs	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwctl/configure/ssh	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwctl/configure/tftp	[no test files]
?   	github.com/hpcng/warewulf/internal/app/wwctl/container	[no test files]
=== RUN   TestArgsContainerBuild
--- PASS: TestArgsContainerBuild (0.00s)
PASS
[...]

PR #368, opened 2022-04-07
Nil check node configuration pointers.

MatthewHink:
Fixes https://github.com/hpcng/warewulf/issues/366

The CI build will fail since I did not update the failing test yet.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/opencontainers/umoci v0.4.6
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.1.1
+	github.com/stretchr/testify v1.7.0
 	github.com/talos-systems/go-smbios v0.1.1
 	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1
 	gopkg.in/yaml.v2 v2.4.0
--- a/internal/pkg/node/modifiers.go
+++ b/internal/pkg/node/modifiers.go
@@ -61,14 +61,16 @@ func (config *nodeYaml) NodeUpdate(node NodeInfo) error {
 	config.Nodes[nodeID].ClusterName = node.ClusterName.GetReal()
 	config.Nodes[nodeID].Ipxe = node.Ipxe.GetReal()
 	config.Nodes[nodeID].Init = node.Init.GetReal()
-	if node.Kernel.Override.GotReal() || node.Kernel.Args.GotReal() {
+
+	if node.Kernel != nil && (node.Kernel.Override.GotReal() || node.Kernel.Args.GotReal()) {
 		config.Nodes[nodeID].Kernel = new(KernelConf)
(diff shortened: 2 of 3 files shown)


MatthewHink:
This may be similar to #365 I'm not sure yet.

To set this up I did:
```
git pull origin main
sudo mv  /usr/local/etc/warewulf/ /usr/local/etc/warewulf.prev
sudo make install
```
which gave me the node.conf in #366.

We didn't fail in the past because there were no pointers in the NodeConf structure here: 
https://github.com/hpcng/warewulf/blob/6c0fd068cc7118b8bf9a3f9e901b179a30474858/internal/pkg/node/datastructure.go#L28-L29

This PR adds a nil check and a regression unit test that fails in current mainline with this failure:
```
=== RUN   TestNodeUpdate
--- FAIL: TestNodeUpdate (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x5ac26f]

goroutine 6 [running]:
testing.tRunner.func1.2({0x5d0140, 0x77f6a0})
	/usr/local/go/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1212 +0x218
panic({0x5d0140, 0x77f6a0})
	/usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/hpcng/warewulf/internal/pkg/node.(*nodeYaml).NodeUpdate(_, {{{0xc000051b30, 0x1, 0x1}, {0x0, 0x0, 0x0}, {0x0, 0x0}, {0x0, ...}}, ...})
	/home/mhink/go/src/github.com/hpcng/warewulf/internal/pkg/node/modifiers.go:63 +0x38f
github.com/hpcng/warewulf/internal/pkg/node.TestNodeUpdate(0xc00008eea0)
	/home/mhink/go/src/github.com/hpcng/warewulf/internal/pkg/node/node_test.go:34 +0x185
testing.tRunner(0xc00008eea0, 0x6139a0)
	/usr/local/go/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1306 +0x35a
FAIL	github.com/hpcng/warewulf/internal/pkg/node	0.005s
```

mslacken:
Can you please check if this is needed withupdated main, there are also now `nil` checks involved

MatthewHink:
Looking at this now. We may no longer need this full PR, but the unit test is useful.

MatthewHink:
The unit test is failing because node.Kernel is nil.

MatthewHink:
Rebased this PR. The nil check passes the unit test.

mslacken (approved):
Great, so this test would have catched the memory bugs for the nodes. 
Do we need also similar tests for the profiles?

MatthewHink:
This test would have caught the failure. I did not investigate profiles, but probably should

PR #370, opened 2022-04-07
Vet and test in CI.

MatthewHink:
I think I need to open a PR to trigger tests (?)
We'll see if they run.


MatthewHink:
Tests are running in CI now.

PR #379, opened 2022-04-11
set default kernel.version to container name

mslacken:
Always set the default value of kernel.versionoverride to the name of the container, if this is a valid kernel. Also try to automatically import a kernel on container import.

In a normal workflow this will endup in just one single command which will import the container and its kernel. The kernel will be dereived from the container name, if this there is a kernel with the same name as the container.


MatthewHink (approved):

MatthewHink (approved):

mslacken:
Kernels need not to be set that way

PR #384, opened 2022-04-13
Fix uppercase MAC error

jcsiadal:
When set, node HW addresses are converted to lowercase. It doesn't affect unmarshalling.


mslacken (changes requested):
I think it is better to use
```
n.Netdevs[devname].Hwaddr.Set(strings.ToLower(netdev.Hwaddr))
```
instead

mslacken (approved):

PR #398, opened 2022-04-18
Fix host overlay support for static DHCP

jcsiadal:
Updates the dhcpd.conf.ww overlay file to include text from the static DHCP example file. It also fixes the struct variable name and changes the template to use the default NetDev instead of being hardcoded to "eth0".

If in warewulf.conf, dhcp -> template is set to "static", the resulting host overlay will be a static dhcpd.conf. If any other value is used, it will be the current dynamic dhcpd.conf.

I made a few changes to the control dashes to keep newlines consistent in the output file.


MatthewHink:
Static overlays did not work.

gmkurtzer (approved):
Looks good to me.

Issue #395, opened 2022-04-17
Cannot delete network name in a profile


jcsiadal:
Cannot determine a way to remove the default network using wwctl.
```
    network devices:
      default: {}
      eth0:
        type: ethernet
        ...
```

The obvious way produces an error, which appears to be a bug.
```
[root@frontend ~]# wwctl profile set default --netdel --netname default
[ERROR]      : You must include the '--netname' option
```

mslacken:
@jcsiadal can you please check #405 fixes this issue

jcsiadal:
#400 fixes this issue, as well as a few others. I don't have permission to manually link issues to PRs.

Issue #392, opened 2022-04-14
Add ability to set primary network device in a profile.


jcsiadal:
The ability to set the primary network device is currently supported in "node set", but not "profile set". It can be added by manually editing nodes.conf, but not through wwctl.

Per #391, the current option "--default" needs to be changed to another term, such as "--primary".

anderbubble:
This appears to be resolved now by way of the unification of nodes and profiles.

Issue #391, opened 2022-04-14
Change "default" network device to another term.


jcsiadal:
We're overloading the term "default" in the configuration, and this can be confusing, especially when trying to discuss an issue.

Default means failure, fail-over, or fall-back.  In computer terminology, "a preselected option adopted by a computer program or other mechanism when no alternative is specified by the user or programmer"

For example, the default boot network would be the last one tried, if all of the others failed. We currently define it as the opposite. This:
```
    network devices:
      eth0:
        default: "true"
```
Needs to be changed to something like this:
```
    network devices:
      eth0:
        primary: "true"
```

mslacken:
was fixed in 4.3.0

PR #400, opened 2022-04-19
Fix wwctl bug deleting networks and ability to set default network

jcsiadal:

Fixes issues #395, #391, and #392 

- Updates the netdel argument to use the network name instead of the network device, which caused the command to fail.
- Enables an option to set the default network in a profile.
- Fixes the boolean set method in the node datastructure. Previously, a false value would do nothing, so a "true" value couldn't be unset
- Changes the main network device identifier from "default" to "primary". Default is defined as fallback or failover, so term was used incorrectly here (in addition to being overloaded).

### DIFF
--- a/internal/app/wwctl/profile/set/root.go
+++ b/internal/app/wwctl/profile/set/root.go
@@ -118,6 +118,7 @@ func init() {
 
 	baseCmd.PersistentFlags().StringVarP(&SetNetName, "netname", "n", "default", "Define the network name to configure")
 	baseCmd.PersistentFlags().StringVarP(&SetNetDev, "netdev", "N", "", "Set the node's network device")
+	baseCmd.PersistentFlags().StringVar(&SetNetDefault, "primary", "", "Enable/disable device as primary (yes/no)")
 	baseCmd.PersistentFlags().StringVarP(&SetNetmask, "netmask", "M", "", "Set the node's network device netmask")
 	baseCmd.PersistentFlags().StringVarP(&SetGateway, "gateway", "G", "", "Set the node's network device gateway")
 	baseCmd.PersistentFlags().StringVarP(&SetType, "type", "T", "", "Set the node's network device type")
--- a/internal/pkg/node/datastructure.go
+++ b/internal/pkg/node/datastructure.go
@@ -75,7 +75,7 @@ type NetDevs struct {
 	Prefix  string            `yaml:"prefix,omitempty"`
 	Netmask string            `yaml:"netmask,omitempty"`
 	Gateway string            `yaml:"gateway,omitempty"`
-	Default string            `yaml:"default,omitempty"`
+	Default string            `yaml:"primary,omitempty"`
 	Tags    map[string]string `yaml:"tags,omitempty"`
 }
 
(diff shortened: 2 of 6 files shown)


MatthewHink:
Main thing here is that you could not delete networks from profiles. (netdev.device to netdev.name).
Could not set primary device in a profile.
Could not unset device in a profile. Changing to false would ignore the value and move on.
Default and primary are different terms.

internal/app/wwctl/profile/set/root.go
```diff
@@ -118,6 +118,7 @@ func init() {
 
 	baseCmd.PersistentFlags().StringVarP(&SetNetName, "netname", "n", "default", "Define the network name to configure")
 	baseCmd.PersistentFlags().StringVarP(&SetNetDev, "netdev", "N", "", "Set the node's network device")
+	baseCmd.PersistentFlags().StringVar(&SetNetDefault, "primary", "", "Enable/disable device as primary (yes/no)")
```
gmkurtzer, line 121: I don't believe this is correct. It should still be `--default` not primary, as this will change the NetDev.Default struct member.

internal/pkg/node/datastructure.go
```diff
@@ -75,7 +75,7 @@ type NetDevs struct {
 	Prefix  string            `yaml:"prefix,omitempty"`
 	Netmask string            `yaml:"netmask,omitempty"`
 	Gateway string            `yaml:"gateway,omitempty"`
-	Default string            `yaml:"default,omitempty"`
+	Default string            `yaml:"primary,omitempty"`
```
gmkurtzer, line 78: This should also be default, not primary. We should only changing the name of the device to "default".

gmkurtzer (changes requested):
A couple of change requests on backend datastore as it should still be "default" as that pertains to the network configuration itself.

internal/pkg/node/datastructure.go
jcsiadal: It should be "primary" or something similar. It's neither the default network nor device, so using default is incorrect. The internal structs also should be changed, but I was leaving them intact or now because this involves a much more significant change and regular users won't see them.

internal/app/wwctl/profile/set/root.go
jcsiadal: I didn't change the internal struct member, as regular users don't see or use it. This changes only the command option and YAML tag.
 
The internal struct should also be updated, but it will be lot more changes and I wanted to wait until the new API is implemented.

gmkurtzer: Sorry, that's not what I meant.

the `SetNetDefault` variable triggers the `DEFAULT=...` in the ifcfg-* templates. Meaning, that has nothing to do with "primary".

internal/pkg/node/datastructure.go
gmkurtzer: But that's my point, we should't be changing "default" to "primary" for this, as that is the name of the specific configuration that we have in the templates. Or we have to change it there too, but I think there is precedent there already not to change that.

gmkurtzer:
It took me a bit to put my finger on what is confusing here, and I figured it out....

We are now overloading the term "primary", but in terms of nomenclature as well as an object member to also define it as "default" (also conflated with "primary" now). I'm going to leave it for now as you have it, but this is a very confusing redundancy which needs to be cleaned up in the next release.

gmkurtzer (approved):

PR #405, opened 2022-04-26
fix deletion of network for profile

### LINKED ISSUE #395: Cannot delete network name in a profile

mslacken:
Should fix #395


jcsiadal:
I already submitted a PR for this two weeks ago.

mslacken:
r emvoed in favor primary

Issue #394, opened 2022-04-16
is the quickstart missleading?


MostafaMamouni:
In this part of the quickstart https://warewulf.org/docs/quickstart/el8#set-up-the-default-node-profile, there is that mention of `--netname` but when I tried it, `wwctl` raises `ERROR: unknown flag: --netname`,
Thank you.

MostafaMamouni:
the docs too, sometimes the sysconf dir`warewulf` is created in in `usr/local/` and not in /etc itself,
in https://warewulf.org/docs/enchiridion/nodeconfig#configuring-the-nodes-kernel, this command
 ```
sudo wwctl node set --kernel $(uname -r) n0000
```
raises `ERROR: unknown flag: --kernel`,
also, I couldn't find the `/usr/local/etc/warewulf/templates` directory that is mentioned in the docs,  `/usr/local/etc/warewulf, is it important to get werewulf working.
Thank you

PR #406, opened 2022-04-26
more verbose wwctl version

mslacken:
`wwctl version` will print out all complied in vars. This will make debugging for users much easier, e.g #394


mslacken:
Should go into 4.3.0

gmkurtzer:
I completely agree that this is necessary and valuable information, but shouldn't the `version` output be just the single line output of the version?

tcooper:
I agree that `wwctl version` should only print the version and not require more complicated parsing. If you want the build configuration perhaps we want `wwctl buildcfg` as a command?

gmkurtzer:
Looks good to me, thanks!

PR #411, opened 2022-05-03
syncuser will sync users now

mslacken:


gmkurtzer:
I believe this is no longer needed due to the changes I made in #415.

PR #414, opened 2022-05-03
Use SetNetName instead of SetNetDev when performing netdel

kcdodd:
I wasn't sure whether 'netdev' or 'netname' was intended here, but  went with netname.


gmkurtzer:
Great catch, thanks!

PR #415, opened 2022-05-04
Minor fix ups for 4.3.0

gmkurtzer:
Some minor fix ups in my initial testing of 4.3.0.


gmkurtzer:
These fixes will render #411 obsolete as the syncuids option is no longer default, and has to be requested explicitly.

PR #419, opened 2022-05-05
Renames instances of "Default" to "Primary" in netdev source code

jcsiadal:

Within netdev structs, Default is renamed to Primary. Cobra commander variables using "default" for the primary network are also renamed.

### DIFF
--- a/cmd/update_configuration/vers43/datastructure.go
+++ b/cmd/update_configuration/vers43/datastructure.go
@@ -60,6 +60,6 @@ type NetDevs struct {
 	Prefix  string            `yaml:"prefix,omitempty"`
 	Netmask string            `yaml:"netmask,omitempty"`
 	Gateway string            `yaml:"gateway,omitempty"`
-	Default string            `yaml:"default,omitempty"`
+	Primary string            `yaml:"primary,omitempty"`
 	Tags    map[string]string `yaml:"tags,omitempty"`
 }
--- a/cmd/update_configuration/update_configuration.go
+++ b/cmd/update_configuration/update_configuration.go
@@ -104,7 +104,7 @@ func update42to43(conf42 vers42.NodeConf) vers43.NodeConf {
 		var device vers43.NetDevs = vers43.NetDevs{
 			Type:    netdev.Type,
 			Device:  devn,
-			Default: printB(netdev.Default),
+			Primary: printB(netdev.Default),
 			Hwaddr:  netdev.Hwaddr,
 			Ipaddr:  netdev.Ipaddr,
 			IpCIDR:  netdev.IpCIDR,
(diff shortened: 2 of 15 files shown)


cmd/update_configuration/vers43/datastructure.go
```diff
@@ -60,6 +60,6 @@ type NetDevs struct {
 	Prefix  string            `yaml:"prefix,omitempty"`
 	Netmask string            `yaml:"netmask,omitempty"`
 	Gateway string            `yaml:"gateway,omitempty"`
-	Default string            `yaml:"default,omitempty"`
```
mslacken, line 63: We should leave out this config string as this will render old configurations broken. So we should read in this value, but store it in as Primary and so not save it.

jcsiadal: I haven't looked too much into the update_configuration tool. Should I just delete that one line?

mslacken: They need not match exactly.  There are two data definition in warewulf, which is the `nodeYaml`, which comes an goes to disk, and the datastore, which is of type `NodeInfo`. At the start the `nodeYaml` is parsed and all the datastore of is populated with it. If the storage of `nodes.con` is triggered, a new `nodeYaml` is created from the values of the datastore. So we will have to read in `Default` and use its value for the initialization of the Primay element in the datastore constructor https://github.com/hpcng/warewulf/blob/3edb8d5294f7495ebb04a56adbcd1b02d8d7895e/internal/pkg/node/constructors.go#L52
but leave it out in the code which stores the datastore in https://github.com/hpcng/warewulf/blob/3edb8d5294f7495ebb04a56adbcd1b02d8d7895e/internal/pkg/node/modifiers.go#L52

jcsiadal: Actually, let me restate this. To update the file, we are concerned only with the structure of nodes.conf (nodeYaml). If the file is older, you read it using the correct older datastructure, modify it (looping through each intervening version), and write it using the current datastructure. Then you go back to normal operation and read nodes.conf again.
There should never be another datastructure defined for the current release. That's already defined in the nodes package.
Older datastructures should be full copies that shipped with those releases.

jcsiadal: The file cmd/update_configuration/vers43/datastructure.go should not exist. We need to load the output datastructure directly from internal/pkg/node.
Anytime we need to keep two structs in sync manually, we probably doing it wrong.

mslacken: > The file cmd/update_configuration/vers43/datastructure.go should not exist. 
We can remove it, as the resulting tool is not in the `all` target of the make, just in the update config, but still it has documentation purpose.

PR #423, opened 2022-05-14
Refactor of server, logging, building of images, template updates 

### LINKED ISSUE #346: Container import generates empty image on Arch
When I run the container import command it completes without error but the resulting container image is empty. I discovered that that the [find command](https://github.com/hpcng/warewulf/blob/main/internal/pkg/container/build.go#L55) was failing but the error wasn't being reported. It looks like find doesn't recognize the `` -xautofs `` and with more research it doesn't seem to exist in the most recent version of [gnu find](https://www.gnu.org/software/findutils/manual/find.pdf). My find version of find is `` find (GNU findutils) 4.9.0 `` btw.

kcdodd:
fixes #346

**Refactors**
* Combine server response handling, reduce code duplication
* Simplify stage cases, backward compatible uri
* Implement image creation as functions instead of shell scripts
* Combine daemonLogf and wwlog to single logging output
* Switch to named wwlog functions instead of constants

**Fixes**
* Fix hidden errors occurring in shell pipes
* Fix kernel search path for Ubuntu
* Fix http response when template rendering fails
* Fix template newline insertion, update dhcp and hosts tmpl

**Updates**
* Add hwaddr and interface to dhcp
* Add control over image compression to client
* Update IPXE template to handle non-compressed files
* Update DHCP template to set network interfaces and server IP assignment
* Update DHCP/hosts templates to choose a host-name self-consistently

**Possible Breaking Changes**
* "excludes" for container import would now be passed to ``filepath.Match`` instead of ``grep``
* ``Stat_t.Dev`` is used instead of ``-xdev`` flag to ``find``
* Consequence of ``-xautofs`` not known, is not available in all implementations of ``find``, and how this is different than "normal" cross-device mounts. My guess would be that giving ``-xdev`` would already exclude autofs mounts, but I am not familiar with this feature.

**Summary**

This request includes quite a few changes, and it is understood that the consequences of some of the changes may not be fully known, and there may be some discussion. 
This refactoring initially started because Warewulf did not report any errors, yet the test node was unable to boot with immediate kernel panic.
Investigating the source code initially did not reveal why these problems happened, and the process of debugging lead to opportunities to try to reduce code duplication and enhance logging (in addition to fixing the problem).
The initial problem was eventually narrowed to a combination of silent errors happening in the shell pipes during container image creation.
The ``find`` command in the version of Ubuntu I am using does *not* have an ``-xautofs`` option, which immediately failed and piped an empty string to the ``cpio`` command, since these were being executed as a shell script.
Instead, these were separated out and implemented natively where possible to track exactly what it was doing and catching any errors.


mslacken:
Thanks a lot for your contribution. I will review it as soon as I have some time and as a lot of changes are inside, the pr won't go into the 4.3.0 release, as it touches a lot of areas.
You also might want to join the slack channel for further discussion.

internal/pkg/warewulfconf/datastructure.go
```diff
@@ -38,6 +38,10 @@ type DhcpConf struct {
 	RangeStart  string `yaml:"range start" default:"192.168.200.50"`
 	RangeEnd    string `yaml:"range end" default:"192.168.200.99"`
 	SystemdName string `yaml:"systemd name" default:"dhcpd"`
+	Syslog      string `yaml:"syslog" default:""`
+	Syslogs     string `yaml:"syslogs" default:""`
+	Interface   string `yaml:"interface" default:""`
```
jcsiadal, line 43: What's the rationale here for a separate interface element? Why would an interface set here be different from the interface set by the IPaddr or IPaddr6 elements?

jcsiadal, line 43: What's the rationale here for a separate interface element? Why would an interface set here be different from the interface set by the IPaddr or IPaddr6 elements?

griznog:
I think the network stuff should be nested under a "networks" element, we shouldn't be restricted to just serving a single network. It'd also make sense to have it somewhere that a `network` could be used in an overlay templates and in node configs (not sure if anything in `/etc/warewulf.conf` is visible to overlay templates?) It's possible to multihome the warewulf master now, I currently have 4 networks on my production setup. It'd just be nice to have all those networks be first class warewulf config citizens.

internal/pkg/warewulfconf/datastructure.go
kcdodd: The intention was to use this option to set the dhcp config option 'INTERFACESv4' for it to listen on, since there is likely more than one (e.g. one connected to the node network and one connected to the 'external' LAN). The Hwaddr here would also be used to allow a host overlay to configure it, and assign the Ipaddr/Ipaddr6 to the correct interface.

kcdodd:
If I understand correctly, it seems like that would also replace the need for a 'Netmask' and 'Gateway' per each node network device, if instead each device was assigned a network listed in the main config. It also seems like the host should have a list of 'NetDevs' like nodes/profiles to be more uniform, replacing a need for any more options to 'dhcp'.

jcsiadal:
I agree on the need for provisioning support on multiple networks. I think we should add this in a future release. I prefer defining all networks separately. Then you just need to define Device X with IP Y for each node. You don't even need to specify the network since it's clear where the interface belongs.

kcdodd:
I should probably remove the network related changes from this pr.

griznog:
FWIW I already keep a  config block in `nodes.conf` for my master node(s) to use in overlays by treating the master as just another node and assigning a subset of overlays to it for things that make sense to manage that way. It's a shame this config file is named `nodes.conf`, it seems like it'd be simple enough to have `networks` be a top level item like `nodeprofiles` and `nodes`, but that clearly overloads the function of this file implied by the file name.

mslacken (approved):
Looks good to me

gmkurtzer:
This looks really amazing! I found a few minor fixes in #428, but nothing major.

The one question I have, which is not a blocker, is the move to uncompressed Vs. compressed iPXE layers. The kernel can decompress the layers just fine, so sending uncompressed layers over the wire will just result in more bandwidth usage which means we also don't need to persist the uncompressed layers, nor do we need to fail over from `imgextract` then.

So before I change some of that back, I was curious if I'm missing something here.

Thanks again, really fantastic work!

kcdodd:
I was not sure if it was necessary, and its probably the right thing to put it back. The reason I thought it might be necessary was from reading of the iPXE documentation and some apparent problems people were having that lead to using 'imgextract' in the first place, and seemed like the only sure-fire way was to just send the uncompressed version to take the GZIP burden off of iPXE and the kernel. I was aware that the kernel itself is self-extracting, but it seemed ambiguous what exactly the the `initrd` command does with compressed images. It sounded from the docs that iPXE expected usage is to use `imgextract` for gzip-compressed images, otherwise the `initrd` command expects a non-compressed image.

https://ipxe.org/cmd/imgfetch?redirect=1
>There is no difference between the commands imgfetch, module, and initrd. [omitted]...then the file initrd.img will be treated as a normal initramfs filesystem image

https://ipxe.org/cmd/imgextract
> **Download and extract a gzip-compressed initrd image**
>```
>  imgextract http://192.168.0.1/images/initrd.gz
>```
> **Notes**
> You can use the [chain](https://ipxe.org/cmd/chain) or [kernel](https://ipxe.org/cmd/kernel) commands to execute a single-member archive image directly without using the imgextract command. For example, if you have a GZIP-compressed kernel image vmlinuz.gz:
>```
>  #!ipxe
>  kernel http://192.168.0.1/images/vmlinuz.gz
>  initrd http://192.168.0.1/images/initrd.img
>  boot
>```

gmkurtzer:
@kcdodd Got it, also in our community meeting today, @griznog mentioned another reason to do this, which has to do with the kernel having a 4GB limit on uncompressing zlib data, so this gets around that.

Great work, and thank you!

PR #429, opened 2022-05-25
add commad for overlay rendering

mslacken:
added the wwctl overlay parse command

The `wwctl overlay parse -n NODE FILE` will interpret the given overlay file
and print it stdout, via wwlog.PRINTF.  The overlay code had to be refactored
for this, but hopefully this makes the code a bit more readable. Also the
 tstruct has not its own initilization code.


mslacken:
With the `--quiet` option nice hacks like
```
cat > ping.ww <<EOF
{{range $node := $.AllNodes}}
ping -c 1 {{$node.Id.Get}}
{{- end}}{{/* end for each node */}}
EOF
wwctl overlay parse -q | bash
```
are possible.

gmkurtzer:
Super cool! Can we make it so it supports this as well?

```
$ sudo wwctl overlay edit [overlay name] [overlay file]
$ sudo wwctl overlay show --render=$NODENAME [overlay name] [overlay file]
```

PR #432, opened 2022-05-26
4.3.0rc6 fixes

gmkurtzer:
DO NOT MERGE YET!

A minor fix/workaround to always use compression for what Norm Gaywood found in Slack with regards to the compute nodes running out of memory while trying to fetch the container images.

Also, I included an optimization in the `sendFile()` function to not load full binary data blobs into memory before sending over the the HTTP writer as doing that in parallel would kill a control server.

note: We might decide a different/better fix for the first issue and I'd like validation on the second.

### DIFF
--- a/etc/ipxe/default.ipxe
+++ b/etc/ipxe/default.ipxe
@@ -19,40 +19,41 @@ echo Warewulf Controller: {{.Ipaddr}}
 echo Downloading Kernel Image:
 kernel --name kernel ${uri_base}&stage=kernel       || goto reboot
 
-## try extracting compressed images first
-## NOTE: system overlay tends to be the smallest, so failure here is the cheapest
-#echo Downloading Container Image:
-#imgextract --name container ${uri_base}&stage=container&compress=gz || goto nocompress
-#
-#echo Downloading System Overlay:
-#imgextract --name system ${uri_base}&stage=system&compress=gz       || goto reboot
-#
-#echo Downloading Runtime Overlay:
-#imgextract --name runtime ${uri_base}&stage=runtime&compress=gz     || goto reboot
-#
-#{{if ne .KernelOverride "" -}}
-#echo Downloading Kernel Modules:
-#imgextract --name kmods ${uri_base}&stage=kmods&compress=gz         || goto reboot
-#{{- end}}
-#
-#goto imoktogo
-#
-#:nocompress
-#
-#echo Extraction failed, attempting un-compressed version of images
-
+# try extracting compressed images first
+# NOTE: system overlay tends to be the smallest, so failure here is the cheapest
 echo Downloading Container Image:
-initrd --name container ${uri_base}&stage=container     || goto reboot
+imgextract --name container ${uri_base}&stage=container&compress=gz || goto nocompress
 
 echo Downloading System Overlay:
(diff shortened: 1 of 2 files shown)


ngaywood:
One comment, the compute had 700GiB of main memory. So I doubt it was *that* memory running out
rpviewer
.

ngaywood:
The container size:
# ll /srv/warewulf/container/
total 1984340
-rw-r--r--. 1 root root 1377561088 May 26 14:59 miller-rocky-8.img
-rw-r--r--. 1 root root  654397503 May 26 14:59 miller-rocky-8.img.gz

ngaywood:
There are the changes I made to make the node boot:
 ```
# diff /root/default.ipxe /usr/local/etc/warewulf/ipxe/default.ipxe 
45c45
< initrd --name container ${uri_base}&stage=container     || goto reboot
---
> initrd --name container ${uri_base}&stage=container&compress=gz     || goto reboot
48c48
< initrd --name system ${uri_base}&stage=system           || goto reboot
---
> initrd --name system ${uri_base}&stage=system&compress=gz           || goto reboot
51c51
< initrd --name runtime ${uri_base}&stage=runtime         || goto reboot
---
> initrd --name runtime ${uri_base}&stage=runtime&compress=gz         || goto reboot
```

gmkurtzer:
And to be clear, sending a compressed image didn't trigger this failure (`&compress=gz` to the iPXE template GET request worked).

kcdodd:
I am looking into best method to serve the large files with minimal memory. That was a dumb thing on my part, looking at the go source where it failed, and it is even worse since that buffer was dynamically grown by 2x which means it could end up twice the size of the file being sent. Will prototype a chunked transfer method.

mslacken:
Unfortunately I have no experience with the buffer and network send file semantic. So I can't really review this,

mslacken (commented):
Looks good to me, but I have no experience in the network/send buffer semantic, so my review is not helpful here.

kcdodd:
I have been meaning to follow up on the file transfer, but I haven't had time to do the benchmarks and testing of ipxe http capabilities. However, I came across the standard library function `http.ServeContent` (https://pkg.go.dev/net/http#ServeContent), which is what the http.ServeFile uses internally, and have an example (untested) commit that uses it kcdodd/warewulf@643b76e68b8a2db32aa3e6a936bb64c340526d11.

I also started to think there may be a concurrency issue with dynamically building overlays and also FindDiscoverableNode. I don't see any mutex/locks for these operations, and if multiple nodes make requests at the same time there could be race conditions that corrupt the database or the dynamically rebuilt images. (thoughts?)

kcdodd:
I have added an example commit on what I was thinking about the discoverable nodes kcdodd/warewulf@9ff83ddf1e99913cdf593308612d7fd020301ee1.

But I take back the concurrency issue I thought for the overlays, since there is a dedicated output directory for each node.  The only common operation is MkdirAll, which I think is an atom operation (or equivalent).

gmkurtzer:
@kcdodd Regarding the chunking, the io.copy() automatically chunks it and sends over the HTTP writer FD in pieces.

The discoverable mutex lock is a good catch, please submit a subsequent PR for it.

Thanks!

PR #434, opened 2022-05-27
4.3.0 pxe update

gmkurtzer:
This updates iPXE files within the build


ipxe-update.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
```
mslacken, line 1: For 4.4 we should try to get this file removed and put the stuff in the central makefile.
So that we have something like `make ipxebinaries`

internal/pkg/configure/tftp.go
```diff
@@ -27,7 +27,7 @@ func TFTP() error {
 	}
 
 	fmt.Printf("Writing PXE files to: %s\n", tftpdir)
-	for _, f := range [4]string{"x86.efi", "i386.efi", "i386.kpxe", "arm64.efi"} {
+	for _, f := range [4]string{"x86_64.efi", "x86_64.kpxe", "arm64.efi"} {
 		err = util.SafeCopyFile(path.Join(buildconfig.DATADIR(), "warewulf", "ipxe", f), path.Join(tftpdir, f))
```
mslacken, line 31: Can we just not copy all files from `$(WWDATADIR)/ipxe/` to the right place, instead of having a explicit file list like we have now?

mslacken (approved):

ipxe-update.sh
griznog: @gmkurtzer Copying this here from slack, the sed commands weren't working for me.

```
sed -i.bak \
    -e 's,//\(#define.*CONSOLE_SERIAL.*\),\1,' \
    -e 's,//\(#define.*CONSOLE_FRAMEBUFFER.*\),\1,' \
    config/console.h

sed -i.bak \
    -e 's,//\(#define.*IMAGE_ZLIB.*\),\1,' \
    -e 's,//\(#define.*IMAGE_GZIP.*\),\1,' \
    config/general.h
```

gmkurtzer: This fix has been incorporated, thanks!

internal/pkg/configure/tftp.go
gmkurtzer: Yep, totally agree. Do you want to do that when you move the ipxe-update.sh into the Makefile? ... BTW, I'm happy to do that too, just let me know.

Issue #437, opened 2022-06-01
No file locking on nodes.conf 


jmechalas:
Writes made to nodes.conf via wwctl do not use any form of file locking. Further, when the file is opened for writing, it is opened with O_TRUNC which zeros the file out before any writes occur. This creates several possible race conditions:

1. Two people running wwctl at the same time will stomp on each other.
2. Having the file open in an editor will potentially stomp on changes made by wwctl.
3. Opening with O_TRUNC presents a read race condition where a simultaneous process will see an empty nodes.conf.
4. The write to nodes.conf is not atomic, so an abnormal exit of wwctl (^C or a stray signal from, say, system shutdown) could result in an empty or partial file

I suggest doing the following:

- Implement advisory locking using fcntl(2)
- Write to a temporary file alongside the original (e.g. nodes.conf.tmp) rather than writing directly to nodes.conf, and use rename(2) to move it into place. rename(2) is an atomic filesystem operation.
- Create a wrapper script for opening nodes.conf in an editor that makes use of this file locking (similar to what vipw does) and temporary file. This wrapper could also run a syntax check before committing the file, thus reducing the odds of an accident due to a manual edit.

griznog:
Having a flat file or files is a huge positive feature for many reasons not the least of those being that `svn commit -m "blah blah"` or the `git` equivalent nicely backs up my entire warewulf config in a way that requires nothing but a simple checkout to recover or roll back or use to build my next warewulf master. A database backend is the kind of complexity creep/bloat that should be avoided at all costs, didn't we try that in previous versions of WW and ultimately abandon it?  It also has impact on the ease of quick-starting a cluster for someone without database experience, for example, the MySQL step of setting up Slurm remains (for me) the most annoying aspect of initial setup and crippling single point of failure for Slurm "HA". 

[...]

PR #438, opened 2022-06-03
Fix server thread safety of discoverable node and overlay build

kcdodd:
Fixes

* Thread safety of configuring discoverable nodes and overlay builds
* Update to `wwctl overlay build` command in handling of lists of overlay names.

### DIFF
--- a/internal/pkg/warewulfd/util.go
+++ b/internal/pkg/warewulfd/util.go
@@ -4,6 +4,9 @@ import (
 	"net/http"
 	"os"
 
+	nodepkg "github.com/hpcng/warewulf/internal/pkg/node"
+	"github.com/hpcng/warewulf/internal/pkg/overlay"
+	"github.com/hpcng/warewulf/internal/pkg/util"
 	"github.com/hpcng/warewulf/internal/pkg/wwlog"
 )
 
@@ -36,4 +39,44 @@ func sendFile(
 	wwlog.Send("%15s: %s", sendto, filename)
 
 	return nil
+}
+
+func getOverlayFile(
+	nodeId string,
+	stage_overlays []string,
+	autobuild bool ) (stage_file string, err error) {
+
+	stage_file = overlay.OverlayImage(nodeId, stage_overlays)
+	err = nil
+
+	build := !util.IsFile(stage_file)
+
+	if !build && autobuild {
+		build = util.PathIsNewer(stage_file, nodepkg.ConfigFile)
+
+		for _, overlayname := range stage_overlays {
+			build = build || util.PathIsNewer(stage_file, overlay.OverlaySourceDir(overlayname))
+		}
+	}
+
+	if build {
+		wwlog.Serv("BUILD %15s, overlays %v", nodeId, stage_overlays)
+
+		args := []string{"overlay", "build"}
+
+		for _, overlayname := range stage_overlays {
+			args = append(args, "-O", overlayname)
+		}
+
+		args = append(args, nodeId)
+
+		out, err := util.RunWWCTL(args...)
+
+		if err != nil {
+			wwlog.Error("Failed to build overlay: %s, %s, %s\n%s",
+				nodeId, stage_overlays, stage_file, string(out))
+		}
+	}
+
+	return
 }
--- a/internal/pkg/warewulfd/util.go
+++ b/internal/pkg/warewulfd/util.go
@@ -1,16 +1,18 @@
 package warewulfd
 
 import (
-	"io"
 	"net/http"
 	"os"
-	"strconv"
 
 	"github.com/hpcng/warewulf/internal/pkg/wwlog"
-	"github.com/pkg/errors"
 )
 
-func sendFile(w http.ResponseWriter, filename string, sendto string) error {
+func sendFile(
+	w http.ResponseWriter,
+	req *http.Request,
+	filename string,
+	sendto string) error {
+
 	fd, err := os.Open(filename)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


gmkurtzer:
This is awesome, let me know when you are ready for review and testing.

kcdodd:
@gmkurtzer Unfortunately, I am going to have to un-take-back my concern over dynamically building overlays. It turns out building overlays is **not** thread safe because it requires changing the current working directory, and `os.Chdir` is a process-wide operation. If one thread is busy building an overlay, and another request comes in, the new thread will Chdir to somewhere else while the first one is still trying to process the files it assumes are relative to wherever if Chdir'd to, likely crashing the first request's thread. I do not have a good solution for this right now, but it almost seems that a sub-process should be spawned the does the re-building, which would circumvent the issue without needing to block other requests from also rebuilding overlays.

griznog:
What's the goal/benefit to building these dynamically? Seems like just rebuilding on a detected change would be sufficient. As long as the new/updated image is put in place with an atomic `mv`, then any processes that have the old image/overlay open will complete using the old copy and any new processes that `open()` will see the new image. `inotify` and events produced by `wwctl` would seem to cover most/all the cases that would trigger a rebuild, even manual edits. There'd be a possible issue with an incoming stream of changes that are faster than the rebuild time-to-completion, but should be straightforward to catch those changes in a fifo queue and just trigger the next rebuild flushing the queued events it covers. 

One option for scaling up WW is to use simple caching http(s) proxies in front of it. It'd be nice if those caches could quickly use a HEAD request to (in)validate the cache. I don't know if dynamic generation affects that, but if so I'd consider that a big downside. 

griznog

kcdodd:
Here are my thoughts, not to be taken as a conclusion on the best approach.

I think using inotify to monitor for changes would be a good optimization (instead of checking the modtime of every overlay file on every request), but there are some considerations:

* The dependency resolution `[node]->[node overlays]->[overlay files]` (a 1:1 mapping per rebuild) is changed to `[overlay files]->[node overlays]->[nodes]` (a 1:N mapping, N=# of nodes). I.E. find all nodes that are affected by a change to a file, and rebuild those images.
* All affected node images would be rebuilt whether or not the nodes are actually running.
* Needs initialization upon restarting the server to catch changes that may have happened when it wasn't watching, probably very similar to the current logic of checking the modtime of every node image to that of every overlay file.
* The http response handling can still be updated either way using http.ServeContent instead of io.Copy to respond to HEAD requests, or If-Modified-Since, to further optimize the network traffic, since this is only taken from the on-disk image without additional logic.

However, the issue still stands that triggered rebuilds either have to be serialized (one node at a time), or spawned to a dedicated process for each node's overlay image.

kcdodd:
I'm sorry about all the fixup commits, I'm using a Mac at home and can't compile.

At least for now, I have added a spawn of `wwctl overlay build`, and updated the CLI to pass the list of overlays to build. I was a little confused at first because it didn't appear to accept a list, but later realized it did split the overlay name using a comma separated list when building the host overlay (but not the others?) That's why I changed the CLI argument type, and I made it backward compatible for that case, but if that should be reverted to be only comma separated I can do that.

kcdodd:
I have run the concurrency test, and appears to be working as expected, so am marking ready for review.

For the interested, I am attaching log files of the test script side and the warewulfd side of things. Test scripts can be made available.

In summary, the test was setup to have 30 processes request each stage simultaneously (synced using MPI on the test script side), and warewulf configured to have only 25 discoverable nodes (meaning the same 5 of the processes should get 400 status codes on all requests). The node hwaddr's are all dummy and randomly generated. For this test, it was limited to spawn up to 16 sub-processes at a time, but that figure will depend on the number of CPUs on the host machine.

[test_warewulfd.log](https://github.com/hpcng/warewulf/files/8848691/test_warewulfd.log)
[test.log](https://github.com/hpcng/warewulf/files/8848693/test.log)

kcdodd:
Absolutely! These changes are **only** made to achieve thread safety of the warewulfd server, not general filesystem safety. The fact that the node config is written back to disk is actually not the factor here, other than what you mention that it just opens it up to more concurrency issues at the fs level. Even without the filesystem issues, the logic used in the server depends on the state of "db", so if one thread is to decide to change the very value that decision is made upon, then it minimally has to keep an RW lock until after it acts on the decision (not separately on the read and write) to prevent another thread from swooping in and grabbing it.

As pointed out in #437, there are many ways in which the nodes.conf file can run into contention even without the server.

gmkurtzer (approved):
This looks good to me, and passing my local tests; granted I don't have thousands of nodes to try, but I'm good with it!

internal/pkg/warewulfd/util.go
```diff
@@ -1,51 +1,82 @@
...
+
+		args := []string{"overlay", "build"}
+
+		for _, overlayname := range stage_overlays {
+			args = append(args, "-O", overlayname)
+		}
+
+		args = append(args, nodeId)
+
+		out, err := util.RunWWCTL(args...)
```
mslacken, line 73: I think in the future we should try to call  `overlay.BuildOverlayIndir(nodeId, stage_overlays, OverlayDir)` directly. The _OverlayDir_ is technical dept which we/I have to fix. With #429 there is a start of refactoring of the overlay handling and get rid of _ChDir_ which triggered this.
But given the late stage of development lets go with this.

mslacken (approved):

gmkurtzer:
Thank you @kcdodd, this is great!

Issue #319, opened 2022-02-24
`wwctl configure tftp` enables tftp.socket with `enabled: false` set in `warewulf.conf`


griznog:
With this in `/etc/warewulf.conf`:

```
tftp:
  enabled: false
  systemd name: tftp
```
I expected `wwctl configure tftp` to set up the files in `/var/lib/tftpboot/warewulf` but not enable the `tftp` socket which on my system is already in use by `dnsmasq`.  When I run the command, this happens:

```
[root@lepomis warewulf]# wwctl configure tftp
Enabling and restarting the TFTP services
Created symlink /etc/systemd/system/sockets.target.wants/tftp.socket → /usr/lib/systemd/system/tftp.socket.
```

aspfate:
This issue was fixed in #440.
You can close this issue. @griznog

PR #440, opened 2022-06-03
can disable tftp services with warewulf.conf

aspfate:
fix issue #319 
We meet three situations in tftp enable config.

1. do nothing(don't mkdir and copy files).
2. prepare dir and copy files to tftproot path.
3. do step 2 and more with start service, enable service

boolean cannot handle theses.

In this fix 'enable' only decide whether we start service.

### DIFF
--- a/internal/pkg/configure/tftp.go
+++ b/internal/pkg/configure/tftp.go
@@ -35,6 +35,11 @@ func TFTP() error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mslacken:
We should also consider to add the tftp functionality to warewulfd. The tftp functionality in golang is quite mature as its also maintained by google.

PR #450, opened 2022-06-10
Fix wwct node set profile for single profile

### LINKED ISSUE #449: `wwctl node set $NODE --profile test` does not work
In 4.2.0 onwards profiles can't be set via the command
```
wwctl node set $NODE --profile test
```
As workaround (under the assumption that $NODE has set the fault profile) following works:
```
wwctl node set $NODE --addprofile test
wwctl node set $NODE --addprofile default
```

mslacken:
Fix #449


gmkurtzer (approved):
Good catch!

PR #455, opened 2022-06-20
Add suse builds mock build

mslacken:


elguero (approved):
Changes look good to me.

PR #458, opened 2022-06-22
add type to network configs

### LINKED ISSUE #457: support for wwinit icfcg.ww for Infiniband network-scripts
In the latest version 4.3.0-1.git_235c23c the setup of infiniband network with wwinit is limited.
The provided ifcfg.ww file did not contain a "TYPE" value for the network script, so rock-8 assumes a default "Type=Ethernet" .
However TYPE=Ethernet will not work for Infiniband. Entries in the corresponding node.conf will therefore not taken into account. 
You can of course modify the ifcfg.ww file under wwinit, but it seems a better way to do this is to include it in the standard distribution. 
Accoring to the source code netdev.Type is available <br>
Example: Add to ifcfg.ww  
{{ if $netdev.Type -}}
TYPE={{ $netdev.Type }}
{{ end -}}
<br>
The nodes.conf under nodes:<nodename>:network-devices can then look like:
      ib0:
        onboot: "true"
        device: ib0
        ipaddr: 192.168.1.9
        type: InfiniBand

Again, this is already possible, but the standard would help people get an HPC cluster up and running faster.

mslacken:
The type of a network was not evaluated in the network config templates.

Close #457

### DIFF
--- a/overlays/wwinit/etc/sysconfig/network-scripts/ifcfg.ww
+++ b/overlays/wwinit/etc/sysconfig/network-scripts/ifcfg.ww
@@ -20,6 +20,9 @@ GATEWAY={{ $netdev.Gateway }}
 {{ if $netdev.Hwaddr -}}
 HWADDR={{ $netdev.Hwaddr }}
 {{ end -}}
+{{ if $netdev.Type -}}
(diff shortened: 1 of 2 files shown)


mslacken:

PR #476, opened 2022-06-27
`container list` enhancement with no containers

### LINKED ISSUE #473: wwctl container show should not show output for containers that do not exist
`wwctl container show` produces predicted output for containers that do not exist; but I recently tried to use `wwctl container show` to determine whether a container is already defined. This didn't work, as the command doesn't differentiate between containers that exist and containers that do not exist.

I suggest that `wwcctl container show` should at least return nonzero for a container that does not exist, even if it continues to produce the same output on stdout; but I think it probably shouldn't produce any stdout either, in stead simply printing an error on stderr.

```
[root@janderson-admin1 ~]# wwctl container list
[root@janderson-admin1 ~]# wwctl container show narcontainer
/var/lib/warewulf/chroots/narcontainer/rootfs
[root@janderson-admin1 ~]# echo $?
0
```

mslacken:
close #473

### DIFF
--- a/internal/app/wwctl/container/list/main.go
+++ b/internal/app/wwctl/container/list/main.go
@@ -13,6 +13,10 @@ import (
 func CobraRunE(cmd *cobra.Command, args []string) error {
 
 	sources, err := container.ListSources()
+	if len(sources) == 0 {
+		wwlog.Info("no containers found\n")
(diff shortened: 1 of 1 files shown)


mslacken:
Does not work on new API

PR #490, opened 2022-07-07
Hide assetkey and ipmi password in list summary

kcdodd:
Consideration of not printing the few pieces of security relation information, like the bios assetkey and IPMI password.

### DIFF
--- a/internal/app/wwctl/node/list/main.go
+++ b/internal/app/wwctl/node/list/main.go
@@ -41,13 +41,14 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "Ipxe", ni.Ipxe.Source, ni.Ipxe.Print)
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "Init", ni.Init.Source, ni.Init.Print)
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "Root", ni.Root.Source, ni.Root.Print)
-			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "AssetKey", ni.AssetKey.Source, ni.AssetKey.Print)
+			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "AssetKey", ni.AssetKey.Source, "****")
 
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "IpmiIpaddr", ni.IpmiIpaddr.Source, ni.IpmiIpaddr.Print)
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "IpmiNetmask", ni.IpmiNetmask.Source, ni.IpmiNetmask.Print)
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "IpmiPort", ni.IpmiPort.Source, ni.IpmiPort.Print)
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "IpmiGateway", ni.IpmiGateway.Source, ni.IpmiGateway.Print)
 			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "IpmiUserName", ni.IpmiUserName.Source, ni.IpmiUserName.Print)
+			fmt.Printf("%-20s %-18s %-12s %s\n", nodeName, "IpmiPassword", ni.IpmiPassword.Source, "****")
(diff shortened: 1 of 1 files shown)


mslacken:
With the reflections this patch won't apply any more. But this is an important issue and we should apply a similar approach for the ipmipassword. With the already merged grpc API we should make sure that the password do not leave warewulfd without a good reason.
So a viable solution would be to add a bool to the Entry struct, if it's desired to print or not.

griznog:
I could see wanting to be able to list those for a number of reasons, could the `list` command get a flag `--show-secrets` and fields with secrets in them be tagged by the bool as secret rather than print or don't print?

This touches on a larger issue, is it worth adding a formatting structure to `list`, something like Slurm for instance where you can use "%.24a" to show whatever field a represents with 24 spaces in the output? I'm not convinced it's needed as it's fairly trivial to just read `nodes.conf` to get values, but having a dynamic way to control the `list` formatting would make it easier to write useful local scripts that parse those outputs. Also such an option could just leave secrets out of the defaults and allow someone to add them to the output format if they wanted to include them, sidestepping the secrets issue altogether.

griznog:
This also begs the question, how secret is secret? `nodes.conf` on my master is 0644, which I assume is by default as I don't recall ever explicitly changing any permissions there. If these are truly secret and given the simple/default cluster setup uses the master as a login node, we should probably also better protect `/etc/warewulf` from prying eyes.

kcdodd:
The default warewulf installation does not provide much security, since the nodes.conf is readable, and the wwctl command is not really set up with security, or cleanly separating admin from user-like operations. My use-case was wanting to allow users to view the status of the nodes, but still would like to not have the passwords or asset key available, since those would theoretically allow the nodes to be compromised remotely. Blanket removing the read permissions on nodes.conf is not an option because the wwctl command will not work at all in that case for non-root users. My solution was to create a 'warewulf' user, and change the owner and group of wwctl and all config files to 'warewulf', ensuring that the 'group' permission was readable by the 'warewulf' group, but noone else. Then, set the 'SGID' bit on wwctl. This way when wwctl is run by any user it has the permissions of the 'warewulf' group, but not as the owner, allowing it to read nodes.conf'. But since the group does not have write permissions, users still cannot change any of the configuration, and since users themselves do not belong to the 'warewulf' group, they cannot directly read the files without using 'wwctl'. But of course, if the wwctl command can print the password and asset key anyway, this doesn't provide any security.

griznog:
@kcdodd That's a good approach if you trust all the users logging into the master, but if the master is also the cluster login node that is likely not the case. I think we need a way to define users, operators and admins, with `/etc/warewulf.conf` being the logical place and then set a level on each command/option for which privilege level can run it and the same simple approach on the data itself. That may be a lot simpler to do now given the reflection foo @mslacken has done. 

I can see how this would be done on the master itself just using logged in user details, secondary groups, etc... What's not clear to me is how that would work remotely via the API, even if it's called on loopback. There needs to be an authentication layer there. 

Also seems simple enough to make things explicitly defined in the data structures restricted with an ACL, but how to do that for user defined tags which should be secret/restricted? That's where it'd be nice to have those nested. In fact, thinking about it it might be nice for all values to be a dict, with the dict containing a `value` and an `acl`. And of course, once everything is a dict then all sorts of possibilities open up.

kcdodd:
Yes, for us it is a small internal cluster, and the login node is the same as the master. And if something bad happens it would be nice for other users to diagnose the problem without admin being on call. As for "trust", I figured the the "my solution" had at least as much security as the default installation, since the only thing it should be doing is to prevent directly reading the contents of "nodes.conf", which was already possible.

> I can see how this would be done on the master itself just using logged in user details, secondary groups, etc... What's not clear to me is how that would work remotely via the API, even if it's called on loopback.

If I were to recommend an approach, it would be to have say a wwctl (or similar) command that can locally generate an API token for a user, that could then be used to access the API over https. Basically, if a user can ssh to somewhere that warewulf trusts, then they can generate a token to access it from somewhere else too.

mslacken:
With the new API we have some basic access control anyway, although its just boolean in the sense that you either have a key the API key or not.
For a more sophisticated control I would control I could think of following two approaches:
1. Add a security level field to every layer of the configuration, so the NodeInfo would have, the Ipmi another and so on. The advantage is that with this approach we would still have a single configuration, but it would be just be very simple rules there, if we do not want utterly complicated code for checking this stuff.
2. Add a new configuration file called e,g, 'nodes-securitiy.conf'  which has exact the same yaml structure like 'nodes.conf`, but not store the values of the entries there but their security context. Every entry would then have an additional security flag, which could be read out and set. This would enable a real granular control of what someone can do or not. Going down that road (multiple configuration files) would us also enable to have something like 'nodes.hist` where its stored when which entry was set by whom.

mslacken:
Can't merge any more due to conflicts. Also with #540 it would again possible to see the password.

PR #495, opened 2022-07-18
Use reflection for node and profile manipulation

mslacken:
As there was a lot of double code in the node/porfile adding and setting this PR uses reflections to create the commandline options. Subsequently this also simplifies the code for setting and adding nodes/profiles.
As imporvments all the avaiable properites can now be set on node/profuile creation.
As donwside some options have changed, but behave in the same way .


mslacken:
So reflections should now ready to be used. Did up to now just testing on the basic functions, so no merge yet.
I will do a full test, what means rolling out a small cluster on vms, and after that it should be ready to merge.

mslacken:
At this point more tests are needed. The main functions do now the following:
* set all the values of a NodeInfo (datastore) from a NodeConf (yaml) via Set or SetAlt (transform NodeInfo to NodeConf)
* set all the values of a NodeConf(yaml) from a NodeInfo (datastore) via GetReal/Get (transform Nodeconf to NodeInfo)
* create all the command line flags from a NodeConf

The constructor does now following thing, read the yaml file into a NodeConf and then transform this to a NodeInfo. If anything has changed, the NodeInfo is transformed to a NodeConf and stored to disk. A similar approach is done for the overlays, where the NodeInfo also goes to a NodeConf (via Get and not GetReal) so that the values are available in the templates.
Adding a new node.Entry for Nodes/Kernel/Ipmi/Netdev will need no changes in the transform functions. But deeper changes like adding a map[string]FancyNewStructLikeDevice will need changes in the transform functions.

Issue #497, opened 2022-07-21
Size of container images for Legacy BIOS machines is very limited


IngoMeyer441:
First of all: Thanks for this great project! We still have some Legacy BIOS machines which stopped to work with Warewulf after the upgrade to version 4.3.0. After reading the source code changes, I could find out that PR #235 broke the boot process for us. AFAIK, the problem is caused by Legacy PXE boot ROMS in 32 bit mode, which are limited in RAM by the 3 GB barrier. If PXE operates in BIOS mode it can only use less than 3 GB of RAM for storing and extracting data, so running `imgextract` in iPXE hits the RAM limitations very quickly. Thus, I suggest using the old `initrd` boot method of Warewulf version <= 4.2.0 for Legacy BIOS machines. In this mode, the iPXE script also downloads compressed images, but hands over the extraction step to the Linux kernel.

PR #498, opened 2022-07-21

IngoMeyer441:
GitHub PR #235 adds an advanced boot mechanism for UEFI machines which
causes old Legacy BIOS nodes to hit the 3 GB RAM barrier in 32 Bit iPXE
very quickly. Therefore, detect if an EFI or Legacy BIOS is loaded and
fall back to the old boot path in the second case.

This PR fixes GitHub issue #497.


mslacken:
Great catch, thanks for your contribution. I think we should include this also for the stable release.

@griznog Can you please have a look a this. I think you had more issues with that.

griznog:
This looks like the best way to handle this to me, definitely belongs in the next stable.

griznog:
Also, thanks @IngoMeyer441 for catching this and submitting the PR.

PR #499, opened 2022-07-21

mslacken:
- Fix the boot process for Legacy BIOS machines


mslacken:
wrong branch

PR #500, opened 2022-07-21

mslacken:
GitHub PR #235 adds an advanced boot mechanism for UEFI machines which
causes old Legacy BIOS nodes to hit the 3 GB RAM barrier in 32 Bit iPXE
very quickly. Therefore, detect if an EFI or Legacy BIOS is loaded and
fall back to the old boot path in the second case.

This commit fixes GitHub issue #497.


mslacken:
@griznog @IngoMeyer441 Thanks again, "backported" this to main branch

Issue #510, opened 2022-08-12
docs: manpages for warewulf.conf and nodes.conf


anderbubble:
We should have manpages for warewulf.conf and nodes.conf that detail all the available settings and their acceptable values.

anderbubble:
I'm happy to work on this; can it be assigned to me?

mslacken:
Thanks for looking into this.

anderbubble:
@mslacken It looks to me like warewulf.conf:dhcp['template'] is no longer used now that DHCP configuration has been moved to the host overlay. Can you confirm?

anderbubble:
warewulf.conf:tftp['tftproot'] also appears to be unused: rather, it appears to be compiled-in, defined in Makefile and warewulf.spec. Can you confirm? Should this be filed as a bug, or removed from the documentation?

mslacken:
> @mslacken It looks to me like warewulf.conf:dhcp['template'] is no longer used now that DHCP configuration has been moved to the host overlay. Can you confirm?
Right. In the 4.3.0 branch the warewulf.conf:dhcp['template'] variable is without effect. 
The question we have right now is, how to handle the host overlays in the future. Due to now solved limits in the templating mechanism, the actual host overlays are just a 1:1 copy from the overlay directory to the host, as it was not possible to address single files for overlay rendering. With that changes we could now go back and have and store this templates for the host under `/etc/` and store distro/site specific changes in `warewulf.conf`. Now this site/distro specifc paths are handled with different rpms.
So whats your take on this?

anderbubble:
I'd probably replace the `warewulf.conf:warewulf:host overlay` with a parameter that takes the name of one or more host overlays to apply, and then use that to specify one or more service-, distro-, and site-specific host configuration overlays. That way host overlays behave even more like compute node overlays.

anderbubble:
Fixed by #511.

PR #511, opened 2022-08-24
new man pages for warewulf.conf and nodes.conf

anderbubble:
This is a draft / work-in-progress PR for new `warewulf.conf` and `nodes.conf` man pages. Future commits may be reorganized / squashed as work is done.

Addresses #510


mslacken:
Great rephrasing of "here is the gun, there is your foot" for nodes.conf

PR #513, opened 2022-08-25
Capture wwapi binaries in rpm spec

anderbubble:
Naive implementation; but without some declaration rpmbuild fails.


mslacken:
thx

PR #514, opened 2022-08-25
Trim whitespace from IPMI output

anderbubble:
Historically, warewulf IPMI commands output extraneous newlines, producing output like this:

```
$ sudo wwctl power status c[1-4].dell-ciq.lan
c2.dell-ciq.lan: Chassis Power is on

c4.dell-ciq.lan: Chassis Power is on

c3.dell-ciq.lan: Chassis Power is on

c1.dell-ciq.lan: Chassis Power is on

```

This change trims extraneous whitespace, leading to output like this:

```
$ sudo wwctl power status c[1-4].dell-ciq.lan
c2.dell-ciq.lan: Chassis Power is on
c4.dell-ciq.lan: Chassis Power is on
c3.dell-ciq.lan: Chassis Power is on
c1.dell-ciq.lan: Chassis Power is on
```

This makes the output more like the output from `wwctl ssh` and other similar tools, improves readability, and better supports use with `dshbak`.

### DIFF
--- a/internal/pkg/power/ipmitool.go
+++ b/internal/pkg/power/ipmitool.go
@@ -3,6 +3,7 @@ package power
 import (
 	"os"
 	"os/exec"
+	"strings"
 )
 
 type IPMIResult struct {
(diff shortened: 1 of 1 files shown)


mslacken:
Good idea. Thanks.

PR #524, opened 2022-09-02
Fix a typo in symlink wwbackup implementation

anderbubble:


mslacken:
I definitely need a spell code correction. :-)

PR #530, opened 2022-09-12
Fix handling of uid and gid arguments during chown

### LINKED ISSUE #486: bug: wwctl overlay chown can't set gid
`wwctl overlay chown` documents an ability to set gid, but it doesn't appear to work.

```
# wwctl overlay create nothing
# wwctl overlay edit nothing nothing
# wwctl overlay ls -l nothing
PERM MODE    UID GID   SYSTEM-OVERLAY     FILE PATH
-rw-r--r--     0 0     nothing            /nothing

# id nobody
uid=65534(nobody) gid=65534(nobody) groups=65534(nobody)
# wwctl overlay chown nothing nothing 65534 65534
# wwctl overlay ls -l nothing
PERM MODE    UID GID   SYSTEM-OVERLAY     FILE PATH
-rw-r--r-- 65534 0     nothing            /nothing
```

anderbubble:
It appears that the intended behavior of overlay chown was changed in 00d8d42e9, but the argument handling was not updated to reflect the removal of overlayKind. This led to a crash when only the uid was specified, and a processing of the gid as the uid otherwise.

Fixes #486 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,8 +38,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to be changed accordingly
 - host overlays can globaly disbaled, but are enabled per default
 - `wwctl overlay build -H` will only build the overlays which are assigned to the nodes
-### Fixes
+### Fixed
 - GID is no longer changed to `0` when unspecified during chown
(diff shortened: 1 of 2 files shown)


MatthewHink (approved):
LGTM

PR #531, opened 2022-09-12
Don't change unspecified gid to 0 during chown

anderbubble:
Specifying 0 for gid as a default during chown causes the gid to be changed to 0 when left unspecified. Changing the default value to -1 causes the gid to be left unmodified when unspecified, which is a less surprising behavior.


MatthewHink:
LGTM

MatthewHink (approved):

mslacken:
Great catch, thanks

PR #533, opened 2022-09-13
Fix of container build include/exclude handling

### LINKED ISSUE #532: bug: /etc/warewulf/excludes is not processed
`/etc/warewulf/excludes` is not processed as expected. This is initially a result of the `ignore` variable being masked by redeclaration at https://github.com/hpcng/warewulf/blob/development/internal/pkg/container/build.go#L34; but fixing this reveals other errors in `FindFilterFiles`. https://github.com/hpcng/warewulf/blob/development/internal/pkg/util/util.go#L248

anderbubble:
This PR fixes the handling of the `/etc/warewulf/excludes` file read from the container definition, and further excludes files that reside in an excluded directory. (This functionality is otherwise missing due to deficiencies in `filepath.Match`. See https://github.com/golang/go/issues/11862)

This PR also largely rewrites `FindFilterFiles` in an attempt to prevent such errors in the future.

This is a rather significant rewrite of an integral function, and likely merits close review by senior developers / maintainers. However, it also includes a fix for a significant functional regression, so I suggest that it also merits priority.

Fixes #532

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - GID is no longer changed to `0` when unspecified during chown
 - Proper handling of uid/gid arguments during `wwctl overlay chown`
+- /etc/warewulf/excludes stored in the container definition is now
+  processed correctly, also excluding the contents of an excluded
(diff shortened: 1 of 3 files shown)


anderbubble:
Rebased on development and fixed linter.

PR #534, opened 2022-09-13
Replace all instances of wwlog.Printf

anderbubble:
wwlog provides named loggers for each level, which requires less code and is clearer than wwlog.Printf. The code has included a mix of both, but this commit consolidates existing code on the per-level functions.


elguero (commented):
Changes look good.

PR #539, opened 2022-09-15
Read the defaults for a node from defaults.conf

mslacken:


gmkurtzer:
What are the pros/cons between defaults.conf and the `default` profile that we already support?

mslacken:
> What are the pros/cons between defaults.conf and the `default` profile that we already support?

These are not the defaults from the profile, but basic things like kernel command line and the place where to find init ...

I will have to rewrite this a bit in order for NetworkManager overlay config as here e.g. every network needs to be ethernet if not defined otherwise

mslacken:
@elguero Can you please check if the NetworkManager stuff works on rocky-9?
Would especially interesting if the infiniband config works.

PR #540, opened 2022-09-16
add edit command for direct edit of node entries

mslacken:
Adds the command `wwctl node edit` which allows the direct edition for the
nodes of nodes.conf in yaml.
Internally it will detect if there was a modification and if yes, delete all
the requested nodes and adds the modified yaml back.


mslacken:
@elguero Can you please check if the NetworkManager stuff works on rocky-9?
Would especially interesting if the infinband config works.

mslacken:
@gmkurtzer @jcsiadal  Added now a simple import of nodes via a csv file. As I do not really know warewulf3 can you please check if that is enough?

PR #542, opened 2022-09-16
Fix output of wwctl container show

anderbubble:
Commit d8cd6049 introduced erroneous output coming from an inner function of container show, and (seemingly) erroneously replaces valid kernel versions with "not found". This commit fixes both of those issues, and moves the output to the outer cli function.

### DIFF
--- a/internal/app/wwctl/container/show/main.go
+++ b/internal/app/wwctl/container/show/main.go
@@ -24,8 +24,12 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	if !ShowAll {
 		fmt.Printf("%s\n", r.Rootfs)
 	} else {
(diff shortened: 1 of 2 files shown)


anderbubble:
@MatthewHink please let me know if I have misunderstood and broken something. 😅

anderbubble:
Actually, I'm thinking maybe I should change this to move the "not found" string to the outer cli function as well. That seems like something that should be a view concern rather than something that should be output by the API.

What do you think?

anderbubble:
I went ahead and did that, since it's what I think should happen; but if there's disagreement, it'd be easy to change back.

MatthewHink (approved):
+1 Thanks @anderbubble

MatthewHink (approved):

PR #543, opened 2022-09-16
Fix argument handling for overlay show rendering

anderbubble:
Error logging for overlay show rendering incorrectly checked the overlay name rather than the path to the file. It also used wwlog (which goes to stderr) to output the render. This commit replaces explicit args[] references with their named equivalents from earlier in the function, correcting those references along the way, and outputs the final render with fmt.Print.

This commit also causes the ERROR condition to Exit(), similarly to other ERROR conditions in this function.

### DIFF
--- a/internal/app/wwctl/overlay/show/main.go
+++ b/internal/app/wwctl/overlay/show/main.go
@@ -46,7 +46,15 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 
 		fmt.Print(string(f))
 	} else {
-		var host node.NodeInfo
+		if !util.IsFile(overlayFile) {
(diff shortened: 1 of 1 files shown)


anderbubble:
@mslacken let me know what you think.

anderbubble:
An example of the output I was seeing before:

```
[root@janderson-wwctl1 overlays]# wwctl overlay show -r compute1 rocky-8-split /etc/fstab.ww
ERROR  : rocky-8-split is not a file
WARN   : rocky-8-split has not the '.ww' so wont be rendered if in overlay
```

(The template was then rendered anyway, because the ERROR didn't cause an exit.)

mslacken:
Also noticed this strange behavior, but as you wrote I just ignored the warning.
Thanks,

PR #544, opened 2022-09-20
Use primary netdev for ssh

mslacken:
`wwctl ssh` has the "default" network interface hardcoded. Use the interface with the primary field instead.

@anderbubble can you please review

### DIFF
--- a/internal/app/wwctl/ssh/main.go
+++ b/internal/app/wwctl/ssh/main.go
@@ -40,21 +40,25 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 	}
 
 	for _, node := range nodes {
-
-		if _, ok := node.NetDevs["default"]; !ok {
-			fmt.Fprintf(os.Stderr, "%s: Default network device doesn't exist\n", node.Id.Get())
+		var primaryNet string
+		for netName := range node.NetDevs {
+			if node.NetDevs[netName].Primary.GetB() {
+				primaryNet = netName
+			}
+		}
+		if primaryNet == "" {
+			fmt.Fprintf(os.Stderr, "%s: Primary network device doesn't exist\n", node.Id.Get())
 			continue
 		}
-
-		if node.NetDevs["default"].Ipaddr.Get() == "" {
-			fmt.Fprintf(os.Stderr, "%s: Default network IP address not configured\n", node.Id.Get())
+		if node.NetDevs[primaryNet].Ipaddr.Get() == "" {
+			fmt.Fprintf(os.Stderr, "%s: Primary  network IP address not configured\n", node.Id.Get())
 			continue
 		}
 
 		nodename := node.Id.Print()
 		var command []string
 
-		command = append(command, node.NetDevs["default"].Ipaddr.Get())
+		command = append(command, node.NetDevs[primaryNet].Ipaddr.Get())
 		command = append(command, args[1:]...)
 
 		batchpool.Submit(func() {
--- a/internal/app/wwctl/ssh/main.go
+++ b/internal/app/wwctl/ssh/main.go
@@ -38,20 +38,11 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 		cmd.Usage()
 		os.Exit(1)
 	}
-	globalPrimary := ""
 	for _, node := range nodes {
 		var primaryNet string
-		if _, ok := node.NetDevs[globalPrimary]; ok {
-			if node.NetDevs[globalPrimary].Primary.GetB() {
-				primaryNet = globalPrimary
-			}
-		}
 		for netName := range node.NetDevs {
 			if node.NetDevs[netName].Primary.GetB() {
 				primaryNet = netName
-				if globalPrimary == "" {
-					globalPrimary = netName
-				}
 				break
 			}
 		}
--- a/internal/app/wwctl/ssh/main.go
+++ b/internal/app/wwctl/ssh/main.go
@@ -51,7 +51,7 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 			continue
 		}
 		if node.NetDevs[primaryNet].Ipaddr.Get() == "" {
-			wwlog.Error("%s: Primary  network IP address not configured\n", node.Id.Get())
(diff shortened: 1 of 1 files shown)


internal/app/wwctl/ssh/main.go
```diff
@@ -40,21 +40,25 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 	}
 
 	for _, node := range nodes {
-
-		if _, ok := node.NetDevs["default"]; !ok {
-			fmt.Fprintf(os.Stderr, "%s: Default network device doesn't exist\n", node.Id.Get())
+		var primaryNet string
+		for netName := range node.NetDevs {
+			if node.NetDevs[netName].Primary.GetB() {
+				primaryNet = netName
```
anderbubble, line 46: I suggest that there should be a `break` statement after a `Primary` interface is found. A node can have multiple interfaces marked "primary", and the current behavior uses the _last_ one; I think it should use the first one.

```bash
$ sudo cat /etc/warewulf/nodes.conf
WW_INTERNAL: 43
nodeprofiles:
  default:
nodes:
  c1:
    profiles:
    - default
    network devices:
      default:
        ipaddr: 10.0.0.100
        primary: "true"
      second:
        ipaddr: 10.0.0.101
        primary: "true"

$ sudo wwctl ssh c1 hostname # times out because this node doesn't actually exist; but shows what IP it's trying to use
c1: ssh: connect to host 10.0.0.101 port 22: Connection timed out
```

internal/app/wwctl/ssh/main.go
```diff
@@ -40,21 +40,25 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
...
+		if node.NetDevs[primaryNet].Ipaddr.Get() == "" {
+			fmt.Fprintf(os.Stderr, "%s: Primary  network IP address not configured\n", node.Id.Get())
 			continue
 		}
 
 		nodename := node.Id.Print()
 		var command []string
 
-		command = append(command, node.NetDevs["default"].Ipaddr.Get())
+		command = append(command, node.NetDevs[primaryNet].Ipaddr.Get())
```
anderbubble, line 61: I suggest the addition of a `wwlog.Debug` for the IP address that will be used.

anderbubble (changes requested):
A few suggestions, some of which may be varying degrees of a matter of taste.

internal/app/wwctl/ssh/main.go
mslacken: A break would be faster, but maps are not ordered in golang, so it does not matter if I use a break or continue.
A general plausibilty check of `nodes.conf` would be needed, before saving it.

internal/app/wwctl/ssh/main.go
```diff
@@ -40,21 +40,25 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
...
-		if _, ok := node.NetDevs["default"]; !ok {
-			fmt.Fprintf(os.Stderr, "%s: Default network device doesn't exist\n", node.Id.Get())
+		var primaryNet string
+		for netName := range node.NetDevs {
+			if node.NetDevs[netName].Primary.GetB() {
+				primaryNet = netName
+			}
+		}
+		if primaryNet == "" {
+			fmt.Fprintf(os.Stderr, "%s: Primary network device doesn't exist\n", node.Id.Get())
```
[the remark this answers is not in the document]
mslacken: Yes, just copied the old prinf.

mslacken:
Using now wwlog and trying to use the same netdev for all ssh communication.

internal/app/wwctl/ssh/main.go
```diff
@@ -38,23 +38,36 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
...
-
-		if _, ok := node.NetDevs["default"]; !ok {
-			fmt.Fprintf(os.Stderr, "%s: Default network device doesn't exist\n", node.Id.Get())
+		var primaryNet string
+		if _, ok := node.NetDevs[globalPrimary]; ok {
+			if node.NetDevs[globalPrimary].Primary.GetB() {
+				primaryNet = globalPrimary
+			}
+		}
+		for netName := range node.NetDevs {
```
anderbubble, line 49: I think this `for` loop needs to be wrapped with an `if` that tests whether `primaryNet` is already set. Otherwise, the discovery from `globalPrimary` above is discarded regardless.

internal/app/wwctl/ssh/main.go
```diff
@@ -38,23 +38,36 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 		cmd.Usage()
 		os.Exit(1)
 	}
-
+	globalPrimary := ""
```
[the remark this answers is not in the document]
mslacken: The simple solution is to allow just one primary interface. 
There is anyway the desire to have a checking tool, which could then also check e.g.
* ip adresses
* that a container exists
* that a profile exists
* ....
So in this case, we can assume that there is just *one*  primary interface per node.

internal/app/wwctl/ssh/main.go
mslacken: Yes, so if as assumed further down, that there is always *one* primary device, I can discard the globalPrimary.

internal/app/wwctl/ssh/main.go
```diff
@@ -38,23 +38,27 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
...
+		}
+		if primaryNet == "" {
+			wwlog.Error("%s: Primary network device doesn't exist\n", node.Id.Get())
 			continue
 		}
-
-		if node.NetDevs["default"].Ipaddr.Get() == "" {
-			fmt.Fprintf(os.Stderr, "%s: Default network IP address not configured\n", node.Id.Get())
+		if node.NetDevs[primaryNet].Ipaddr.Get() == "" {
+			wwlog.Error("%s: Primary  network IP address not configured\n", node.Id.Get())
```
anderbubble, line 54: There's an extra space here.

anderbubble (approved):
I recommend a tiny change to remove extra whitespace in a log line; but looks good to me otherwise!

PR #545, opened 2022-09-21
Remove trailing newline from wwlog

anderbubble:
I noticed that some wwlog calls included a trailing newline, but others did not. I tested both in isolation and discovered that the behavior was consistent regardless of whether a trailing newline was included. I further confirmed in code that wwlog appends a trailing newline automatically if it is not present; so a trailing newline is unnecessary in individual calls.

This commit removes trailing newlines from all calls to make them consistent. It also replaces two calls to wwlog.Printf. (see #534)


anderbubble:
My test script:

```golang
package main

import (
	"os"
	"github.com/hpcng/warewulf/pkg/hostlist"
)

func main() {
	for _, s := range hostlist.Expand(os.Args[1:]) {
		wwlog.Info("without explicit newline: %s", s)
		wwlog.Info("with explicit newline: %s\n", s)
	}
}
```

```console
[janderson@rocky warewulf]$ go run hello-world.go n[1-2]
without explicit newline: n1
with explicit newline: n1
without explicit newline: n2
with explicit newline: n2
```

PR #562, opened 2022-10-14
check if new template file was modified

### LINKED ISSUE #522: bug: wwctl overlay edit creates new overlay file even when editor is quit without writing
We recently encountered a situation where `wwctl overlay edit host etc/passwd.ww` was run by mistake. The resultant `vi` session was quit with `:q!` in an attempt to quit without writing a new file. A new file was written regardless, populated with the templating boilerplate due to the `.ww` suffix. This resulted in undesirable behavior when the host overlay was later accidentally deployed. (#523)

Note: this does _not_ occur for non-template overlay files.

If nothing is written out from the editor, no new overlay file should be written, similar to the end-result of a non-templated file. One potential implementation might be to write the boilerplate to a temporary file and inspect mtime and content? If neither mtime nor content has changed, don't write an overlay. (inspecting mtime is useful in case someone wants to just write out the boilerplate)

### LINKED ISSUE #523: wwctl overlay build applies host overlay to server
Running `wwctl overlay build` causes the host overlay to be deployed on the Warewulf server. I think this is confusing/surprising compared to how overlays are handled with compute nodes: that being, overlay "build" should be preparatory, and deployment of the host overlay should only occur via `wwctl configure`.

See also #522.

mslacken:
@anderbubble please check if this close #522


anderbubble:
Shouldn't it stop in this error condition? https://github.com/hpcng/warewulf/pull/562/files#diff-5fb3c9e719961a1b5b4867d06c9f825d2c44c8454b3e07b922163fda4cdd1db5R98-R100

By my reading, it could fail to update hasher, but then try to use it anyway at https://github.com/hpcng/warewulf/pull/562/files#diff-5fb3c9e719961a1b5b4867d06c9f825d2c44c8454b3e07b922163fda4cdd1db5R101.

PR #563, opened 2022-10-17
Enable VLAN_CMD in ipxe

anderbubble:
In order to support vlan tagging during the boot process, enable VLAN_CMD in ipxe.

Documented at https://ipxe.org/cmd/vcreate


mslacken:
@anderbubble Can you please update the binaries?

anderbubble:
@mslacken I've updated the x86_64 builds for ipxe, and I also added a new feature to `ipxe-update.sh` to allow an EMBED script to be supplied. (This is not used for the builds here, but is useful for when end-users need to build their own ipxe.)

PR #568, opened 2022-11-01
Fix use of uninitialised Tags map

Prepultrue:
NodeConf.Tags doesn't get initialized and panics when building an overlay that has a .ww template file for a node which has tags set.


mslacken:
Great catch, thanjs

PR #573, opened 2022-11-11
Correct host overlay processing during configure

### LINKED ISSUE #571: erroneous error: 'the overlay template '/etc/hosts.ww' does not exists in 'host' overlay
I just encountered this running `development` on an el9 system. I don't think that el9 is relevant here; but just in case.

```bash
[demo@demo2 warewulf]$ sudo wwctl configure hostfile
ERROR  : 'the overlay template '/etc/hosts.ww' does not exists in 'host' overlay

[demo@demo2 warewulf]$ sudo wwctl overlay list host -a
OVERLAY NAME                   FILES/DIRS  
host                           /etc/        
host                           /etc/dhcp/   
host                           /etc/dhcp/dhcpd.conf.ww
host                           /etc/dhcpd.conf
host                           /etc/exports.ww
host                           /etc/hosts.ww
host                           /etc/profile.d/
host                           /etc/profile.d/ssh_setup.csh
host                           /etc/profile.d/ssh_setup.sh
```

anderbubble:
wwctl configure hostfile was using the /host/ path twice, causing it to look in the wrong place for /etc/hosts.ww

Fixes #571 

### DIFF
--- a/internal/pkg/configure/hostfile.go
+++ b/internal/pkg/configure/hostfile.go
@@ -15,7 +15,8 @@ import (
 Creates '/etc/hosts' from the host template.
 */
 func Hostfile() error {
-	if !(util.IsFile(path.Join(overlay.OverlaySourceDir("host"), "/host/etc/hosts.ww"))) {
+	hostTemplate := path.Join(overlay.OverlaySourceDir("host"), "/etc/hosts.ww")
+	if !(util.IsFile(hostTemplate)) {
 		wwlog.Error("'the overlay template '/etc/hosts.ww' does not exists in 'host' overlay")
 		os.Exit(1)
 	}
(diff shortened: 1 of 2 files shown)


anderbubble:
@mslacken how's this? I'm particularly interested in your thoughts on my removal of / not explicitly declaring `var idEntry node.Entry`. This was previously done in `overlay`, but it appears to not be necessary, at least when using your new constructor. (Doesn't cause a problem during `wwctl configure hostfile` or `wwctl overlay build`.)

mslacken:
> @mslacken how's this? I'm particularly interested in your thoughts on my removal of / not explicitly declaring `var idEntry node.Entry`. This was previously done in `overlay`, but it appears to not be necessary, at least when using your new constructor. (Doesn't cause a problem during `wwctl configure hostfile` or `wwctl overlay build`.)
I declared the idEntry from the hostname only as this Entry is sometimes used as key in a map. So always setting it seemed to be kind of elegant and not really needed. So not setting it seems to be the better option.

PR #576, opened 2022-11-23
Configure interface type and name in ifcfg

### LINKED ISSUE #575: InfiniBand interfaces aren't configured as InfiniBand interfaces, fail to start
InfiniBand interfaces aren't configured as InfiniBand interfaces in ifcfg.ww, which causes them to fail to start.

As an aside, the name of the interface is also hard-coded to "default," ignoring the actual name of the interface.

anderbubble:
Closes #575


mslacken (approved):
Right, I did not update this file when working on the NetworkManager stuff

PR #578, opened 2022-11-29
Dockerfile for warewulf host container

mslacken:
- added source file for podman
- added more container scripts, smaller image
- cleanup in Dockerfile
- added README.md
- Using a builder container
- copy all the stuff to the application container
- added config-warewulf
- adding the ipxe stuff in the container


anderbubble:
@mslacken is this meant to be ready to merge? How would I test it to see that it's working as intended?

PR #579, opened 2022-11-29
Env var to identify when in container shell

### LINKED ISSUE #574: rfe: environment variable identifying `wwctl container shell`
Feature request: an environment variable defined in the `wwctl container shell` environment identifying that the current shell is part of a `wwctl container shell`. e.g., `WAREWULF_IN_CONTAINER_SHELL=true` (though I may personally prefer `WW_CONTAINER_SHELL=true`).

brianphan:
Set environment variable `WW_CONTAINER_SHELL` to the name of the container when a user shells into it via command `wwctl container shell ...`. 

Resolves #574

### DIFF
--- a/internal/app/wwctl/container/shell/main.go
+++ b/internal/app/wwctl/container/shell/main.go
@@ -40,7 +40,7 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 	c.Stdout = os.Stdout
 	c.Stderr = os.Stderr
 
-	os.Setenv("WW_CONTAINER_SHELL", "1")
+	os.Setenv("WW_CONTAINER_SHELL", containerName)
 
 	if err := c.Run(); err != nil {
 		fmt.Println(err)
--- a/internal/app/wwctl/container/shell/main.go
+++ b/internal/app/wwctl/container/shell/main.go
@@ -40,7 +40,7 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 	c.Stdout = os.Stdout
 	c.Stderr = os.Stderr
 
-	os.Setenv("WW_CONTAINER_SHELL", "true")
+	os.Setenv("WW_CONTAINER_SHELL", "1")
 
 	if err := c.Run(); err != nil {
 		fmt.Println(err)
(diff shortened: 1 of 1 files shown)


mslacken:
Thanks for your contribution.
Is there a reason to use "true" as value, as I would preferred "1" in the shell context for true?

brianphan:
I used true as that was what was suggested in the issue, however, I do not have a preference. I updated the environment variable to use value 1.

griznog:
In shell at least "0" is true and non-zero is false. Setting it to "1" to indicate we are in a container shell seems backwards. E.g.:

```
(base) griznog@lepomis:~$ true
(base) griznog@lepomis:~$ echo $?
0
(base) griznog@lepomis:~$ false
(base) griznog@lepomis:~$ echo $?
1
```
The only tattoo I've ever  seriously considered getting was `true == 0 && false != 0` as I have to do the above more often than I'd like to remind/convince myself how that works. 

Also a fan of using `true` and `false` because they are unambiguous when humans read them and the variable value can be executed as a test, e.g.

```
(base) griznog@lepomis:~$ WW_CONTAINER_SHELL=true
(base) griznog@lepomis:~$ ${WW_CONTAINER_SHELL} && echo "It was true"
It was true
(base) griznog@lepomis:~$ WW_CONTAINER_SHELL=false
(base) griznog@lepomis:~$ ${WW_CONTAINER_SHELL} && echo "It was true"
(base) griznog@lepomis:~$ 
```

Where `0` fails:

```
(base) griznog@lepomis:~$ WW_CONTAINER_SHELL=0
(base) griznog@lepomis:~$ ${WW_CONTAINER_SHELL} && echo "It was true"
-bash: 0: command not found
```

The unfortunate caveat to this is that `unset` variables evaluate to true, so you'd want to ensure the variable was set to false when not in a shell if using this approach and writing tests this way:

```
(base) griznog@lepomis:~$ unset WW_CONTAINER_SHELL
(base) griznog@lepomis:~$ ${WW_CONTAINER_SHELL} && echo "It was true"
It was true
```

anderbubble:
For this specific reason (1/0 being confusing in this context) I suggest "true"; I feel like I've seen that in similar circumstances.

Or, perhaps we could obviate this entire question by adjusting the requirements slightly: what if, in stead of setting `WW_CONTAINER_SHELL` to a boolean value, we just set it to the name of the container? That would get you the boolean state (the presence of the variable means we're in a container shell) and it also provides the name of the container it's running in.

brianphan:
I'm open to all the solutions mentioned above. Which option would be the best one to proceed with?

brianphan:
Setting the environment variable to the container name as this was the preferred method of the user who made this request.

mslacken:
thanks

PR #595, opened 2022-12-14
Add --- to the head of FallBackConf

anderbubble:
Why have I done this?

- Because defaults.conf had a starting newline, which bugs me.
- Because moving the first line of FallBackConf looks unlovely.
- Because yaml happens to have a --- document-start syntax.

I completely agree that this is (hopefully) a no-op change; but if anyone else thinks it makes things aesthetically more pleasing, and agrees that it doesn't break anything, go ahead and merge. ^_^

### DIFF
--- a/internal/pkg/node/constructors.go
+++ b/internal/pkg/node/constructors.go
@@ -17,7 +17,7 @@ var ConfigFile string
 var DefaultConfig string
 
 // used as fallback if DefaultConfig can't be read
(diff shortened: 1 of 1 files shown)


anderbubble:
Pinging @griznog in case it's the kind of thing he'd smile at.

mslacken (commented):
When I run `print_defaults` I still see the `----` so is this intended?

PR #596, opened 2022-12-14
Treat wwapi config files as built versions

anderbubble:
The wwapi*.conf files are build from .in variants, so they should not be tracked in the tree. Adding them to gitignore prevents them from being added back in, and adding them to Makefile:clean cleans them up.


anderbubble:
Slack discussion:

https://hpcng.slack.com/archives/C019X4EUFFW/p1670951093330089

PR #599, opened 2022-12-17
New debug overlay with template

anderbubble:
A new debug overlay template serves as an example of how to access all template variables which can also be rendered against any node to display all template variable values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - New man page for `defaults.conf` #593
+- New debug overlay template
 
 ## [4.4.0rc2] 2022-12-09
 
--- a/overlays/debug/warewulf/template-variables.md.ww
+++ b/overlays/debug/warewulf/template-variables.md.ww
@@ -1,0 +1,168 @@
+# Warewulf template variables
+
+This Warewulf template serves as a complete example of the variables
+available to Warewulf templates. It may also be rendered against a
+node to debug its apparent configuration.
+
+```sh
+wwctl overlay show --render $nodename debug /warewulf/template-variables.md.ww
+```
+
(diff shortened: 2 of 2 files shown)


anderbubble:
I'm marking this as a draft for now because I'd like to add additional documentation in the template; but I wanted to get it in front of @mslacken and @griznog for comment.

griznog:
This is way better than my local attempt which just covers the 5 or so things I care about Right Now (tm). Is it worth setting a long (very long) term goal to have this auto-generated from the data structures/YAML schema as potentially part of automated testing?

anderbubble:
I think I would only want to do it if we could auto-generate the template at build-time. If the template auto-generated at run-time it would negate its usefulness as documentation for how to access the template variables.

griznog:
It's definitely a build-time thing. 

I'm now wondering how hard it'd be to have this file produce something that can be `source`'d in BASH. I'm doing that now with a pitiful subset of values.

griznog:
Also is there any reason this couldn't just live in the `wwinit` overlay? It'd be handy if it was just always there for casual troubleshooting on running nodes without taking an extra step to build an overlay or reboot.

anderbubble:
You don't have to build it or reboot; just `wwctl overlay show --render`. I suppose it could live in wwinit; but that wasn't necessary for my purposes.

griznog:
I include `nodes.conf` on all my nodes, then have downstream local tools that use it for managing/configuring the node and services. Also my runtime overlays get pulled and updated on the live nodes so I can update configs without rebooting. Ultimately having `nodes.conf` and having it update when I pull new overlay versions solves this problem with a large hammer, but because I am still struggling to wrap my head around using `yq` to pull values from YAML into BASH, it's on my to-do list to produce a BASH friendly view of all node settings. 

As I'm thinking this through and typing, I'm now wondering if what I really want is an API call to return the current node config from `warewulfd` rather than sending around all these different overlay files.

PR #601, opened 2022-12-18
Correct confirmation output in profile set

### LINKED ISSUE #600: 4.4.0rc2: wwctl profile set --comment asks about nodes when none are set up
`wwctl profile set --comment "OpenHPC on Alpha" openhpc` returns `Are you sure you want to modify 1 node(s): [y/N]` when no nodes have even been set up.

anderbubble:
Closes #600


mslacken (approved):
there could be more occurrences like this in the code, thanks

PR #621, opened 2022-12-22
Refactor docker containers and add centos-7

anderbubble:
These docker containers populate docker.io/warewulf. A new centos-7 container serves as an example for others to build from when required. The new Makefile builds .tar images with Podman which can be imported into Warewulf directly without going through a registry.


mslacken (approved):
no functional code changes, thanks

PR #630, opened 2023-01-04
GetB must not ignore altvalue

### LINKED ISSUE #629: GetB method ignores profile (altvalue) if node (value) unset
https://github.com/hpcng/warewulf/blob/a22996245c56c244e7ccd09a1f4989a8eb8502c2/internal/pkg/node/methods.go#L200-L211

For network devices, I have the "default" network set as primary=true in the default profile, and a second "ib" network defined in an "infiniband" profile with primary=false. The node configs leave primary unset for all networks.

The /etc/hosts file should be generated with the short hostname appearing only for the default network, but as-is, it does not show on any host line.  [Work-around -->] Manually adding primary=true to the default network in the node config overrides as intended. 

For example, I added a comment in the template for each netdev:
     #{{$netdev.Primary.GetB}}, {{$netdev.Primary.Get}}, {{$netdev.Primary}}

results:
Node 1:
  default net: primary=true (node), true (profile)
     node1_ip node1-default node1-eth0 node1
     #true, true, {[true] [true] default []}
  infiniband: primary=unset (node), false (profile) 
    node1_ipoib node1-ib node1-ib0
     #false, false, {[] [false] infiniband []}

Node 2:
  default net: primary=unset(node), true (profile)
    node2_ip node2-default node2-eth0
     #false, true, {[] [true] default []}
  infiniband: primary=unset (node), false (profile)
    node2_ipoib node2-ib node2-ib0
     #false, false, {[] [false] infiniband []}

mslacken:
close #629

### DIFF
--- a/internal/pkg/node/methods.go
+++ b/internal/pkg/node/methods.go
@@ -209,6 +209,11 @@ func (ent *Entry) GetB() bool {
 		if ent.altvalue[0] == "false" || ent.altvalue[0] == "no" ||
 			ent.altvalue[0] == "No" || ent.altvalue[0] == "0" {
 			return false
+		} else if len(ent.altvalue) == 0 {
+			if ent.def[0] == "false" || ent.def[0] == "no" ||
+				ent.def[0] == "No" || ent.def[0] == "0" {
+				return false
+			}
 		}
 	}
 	return true
--- a/internal/pkg/node/methods.go
+++ b/internal/pkg/node/methods.go
@@ -201,9 +201,15 @@ func (ent *Entry) Get() string {
 Get the bool value of an entry.
 */
 func (ent *Entry) GetB() bool {
-	if len(ent.value) == 0 || ent.value[0] == "false" || ent.value[0] == "no" ||
-		len(ent.altvalue) == 0 || ent.altvalue[0] == "false" || ent.altvalue[0] == "no" {
+	if ent.value[0] == "false" || ent.value[0] == "no" ||
+		ent.value[0] == "No" || ent.value[0] == "0" {
 		return false
+	}
+	if len(ent.value) == 0 {
+		if ent.altvalue[0] == "false" || ent.altvalue[0] == "no" ||
+			ent.altvalue[0] == "No" || ent.altvalue[0] == "0" {
+			return false
+		}
 	}
 	return true
 }
--- a/internal/pkg/node/methods.go
+++ b/internal/pkg/node/methods.go
@@ -201,22 +201,22 @@ func (ent *Entry) Get() string {
 Get the bool value of an entry.
 */
 func (ent *Entry) GetB() bool {
-	if ent.value[0] == "false" || ent.value[0] == "no" ||
-		ent.value[0] == "No" || ent.value[0] == "0" {
-		return false
-	}
-	if len(ent.value) == 0 {
-		if ent.altvalue[0] == "false" || ent.altvalue[0] == "no" ||
-			ent.altvalue[0] == "No" || ent.altvalue[0] == "0" {
-			return false
-		} else if len(ent.altvalue) == 0 {
-			if ent.def[0] == "false" || ent.def[0] == "no" ||
-				ent.def[0] == "No" || ent.def[0] == "0" {
-				return false
-			}
+	if len(ent.value) > 0 {
+		return !(strings.ToLower(ent.value[0]) == "false" ||
+			strings.ToLower(ent.value[0]) == "no" ||
+			ent.value[0] == "0")
+	} else {
+		if len(ent.altvalue) > 0 {
+			return !(strings.ToLower(ent.altvalue[0]) == "false" ||
+				strings.ToLower(ent.altvalue[0]) == "no" ||
+				ent.altvalue[0] == "0")
+		} else {
+			return !(len(ent.def) == 0 ||
+				strings.ToLower(ent.def[0]) == "false" ||
+				strings.ToLower(ent.def[0]) == "no" ||
+				ent.def[0] == "0")
 		}
 	}
-	return true
 }
 
 /*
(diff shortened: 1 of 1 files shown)


NateCrawford:
I think this will still effectively ignore a true altvalue when value is unset. Something like this should work:
```
if ent.value[0] == "true" || ent.value[0] == "yes" {
   return true
}
if ent.value[0] == "false" || ent.value[0] == "no" {
   return false
}
if len(ent.value) == 0 {
   if ent.altvalue[0] == "true" || ent.altvalue[0] == "yes" {
      return true
   }
}

return false
```

  Should this method also be able to handle the default case (ent.def[0])? It's there for Get but not GetB.

mslacken:
I have updated this now, you were nearly right, but its better to check just for false, like it is with integers, where everything except 0 is true.

NateCrawford:
Is the default "ent.def[0]" something you need to catch, or are those guaranteed to never be boolean values?

NateCrawford:
In the special case of multiple network devices, if nothing is set at any level (value, altvalue, def), GetB defaulting to either true or false is equally bad, and testing for one and only one primary would be a good double-check.

Assuming it is not overridden, it seems natural to set the network labelled "default" to be primary by default ("default".Primary.def[0]=true), but I'm not sure of the best way to do that.

mslacken:
> In the special case of multiple network devices, if nothing is set at any level (value, altvalue, def), GetB defaulting to either true or false is equally bad, and testing for one and only one primary would be a good double-check.
> 

This special case is caught in the constructor, see
https://github.com/hpcng/warewulf/blob/b64718fd0e01faf1ff56c18a2e3189ecd27572ad/internal/pkg/node/constructors.go#L138

anderbubble (commented):
I don't understand the  logic here. This is what I expect it should be; but if not, can you help me understand what I'm missing?

```
func (ent *Entry) GetB() bool {
	if len(ent.value) > 0 {
		return !(strings.ToLower(ent.value[0]) == "false" ||
			strings.ToLower(ent.value[0]) == "no" ||
			ent.value[0] == "0")
	} else {
		return !(len(ent.altvalue) == 0 ||
			strings.ToLower(ent.altvalue[0]) == "false" ||
			strings.ToLower(ent.altvalue[0]) == "no" ||
			ent.altvalue[0] == "0")
	}
}
```

NateCrawford:
> > In the special case of multiple network devices, if nothing is set at any level (value, altvalue, def), GetB defaulting to either true or false is equally bad, and testing for one and only one primary would be a good double-check.
> 
> This special case is caught in the constructor, see
> 
> https://github.com/hpcng/warewulf/blob/b64718fd0e01faf1ff56c18a2e3189ecd27572ad/internal/pkg/node/constructors.go#L138

Ah, OK. I think this is why I was seeing "default:primary    SUPERSEDED   true" on the first test nodes I set up. This was confusing! Also, being in ALLCAPS made it look important/unusual. I wanted to minimize the per-node config, but this kept coming back.

mslacken:
@anderbubble logic is sometimes weired for me. 

> > > In the special case of multiple network devices, if nothing is set at any level (value, altvalue, def), GetB defaulting to either true or false is equally bad, and testing for one and only one primary would be a good double-check.
> > 
> > 
> > This special case is caught in the constructor, see
> > https://github.com/hpcng/warewulf/blob/b64718fd0e01faf1ff56c18a2e3189ecd27572ad/internal/pkg/node/constructors.go#L138
> 
> Ah, OK. I think this is why I was seeing "default:primary SUPERSEDED true" on the first test nodes I set up. This was confusing! Also, being in ALLCAPS made it look important/unusual. I wanted to minimize the per-node config, but this kept coming back.
Didn't consider that this is not consistent behavior, fixed this in #639

mslacken:
@anderbubble can you please recheck again?

anderbubble:
> @anderbubble can you please recheck again?

I think the nesting of the `ent.def` clause is unnecessary. I'd try to simplify it like this:

```golang
func (ent *Entry) GetB() bool {
	if len(ent.value) > 0 {
		return !(strings.ToLower(ent.value[0]) == "false" ||
			strings.ToLower(ent.value[0]) == "no" ||
			ent.value[0] == "0")
	} else if len(ent.altvalue) > 0 {
			return !(strings.ToLower(ent.altvalue[0]) == "false" ||
				strings.ToLower(ent.altvalue[0]) == "no" ||
				ent.altvalue[0] == "0")
	} else {
			return !(len(ent.def) == 0 ||
				strings.ToLower(ent.def[0]) == "false" ||
				strings.ToLower(ent.def[0]) == "no" ||
				ent.def[0] == "0")
	}
}
```

mslacken:
> > @anderbubble can you please recheck again?
> 
> I think the nesting of the `ent.def` clause is unnecessary. I'd try to simplify it like this:
> 
> ```go
> func (ent *Entry) GetB() bool {
> 	if len(ent.value) > 0 {
> 		return !(strings.ToLower(ent.value[0]) == "false" ||
> 			strings.ToLower(ent.value[0]) == "no" ||
> 			ent.value[0] == "0")
> 	} else if len(ent.altvalue) > 0 {
> 		return !(strings.ToLower(ent.altvalue[0]) == "false" ||
> 			strings.ToLower(ent.altvalue[0]) == "no" ||
> 			ent.altvalue[0] == "0")
> 	} else {
> 		return !(len(ent.def) == 0 ||
> 			strings.ToLower(ent.def[0]) == "false" ||
> 			strings.ToLower(ent.def[0]) == "no" ||
> 			ent.def[0] == "0")
> 	}
> }
> ```

Can you please submit this to the PR?

mslacken:
@anderbubble if it now looks good I will merge in tomorrow

anderbubble:
It looks right to me, but I haven't executed it.

PR #645, opened 2023-01-16
resolve symlink for kernel version

mslacken:
Resolves the kernel version, if e.g the symlink `boot` was found in the first place,

### DIFF
--- a/internal/pkg/container/kernel.go
+++ b/internal/pkg/container/kernel.go
@@ -1,9 +1,11 @@
 package container
 
 import (
+	"os"
 	"path"
 	"path/filepath"
 	"sort"
+	"strings"
 
 	"github.com/hpcng/warewulf/internal/pkg/util"
 	"github.com/hpcng/warewulf/internal/pkg/wwlog"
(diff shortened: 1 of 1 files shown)


anderbubble:
Oh, nice. I think I know what this is. I'll check it out.

anderbubble:
@mslacken thanks for doing this! I can see the output is now improved for a centos-7 image I had laying around.

Before it was

```
centos-7                  0      boot
```

Now it's producing

```
CONTAINER NAME   NODES  KERNEL VERSION   CREATION TIME        MODIFICATION TIME    SIZE    
centos-7         0      3.10.0-1160.81.1.el7.x86_64 22 Dec 22 14:48 CST  22 Dec 22 14:48 CST  1.7 GiB
```

I'll give a code review, too; but functionality looks great!

anderbubble (commented):
I think this code looks just fine; but if you're interested, you might consider this version, which is a little bit simpler, though perhaps slightly less obvious.

https://github.com/anderbubble/warewulf/commit/42f4f1db34cd18f718246c5692086606a73a11fe

PR #651, opened 2023-01-24
make ipxe binary source configurable

mslacken:
## Configurable ipxe binaries
The location and file names of the ipxe binaries are hardcoded. So this change makes them configurable via `warewulf.conf`.

### DIFF
--- a/overlays/host/etc/dhcp/dhcpd.conf.ww
+++ b/overlays/host/etc/dhcp/dhcpd.conf.ww
@@ -20,15 +20,11 @@ option architecture-type   code 93  = unsigned integer 16;
 if exists user-class and option user-class = "iPXE" {
     filename "http://{{$.Ipaddr}}:{{$.Warewulf.Port}}/ipxe/${mac:hexhyp}";
 } else {
-    if option architecture-type = 00:0B {
-        filename "/warewulf/arm64.efi";
-    } elsif option architecture-type = 00:09 {
-        filename "/warewulf/x86_64.efi";
-    } elsif option architecture-type = 00:07 {
-        filename "/warewulf/x86_64.efi";
-    } elsif option architecture-type = 00:00 {
-        filename "/warewulf/x86_64.kpxe";
-    }
+{{range $type,$name := $.Tftp.IpxeBinaries }}
+    if option architecture-type = {{ $type }} {
+        filename "/warewulf/{{ $name }}";
+    } 
+{{ end }}
 }
 
 {{if eq .Dhcp.Template "static" -}}
--- a/internal/pkg/buildconfig/defaults.go
+++ b/internal/pkg/buildconfig/defaults.go
@@ -27,7 +27,7 @@ func BINDIR() string {
 }
 
 func DATADIR() string {
-	wwlog.Debug("DATADIR = '%s'", bindir)
+	wwlog.Debug("DATADIR = '%s'", datadir)
 	return datadir
 }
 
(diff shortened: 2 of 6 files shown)


overlays/host/etc/dhcp/dhcpd.conf.ww
```diff
@@ -20,15 +20,11 @@ option architecture-type   code 93  = unsigned integer 16;
...
-    if option architecture-type = 00:0B {
-        filename "/warewulf/arm64.efi";
-    } elsif option architecture-type = 00:09 {
-        filename "/warewulf/x86_64.efi";
-    } elsif option architecture-type = 00:07 {
-        filename "/warewulf/x86_64.efi";
-    } elsif option architecture-type = 00:00 {
-        filename "/warewulf/x86_64.kpxe";
-    }
+{{range $type,$name := $.Tftp.IpxeBinaries }}
```
anderbubble, line 23: We also need to update `/etc/warewulf/examples/static-dhcpd.conf.ww`

anderbubble: Actually, it's looking like that example file should just be removed, as it looks like the static functionality was already merged into the overlay with a variable.

anderbubble: Worth noting that these were "elsif" before, and now it's just a list of ifs. Probably fine.

anderbubble (changes requested):
@mslacken can you update /etc/warewulf/warewulf.conf with examples of how to configure this? Even if it's commented-out. It isn't immediately obvious to me where in the structure these should go.

mslacken:
> Is there any way we could make it that the specified values are overlayed on top of the default values? So that specifying one doesn't remove the others?
This would be default behavior of a `yaml` unmarshall, but this would lead to an error if the default ipxe binaries are not present. The only way to fix this 'error' for the end user would to rename the own ipxe binaries to the name matching the default ones.

anderbubble:
>> Is there any way we could make it that the specified values are overlayed on top of the default values? So that specifying one doesn't remove the others?
> This would be default behavior of a `yaml` unmarshall, but this would lead to an error if the default ipxe binaries are not present. The only way to fix this 'error' for the end user would to rename the own ipxe binaries to the name matching the default ones.

Here's what I'm talking about:

By default, `dhcpd.conf` contains this:

```
if exists user-class and option user-class = "iPXE" {
    filename "http://192.168.200.1:9873/ipxe/${mac:hexhyp}";
} else {

    if option architecture-type = 00:00 {
        filename "/warewulf/x86_64.kpxe";
    } 

    if option architecture-type = 00:07 {
        filename "/warewulf/x86_64.efi";
    } 

    if option architecture-type = 00:09 {
        filename "/warewulf/x86_64.efi";
    } 

    if option architecture-type = 00:0B {
        filename "/warewulf/arm64.efi";
    } 

}
```

But if I configure `warewulf.conf` like this:

```
tftp:
  enabled: true
  systemd name: tftp
  ipxe:
    '00:00': custom.kpxe
```

Then dhcp replaces the entire block like this:

```
if exists user-class and option user-class = "iPXE" {
    filename "http://192.168.200.1:9873/ipxe/${mac:hexhyp}";
} else {

    if option architecture-type = 00:00 {
        filename "/warewulf/custom.kpxe";
    } 

}
```

I think it would be better if it only replaced the `00:00` value with `custom.kpxe`, as that's the only one that was specified, or if I could add additional architecture-types without having to also explicitly re-specify all the default values.

PR #654, opened 2023-01-25
New CONTRIBUTING file

anderbubble:
CONTRIBUTING file as mentioned in the [draft technical charter](https://docs.google.com/document/d/1SO6i_7of0BlSdajcFK7mvgMIVD9zmOP0n5KF4xEb2tk).

Open questions:

* Is the initial member list for the technical steering committee correct?
* Do we need to draft a code of conduct?
* Do we need signed-off and a DCO.md?
* We need a branching policy to reflect here
* Do the comments here about tests apply? Will they apply?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,104 @@
+# Contributing
+
+## Contributor's Agreement
+
+You are under no obligation whatsoever to provide any bug fixes,
+patches, or upgrades to the features, functionality or performance of
+the source code ("Enhancements") to anyone; however, if you choose to
+make your Enhancements available either publicly, or directly to the
+project, without imposing a separate written license agreement for
+such Enhancements, then you hereby grant the following license: a
+non-exclusive, royalty-free perpetual license to install, use, modify,
+prepare derivative works, incorporate into other computer software,
+distribute, and sublicense such enhancements or derivative works
+thereof, in binary and source code form.
+
+## Technical Charter
+
+- section 2(B) requires a list of Technical Steering Committee (TSC)
+  members here
+
+### Technical Steering Committee
+
+(in alphabetical sort order of surname)
+
+```text
+- Jonathon Anderson <jonathon.anderson@ciq.co>, <anderbubble@gmail.com>
+- Christian Goll <christian.goll@googlemail.com>
+- John Hanks <griznog@gmail.com>
+- Gregory Kurtzer <gmkurtzer@gmail.com>
+- Jeremy Siadal <jeremy.c.siadal@intel.com>
+```
+
+
+## Pull Requests (PRs)
+
+1. All PRs should summarize the purpose of the PR in the attached
+   GitHub conversation.
+2. Larger fixes or enhancemens should be discussed with the project
+   leader or developers, e.g., on Slack or over Email.
+3. Essential bug fix PRs should be sent to both development and
+   release branches.
+4. Small bug fix and feature enhancement PRs should be sent to
+   development only.
+5. Follow the existing code style precedent. For Go, you will mostly
+   conform to the style and form enforced by the "go fmt" and "golint"
+   tools for proper formatting.
+6. For any new functionality, please write appropriate go tests that
+   will run as part of the Continuous Integration (github workflow
+   actions) system.
+7. Make sure that the project's default copyright and header have been
+   included in any new source files.
+8. Make sure your code passes linting, by running `make list` before
+   submitting the PR. We use `golangci-lint` as our linter. You may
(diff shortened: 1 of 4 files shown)


anderbubble:
Added myself to the list as discussed in the community meeting.

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+all your interactions with the project members and users.  
+
+In addition, all commits must be "Signed-off" by using `git commit
+-s`,
+acknowledging that you agree to the [Developer Certificate of Origin](DCO.md).  
+-->
+
+## Pull Requests (PRs)
+
+1. Essential bug fix PRs should be sent to both main and release
```
elguero, line 53: Instead of saying essential, I would be specific.

"All bug fixes and regressions should be submitted against the development and main branch."
"All security fixes need to be submitted against the development branch, main branch and currently released tag"

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+   branches.
+2. Small bug fix and feature enhancement PRs should be sent to main
+   only.
+3. Follow the existing code style precedent. For Go, you will mostly
+   conform to the style and form enforced by the "go fmt" and "golint"
+   tools for proper formatting.
+4. For any new functionality, please write appropriate go tests that
```
elguero, line 60: I would add, before adding new functionality, check in with the community for feedback on your idea before embarking on a path that would not be accepted into the code base.  Something like that.

I do see it in the Getting Started section. It might be beneficial to repeat that idea here as well.  To be honest, I glossed over that point in Getting Started and jumped straight to here at first.

elguero (changes requested):
A few comments on the document.

Plus, I would recommend that we establish a process for code reviews and merging.

We should establish how many code reviewers we need to approve PR.  I would recommend two approvers.

A PR should not be merged for at least two days.  For instance, since contributors are in different time zones, it would be nice to give the community a chance to review the code and make comments before it is merged.

Who is allowed to merge PR?  The TSC?  Maybe just make sure that is clear somewhere.

We should explain what our branching/merging strategy is.

Just some initial thoughts while reviewing this.

anderbubble:
Thanks for the feedback, @elguero!

> I would recommend that we establish a process for code reviews and merging.

> We should establish how many code reviewers we need to approve PR. I would recommend two approvers.

> A PR should not be merged for at least two days.

> We should explain what our branching/merging strategy is.

So far, my goal has been to get a decision-making process in place, and then use that process to make such decisions. Anything that's in here already that goes beyond that is only there because it was in the template I was using, and I didn't see a need to remove stuff unless people wanted it removed.

I've taken these questions / suggestions and added them to a to-do list for the TSC, or whatever it ends up being.

> Who is allowed to merge PR? The TSC? Maybe just make sure that is clear somewhere.

My expectation is that this is just whomever is granted the ability to do that in GitHub, likely what the draft charter refers to as "committers." But this could / should be called out explicitly.

I'll respond to the others where they are.

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+In addition, all commits must be "Signed-off" by using `git commit
+-s`,
+acknowledging that you agree to the [Developer Certificate of Origin](DCO.md).  
+-->
+
+## Pull Requests (PRs)
+
+1. Essential bug fix PRs should be sent to both main and release
+   branches.
+2. Small bug fix and feature enhancement PRs should be sent to main
```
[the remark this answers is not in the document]
anderbubble: This is a mistake. I thought I already updated this, but I'll do it now. Thanks for the catch.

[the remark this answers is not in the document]
anderbubble: Looks like I just hadn't pushed it.

CONTRIBUTING.md
anderbubble: I agree with making it more specific; but I think more discussion needs to be had about the branching strategy in order for that specificity to be correct. I'm afraid that the way people are currently talking about development vs main will lead to painful merges or (more like what's happening now) a divergence that makes one of them not as well used or tested. But rather than try to figure that out (or make my arguments) at this stage, I figured the lesser specificity (using the language from Apptainer) would make it easier to get something down, and then we could discuss the specifics next.

CONTRIBUTING.md
anderbubble: I just merged the Getting Started section into the PRs section, so there wouldn't be two places to look. ^_^

anderbubble:
Since we're targeting a different set of bylaws, this is no longer relevant.

PR #672, opened 2023-02-01
use always @BINDIR@

### LINKED ISSUE #624: warewulfd.service file has inconsistent locations for wwctl
The ExecReload and ExecStop lines in include/systemd/warewulfd.service.in need to have @BINDIR@ in place of /usr/bin

mslacken:
close #624


anderbubble:
Confirmed. https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/

PR #682, opened 2023-02-10
set the primary network in NodeInfo not in NetDev

mslacken:
Prior the primary was a boolean attribute of NetDev, so several
NetDevs could have been primay. Setting is as a sting in NodeConf
assures that only one network could be primary.
If no primary network is set, a valid altvalue will be set.


anderbubble:
Good idea. What happens if the specified network interface doesn't exist?

mslacken:
> Good idea. What happens if the specified network interface doesn't exist?

The primary interface is used for the 
* hosts entry
* ssh connection
So If there is an invalid primary interface you don't have e.g.a node01 entry without suffix and you can't ssh into node via wwctl.
So you get what you want.

PR #685, opened 2023-02-14
Container recipe for leap

mslacken:


anderbubble:
@mslacken this PR appears to include much more than the LEAP docker container; is that intentional?

Also, I have https://github.com/hpcng/warewulf-node-images moved back into hpcng, and I'm trying to get CI/CD configured there. Could we move a new LEAP node image there in stead?

mslacken:
Branched of from the wrong branch. Should be fixed now,. Thanks for pointing this out.

mslacken:
Close as its now open in hpcng/warewulf-node-images/pull/6

PR #686, opened 2023-02-14
updated the documentation to match 4.4.0 features

mslacken:


anderbubble:
@mslacken great! Thanks for doing this.

Do you happen to know which of the issues in https://github.com/hpcng/warewulf/milestone/6 is addressed by this?

I'd like to get these changes into a 4.4.x branch as well, so that we could build a 4.4 version of the docs separately from the development version. Thoughts or concerns?

PR #688, opened 2023-02-15
Fix overlay list --long vs. --all output

### LINKED ISSUE #687: Erroneous overlay listing with --long --all
### Version of Warewulf

Development version from github

### Expected behavior

Overlay file listing with long-style format when `--long` option is provided.

### Actual behavior

Providing both `--long` and `--all` options to the `overlay list` command produces the long-style header but short-style listing of files.

```
$ wwctl overlay list --all
OVERLAY NAME                   FILES/DIRS  
debug                          /warewulf/   
debug                          /warewulf/template-variables.md.ww
generic                        /etc/        
generic                        /etc/group.ww
generic                        /etc/hosts.ww
generic                        /etc/passwd.ww
  :

$ wwctl overlay list --all --long generic
PERM MODE    UID GID   SYSTEM-OVERLAY     FILE PATH
generic                        /etc/        
generic                        /etc/group.ww
generic                        /etc/hosts.ww
generic                        /etc/passwd.ww
  :
```

The code in [internal/app/wwctl/overlay/list/main.go](internal/app/wwctl/overlay/list/main.go) handles the case where **one of the flags** has been provided, but not the case where **both flags** have been provided.

See previous section.

n/a

Built from repo clone.

jtfrey:
The `--long` flag implies `--all` but the inverse is not true.  Reorder the flag testing in the CobraRunE() function to produce expected output versus those two flags.

Output of `wwctl overlay list --all --long` consists of a long-style header and short-style items.  The header output vs. item output conditionals vs. CLI flags are not ordered the same, which produces this mismatch.

 - Fixes #687

### DIFF
--- a/internal/app/wwctl/overlay/list/main.go
+++ b/internal/app/wwctl/overlay/list/main.go
@@ -39,16 +39,11 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
 			files := util.FindFiles(path)
 
 			wwlog.Debug("Iterating overlay path: %s", path)
-			if ListContents {
-				var fileCount int
-				for file := range files {
-					fmt.Printf("%-30s /%-12s\n", name, files[file])
-					fileCount++
-				}
-				if fileCount == 0 {
-					fmt.Printf("%-30s %-12d\n", name, 0)
-				}
-			} else if ListLong {
+            
+            // The header (above) depends primarily on --long (which implies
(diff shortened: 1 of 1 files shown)


anderbubble:
Thanks for the PR, @jtfrey! There's already a pending PR for this at #667; but we should compare the two.

Issue #691, opened 2023-02-21
'wwctl node edit' inconsistent state with warewulfd


altuck00:
### Version of Warewulf

wwctl version:   4.4.0-1.git_afcdb21
rpc version: apiPrefix:"rc1" apiVersion:"1" warewulfVersion:"4.4.0-1.git_afcdb21"

### Expected behavior

When making changes to the node configuration using `wwctl node edit` the changes will be immediately reflected by warewulfd and appropriate ipxe/image/overlays will be applied to nodes.

### Actual behavior

After editing node configuration with `wwctl node edit` nodes become unrecognized by warewulfd during the provisioning step and receive the default iPXE script.

### Steps to reproduce this behavior

Configure and deploy a node, then edit the node with `wwctl node edit` and reboot it.

### What OS/distro are you running

```
NAME="Red Hat Enterprise Linux"
VERSION="9.1 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.1"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.1 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/9/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.1
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.1"
```

### How did you install Warewulf

Warewulf was compiled from afcdb21.

anderbubble:
@mslacken I'm thinking that warewulfd isn't getting poked correctly when the edit is applied. Any ideas?

mslacken:
The problem here is that the nodes are not modified with the edit command, but deleted and then redadded with the changed parameters. So this seems to be an unwanted side effect of this process.

anderbubble:
@mslacken and @JasonYangShadow this is still an issue, even after #719 and after #780 and #781.

anderbubble:
Demonstration of the issue as seen from the head node:

[...]

PR #719, opened 2023-02-28
more secure edit command

mslacken:
## More secure edit commands
As the node database may get modified during a node edit command, this
PR introduces a safe guard for this. At the time the node/profile edit
command starts a hash of the database is calclulated. This hash must match
the hash of the actual database as during the edit command (it opens an editor
and waits for user input) could have been changed (discoverable!).

Could be related to #691


JasonYangShadow (approved):
LGTM

PR #721, opened 2023-03-01
added wwctl genconf used to generate completions

mslacken:


mslacken:
@anderbubble please check internal/app/wwctl/node/add/main_test.go as this can be a biolder plate for more wwctl tests

anderbubble:
I added a couple fixes that were breaking the ability to build RPMs. Otherwise, looks like it's working to me.

PR #727, opened 2023-03-10
added experimental dnsmasq support

mslacken:
## adde dnsmasq support

Added a template which will create the necessary configuration files 
for the use of dnsmasq. Added also a short description in userdocs.


griznog:
I'd recommend adding a tag for warewulf, my dnsmasq dhcp-host entries look like this:

```
40:a6:b7:28:c4:79,set:warewulf,100.64.2.101,cpu-c-1,365d
40:a6:b7:28:b3:99,set:warewulf,100.64.2.102,cpu-c-2,365d
40:a6:b7:28:b4:c9,set:warewulf,100.64.2.103,cpu-c-3,365d
40:a6:b7:28:b5:19,set:warewulf,100.64.2.104,cpu-c-4,365d
```

Then the rest of the config can easily call out warewulf nodes while ignoreing other things that are getting DNS/DHCP support via warewulf, e.g. special configs for how to boot:

```
# Once iPXE calls, we answer...
# Boot with warewulf
dhcp-boot=tag:iPXE,tag:clusternet,tag:warewulf,"http://100.64.15.254:9873/ipxe/${mac:hexhyp}",100.64.15.254,100.64.15.254
dhcp-boot=tag:iPXE,tag:infiniband,tag:warewulf,"http://172.16.0.254:9873/ipxe/${mac:hexhyp}",172.16.0.254,172.16.0.254
dhcp-boot=tag:iPXE,tag:management,tag:warewulf,"http://100.64.47.254:9873/ipxe/${mac:hexhyp}",100.64.47.254,100.64.47.254
```

Those lines only match for clients that are iPXE AND the given network interface AND a warewulf node. 

Second suggestion is a workaround to a DNSMasq quirk. DNSMasq won't add a host for DNS resolution until _after_ it has a lease, so names won't resolve until the node is up. This causes some confusion to Slurm so its useful to also add the host records to DNSMasq, for example in `dnsmasq.conf`:

```
# Get hosts from additional files in this directory
hostsdir=/etc/dnsmasq/dnsmasq.hosts.d
```

Then in that directory a file with records (as from the example above):

```
100.64.2.101 cpu-c-1 cpu-c-1.cluster cpu-c-1.clusternet
100.64.2.102 cpu-c-2 cpu-c-2.cluster cpu-c-2.clusternet
100.64.2.103 cpu-c-3 cpu-c-3.cluster cpu-c-3.clusternet
100.64.2.104 cpu-c-4 cpu-c-4.cluster cpu-c-4.clusternet
```

This ensures the node names resolve in DNS before they are booted.

Doing all the above also means there is no longer any need to maintain `/etc/hosts` on the master or push a `/etc/hosts` to the nodes or keep it updated, as long as nodes `/etc/resolv.conf` points to the master everything will Just Work(tm).

Having all that in place also means there is now no longer a need to pass static IP address configs to the nodes, all interfaces can just DHCP all info including setting up `/etc/resolv.conf`, gateway, etc., eliminating a fair amount of complexity from the `wwinit` overlay/process.

griznog:
Advanced DNSMasq/WW integration, but worth mentioning, 

```
# DHCP Script options. 
dhcp-script=/etc/warewulf/scripts/dhcp-script.sh
```

Tells DNSMasq to call the given script, from the man page:

> Whenever a new DHCP lease is created, or an old one destroyed, or a TFTP file transfer completes, the executable specified by this option is run. 

A fair amount of useful info is passed to that script and it could be used as part of the discovery of new nodes, update WW about the node state via an API call, log activity to a debug log, etc....  There's also a `--dhpc-luascript=` option for calling a script written in lua for lua fans.

griznog:
> In the vanilla config Dnsmasq reads/etc/hosts were we add all the nodes. So dns resolution should be available as soon as the nodes are added to warewulf.

This is ok if nodes are not multihomed and/or DNSMasq will only be working for WW, but if integrating into an organizational DNSMasq that serves many interfaces/networks and stuff outside of WW it'd be better to remove the `/etc/hosts` dependency and let that revert to the well known `localhost` entries. 

In general passing around flat files like `/etc/hosts`, `/etc/passwd` and `/etc/groups` is fine for getting started or toy solutions, but any kind of organizational complexity in the real world quickly outgrows those. Given that for `/etc/hosts` DNSMasq offers a simple alternative that provides a lot of additional functionality it'd be nice to go ahead and use that to its fullest extent even in the most simple deployments. 

It'd be really awesome if DNSMasq was just used by default as the DHCP/TFTP/DNS provider for WW as it integrates all those functions nicely, is a well supported tool available in every distribution and can simply provisioning/managing nodes a lot.

mslacken:
Adds just stuff, so merging

PR #740, opened 2023-03-17
Replace containers with pointer to warewulf-node-images

anderbubble:
Replaces the node images in containers/ with a README that points to the new warewulf-node-images repository.


containers/README.md
```diff
@@ -0,0 +1,9 @@
+# Node images have moved
+
+Warewulf node images are now being maintained in a seprate repository:
```
mslacken, line 3: s/seprate/separated/

mslacken (changes requested):
lgtm

containers/README.md
anderbubble: I had actually meant for this to be "separate." Fixed. Thanks!

PR #754, opened 2023-03-23
Add missing packages to build on EL8

olafmersmann:

Following the instructions [here](https://warewulf.org/docs/development/quickstart/el8.html) will fail when running `make all` on a fresh Rocky Linux 8.7 install because `gpgme-devel` and `libassum-devel` aren't installed by default.


mslacken:
Thanks for you contribution.
For running warewulf in a production environment we recommend to use the rpms, as the development branch is unstable.

PR #766, opened 2023-04-04
add support to install efi and bios from system path

### LINKED ISSUE #689: Included static ipxe v1.21.1 lacks imgextract command
### Version of Warewulf

4.4.0 release

### Issue

The included ipxe binaries in staticfiles are built from a v1.21.1 tag in the ipxe project that for whatever reason lacks code for the imgextract command. This causes the warewulf ipxe script to fall back to old-style initrd, which causes boot failures when container/overlay size exceeds 3GB. This can be seen by the following error message output by the Warewulf ipxe script when it tries to unpack the images:

"Image extract not supported in this iPXE, using standard initrd mode" (from default.ipxe).

default.ipxe attempts to do imgextract first due to its obvious advantages over initrd, so it seems that using it is desired - but the included ipxe binaries lack the support due to tag v1.21.1 from ipxe being broken for this purpose. ipxe-update.sh also specifies this tag, and would similarly rebuild 'lacking' ipxe binaries.

Issue is alleviated by building ipxe binaries from ipxe/main, which includes the code for the imgextract command.

Rocky 8.7 x86_64

Built RPM from source using included specfile

JasonYangShadow:

add support to install efi and bios from system path.

 - Fixes #689 

#### Before submitting a PR, make sure you have done the following:

- Read the [documentation for contributing to Warewulf](https://warewulf.org/docs/development/contributing/contributing.html)
- Added changes to the [CHANGELOG](https://github.com/hpcng/warewulf/blob/development/CHANGELOG.md) if necessary
- Updated [userdocs](https://github.com/hpcng/warewulf/tree/development/userdocs) if necessary
- Based this PR against the appropriate branch (typically [development](https://github.com/hpcng/warewulf/tree/development/userdocs))
- Added myself as a contributor to the [Contributors File](https://github.com/hpcng/warewulf/blob/development/CONTRIBUTORS.md

### DIFF
--- a/ipxe-update.sh
+++ b/ipxe-update.sh
@@ -11,38 +11,101 @@ EFI=bin-${ARCH}-efi/ipxe.efi
 PCBIOS_output=`pwd`/staticfiles/${ARCH}.kpxe
 EFI_output=`pwd`/staticfiles/${ARCH}.efi
 
-set -xe
-
+SYSBIOS="/usr/share/ipxe/undionly.kpxe"
+SYSEFI="/usr/share/ipxe/ipxe-${ARCH}.efi"
 
 TMPDIR=`mktemp -d /tmp/ipxebuild.XXXXXX`
-if [ -f "$1" ] ; then
-  WRKDIR=`pwd`
-  cd "$TMPDIR"
-  tar xzf $WRKDIR/$1
-else
+
+usage() {
+  echo "Usage: $(basename $0) 
+        [-s] (install efi and bios from system location)
+        [-g] (build from git repo and install built efi and bios)
+        [-h] (help)
+        [-f] ipxe_source_gz_file (build from given tar.gz ipxe file and install built efi and bios)
+       "
+  exit 1
+}
+
+sysbuild() {
+  if [ -f "${SYSEFI}" ] ; then
+    cp $SYSEFI $EFI_output
+  else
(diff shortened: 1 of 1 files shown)


MatthewHink:
+1 LGTM Thank you :)

anderbubble:
@JasonYangShadow as I read this, this isn't what I was picturing. This seems to add support for pulling in the system ipxe files at Warewulf build / packaging time; but I think Warewulf itself should be updated to support specifying the location of the ipxe binaries at runtime.

This might already be done in 32ab50f299502fce7bf588852a75c63cf3332cf8, though I think a little bit of packaging cleanup might be in order. I'd advocate for not bundling iPXE at all, if there are no objections.

What do you think, @JasonYangShadow?

How about @mslacken?

griznog:
It'd be cool to have both `imgextract` and console framebuffer support in iPXE if it could be worked into the upstream OS packages for it, but it seems to be to be better to include a script, docs and/or examples for how to build that rather than bundling a custom build with each release of WW. There's also the parallel discussion around using a small shim initrd/initramfs which provisions the image via a different method rather than pulling it with iPXE, which would solve a lot of issues around image size problems.

There are a lot of other iPXE corner cases specific to local environments (VLANs? Security?, ...) and including a set of recipes in the documentation could probably go a long way toward demystifying this for WW admins without bloating what is delivered with WW in an attempt to solve all these in the WW package.

mslacken (approved):
Doesn't change the actual build, just the rebuld of the iPXE binaries./

PR #786, opened 2023-04-19
disk support for warewulf

mslacken:
Disk support is a desired feature for warewulf. Every implementation would
rely on the warewulf database to support disk structures. This PR implements
thenm so that these can be consumed by Ignition or CloudInit.
The logical yaml structure is closely inspired by the ignition yaml structure,
only the disk substructure is not implemented, as these structure only holds
the partition table information.


mslacken:
@anderbubble runs now without errors, can you merge it?

anderbubble:
I'll try to review. It sure is a big merge, though.

mslacken:
thanks a lot

anderbubble:
@mslacken this PR has a lot of strange branching and merging, including the same commits multiple times on different sub-branches that are later merged in different directions.

I tried to rebase it at https://github.com/anderbubble/warewulf/tree/DiskIgnition. If that looks good to you, could you update this PR with those commits? (The tests appear to be passing, so I don't think I've messed it up too badly.)

I'm still reviewing the PR even from that reference point. I'd like to split some of the commits so that they each have a single description of what they're doing and have the tests passing for each commit along the way. If you can help with that, great; but I'll keep working on it otherwise.

mslacken:
@anderbubble Thanks for your work on removing the empty and double commits. These were artifacts of my merge attempts.

We can split this PR in following components, which is for one part the 
* recursive getter/setter function 
* the more advanced commandline functions
* node and profile list advancements
* disk and partition data structure
* systemd scripts in wwinit

We can split these up, but there is a dependency of one parch on each other, where the most fundamental one are the patches for the recursive getter and setter function. They are critical for the functionality of warewulf, but are completely invisible for the warwulf users.

The single commits already describe accurately what is changed.

anderbubble:
> We can split these up, but there is a dependency of one [patch] on each other

My biggest concern is some interleaved bits that change netdev and nettag at the same time; but maybe I just am not yet following how that's just updating them to use new recursive getter/setter functionality.

internal/app/wwctl/node/add/main.go
```diff
@@ -22,17 +22,25 @@ func CobraRunE(vars *variables) func(cmd *cobra.Command, args []string) error {
...
-				delete(vars.nodeConf.NetDevs, "default")
-			}
 		}
+		delete(vars.nodeConf.NetDevs, "UNDEF")
+		if vars.fsName != "" {
+			fs := *vars.nodeConf.FileSystems["UNDEF"]
+			vars.nodeConf.FileSystems[vars.fsName] = &fs
+		}
+		delete(vars.nodeConf.FileSystems, "UNDEF")
+		/*
```
anderbubble: Commented-out code should not be added here.

internal/app/wwctl/node/add/main.go
```diff
@@ -22,17 +22,25 @@ func CobraRunE(vars *variables) func(cmd *cobra.Command, args []string) error {
 				return err
 			}
 		}
-		// remove the default network as all network values are assigned
+		// remove the UNDEF network as all network values are assigned
```
anderbubble, line 29: Why is the behavior of NetDevs being changed as part of this commit? It doesn't seem to relate to adding structures for devices and filesystems.

internal/pkg/node/flags.go
```diff
@@ -25,38 +25,40 @@ func (nodeConf *NodeConf) CreateFlags(baseCmd *cobra.Command, excludeList []stri
...
 				converters = append(converters, createFlags(baseCmd, excludeList, nestType.Elem().Field(j), &field)...)
 			}
-		} else if nodeInfoType.Elem().Field(i).Type == reflect.TypeOf(map[string]*NetDevs(nil)) {
-			netMap := nodeInfoVal.Elem().Field(i).Interface().(map[string]*NetDevs)
-			// add a default network so that it can hold values
-			key := "default"
-			if len(netMap) == 0 {
-				netMap[key] = new(NetDevs)
+		} else if nodeInfoType.Elem().Field(i).Type.Kind() == reflect.Map &&
+			nodeInfoType.Elem().Field(i).Type != reflect.TypeOf(map[string]string{}) { //reflect.TypeOf(map[string]*NetDevs(nil)) {
```
anderbubble: Commented-out code should not be left in the commit.

internal/pkg/node/flags.go
anderbubble, line 1: Do these changes relate to "adding structures for devices and filesystems"? It appears to largely be part of changes to NetDev. Maybe add a test to demonstrate the altered behavior here?

internal/pkg/node/methods.go
anderbubble, line 1: NewNetDevEntry is commented-out here, but still used. If it is to be removed here, and its use should be removed in the same commit. Still, this doesn't appear to relate to "added structures for devices and filesystems"; so I suggest that this should be done in a separate commit.

internal/pkg/node/transformer_test.go
```diff
@@ -141,6 +142,7 @@ func Test_nodeYaml_SetFrom(t *testing.T) {
 		}
 	})
 	t.Run("Get() default network mask", func(t *testing.T) {
+		fmt.Println(len(test_node2.NetDevs))
```
anderbubble, line 145: This debug output shouldn't be left in the commit.

internal/pkg/node/transformers.go
```diff
@@ -23,13 +24,96 @@ Populates a NodeConf struct from a NodeInfo, with the combined
...
-	nodeConf.getterFrom(nodeInfo, (*Entry).Get, (*Entry).GetSlice)
+	recursiveGetter(&nodeInfo, nodeConf, (*Entry).Get, (*Entry).GetSlice)
+}
+
+/*
+Abstract function which populates a NodeConf from the given NodeInfo
+via getter functions. Calls recursive itself for nested structures.
+Panics if the NodeConf has fields which are not type of string,[]string,map[string]*ptr
+*/
+func recursiveGetter(
```
anderbubble, line 35: Can you add some tests to demonstrate that recursiveGetter works correctly?

internal/pkg/node/transformers.go
```diff
@@ -127,45 +211,56 @@ func (nodeConf *NodeConf) getterFrom(nodeInfo NodeInfo,
...
-				// check if a network was deleted
-				if len(netMap) > len(nestedMap) {
-					for netMapKey := range netMap {
-						foundKey := false
-						for nestedMapKey := range nestedMap {
-							if netMapKey == nestedMapKey {
-								foundKey = true
+			} else if nodeInfoVal.Field(i).Type().Kind() == reflect.Map { //} reflect.TypeOf(map[string]*NetDevEntry{}) {
+				// check if a map element was deleted
+				/*
```
anderbubble: Remove commented-out code from the commit.

internal/pkg/node/transformers.go
```diff
@@ -127,45 +211,56 @@ func (nodeConf *NodeConf) getterFrom(nodeInfo NodeInfo,
...
-					netConfType := reflect.TypeOf(*netMap[netName])
-					netConfVal := reflect.ValueOf(netMap[netName])
-					for j := 0; j < netConfType.NumField(); j++ {
-						netVal := netValsType.Elem().FieldByName(netConfType.Field(j).Name)
+					if !confField.MapIndex(entryMapIter.Key()).IsValid() {
+						newPtr := reflect.New(confField.Type().Elem().Elem())
+						confField.SetMapIndex(entryMapIter.Key(), newPtr)
+					}
+					confVals := entryMapIter.Value().Elem()
+					/*
```
anderbubble: Remove commented-out code from the commit.

internal/pkg/node/transformers.go
```diff
@@ -276,8 +373,7 @@ func (node *NodeInfo) setterFrom(n *NodeConf, nameArg string,
 						} else if nestedInfoVal.Elem().Field(j).Type() == reflect.TypeOf(map[string](*Entry){}) {
 							confMap := nestedVal.Interface().(map[string]string)
 							if nestedInfoVal.Elem().Field(j).IsNil() {
-								ptr := nestedInfoVal.Elem().Field(j).Addr().Interface().(*map[string](*Entry))
```
anderbubble, line 279: changes to setterFrom don't appear to relate to "added recursive getter function." They should likely be moved to the following commit.

internal/pkg/node/transformers.go
```diff
@@ -325,116 +144,95 @@ func (node *NodeInfo) SetDefFrom(n *NodeConf) {
...
 }
 
 /*
 Abstract function which populates a NodeInfo from a NodeConf via
-setter functionns.
+setter functions. Panics if other type than string, []string *ptr is used in NodeConf.
 */
-func (node *NodeInfo) setterFrom(n *NodeConf, nameArg string,
-	setter func(*Entry, string, string),
+func recursiveSetter(source, target interface{}, nameArg string, setter func(*Entry, string, string),
```
anderbubble, line 167: I suggest adding tests for recursiveSetter, like recursiveGetter.

internal/pkg/node/transformers.go
anderbubble: Really, I think recursiveGetter and recursiveSetter (later) are symptoms of the whole Conf vs Entry dichotomy being much too complicated. I have some ideas for how to fix that; but, for now, these methods end up being huge and complex, and I think they need additional test functions in the suite to make sure they're behaving the way they're intended to.

internal/pkg/node/datastructure.go
anderbubble, line 1: Changes to the datastructure should be made in their final form earlier in the PR.

overlays/wwinit/warewulf/init.d/30-ignition
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
```
anderbubble: If "simple init.d ignition fails as systemd is needed" why is it included here?

anderbubble (changes requested):
I won't claim this is a complete review; but I'm afraid I find a lot of this difficult to follow, which is making it difficult for me to confidently review.

There are specific comments on several changes; but here are some holistic thoughts:

* The test suite should pass for each commit in the PR
* New functions, especially new large, complex functions, should come with new tests in the suite
* New functions, especially when they replace existing functions, should be added in their own commit with explanatory comments in the commit message about the change.
* Code should not be left commented-out in a commit, but should be removed
* Commits that do multiple things (e.g., add partial support for disk/part/fs and change netdev) should be split into multiple commits (i.e., one that adds support for disk/part/fs, and another that changes netdev)
* The PR shouldn't include commits that fix / alter changes made in previous commits in the PR. Such changes should be squashed into a single commit that includes the final form only.

~jonathon

internal/app/wwctl/node/add/main.go
mslacken: Right,  it doesn't relate directly.
I found a behavior for the NetDevs while adding the tests covering the recursive getter and setters. 
The tests before were aimed for the transformation from NodeConf to NodeYaml and back, the new tests are using the "real" `wwctl` commands, so the changes are combined here.

internal/pkg/node/flags.go
mslacken: As above, I had to fix the errors revealed by the added /changedtesting framework.

anderbubble:
@mslacken I've done my best to complete a review of this PR. My (extensive) rebase is over at https://github.com/anderbubble/warewulf/tree/DiskIgnition. Please check it out and push it to your branch (for this PR) if you think it looks good.

Features of the rebase:

* I broke commits apart and tried to sort and squash them appropriately.
* I audited the commit history and made sure each had an up-to-date `go.mod` and `go.sum`.
* Every commit passes tests now.
* I tried to move new tests earlier when possible so that they would cover more changes.
* I updated commit messages for formatting and grammar and, in some instances, clarity.
* I fixed a few bugs I found.

There are two commits still present that I _suspect_ are technically unrelated:

* https://github.com/anderbubble/warewulf/commit/deb54fdd4e8deec3f8e2156a49dc863a1b86a356
* https://github.com/anderbubble/warewulf/commit/623427eadae04fe617f170e7611c0b135a64d524

I'd be interested in your confirmation of whether these are actually related to the PR. Regardless, it might be worth checking over the updated commit messages to ensure I've reflected them properly.

@jcsiadal I think you're the better person to review this functionally. I'm fine merging it for you to test via `development`; but let me know if you'd rather try it out before merge.

mslacken:
> There are two commits still present that I _suspect_ are technically unrelated:
> 
>     * [anderbubble@deb54fd](https://github.com/anderbubble/warewulf/commit/deb54fdd4e8deec3f8e2156a49dc863a1b86a356)
> 
>     * [anderbubble@623427e](https://github.com/anderbubble/warewulf/commit/623427eadae04fe617f170e7611c0b135a64d524)
> 

Right these commits are technically unrelated, but as I encountered them during the creation of this PR I decided to fix them in place.
N.B. [anderbubble@deb54fd](https://github.com/anderbubble/warewulf/commit/deb54fdd4e8deec3f8e2156a49dc863a1b86a356) is missing the changes in `contructor,go`, but tehse are melded into other and will be hard to get out
> 

Thanks for attributing the commit messages.

anderbubble (approved):
I think this is reasonably ready for merge, though we should add tests for the generated json, and @jcsiadal will test functionality and usability when merged for development.

anderbubble (changes requested):
@mslacken I was going to go ahead and merge this in and open a new issue for adding tests for the generated ignition json; but could you just add such tests to this PR?

anderbubble:
> The change I am referring to, must have fallen under the table. I will fix this, when this PR is merged in.
Its basically just read out the primary netdev and set it accordingly in the constructor.

If there's a change missing here, could you go ahead and update the PR with the necessary code and add tests to match? (I'm concerned that the missing code didn't trigger any failing tests.)

mslacken:
The unrelated PRs are now #897 and #898

anderbubble:
@mslacken I added a couple new tests to at least get some coverage of createDiskIgnition and the new disk metadata.

jcsiadal:
My testing is on Rocky 9.2 (5.14.0-284.25.1.el9_2.x86_64) with build 4.5.x-1.git_84e97d2. The test node uses a 1TB HDD, no partitions, but a preexisting DOS partition table. 
I attempted to add a scratch drive using the command line in the docs (under content). I added it using profile instead of node. It did create the systemd services and fstab updates, but failed to load the service. When I attempted to start the service manually (as root), it failed with this error [systemctl status output cleanup by me]
```
Ignition failed: create partitions failed: create partitions failed: exit status 2:
Cmd: "sgdisk" "--zap-all" "/run/ignition/dev_aliases/dev/sda"
Stdout:
Warning! Disk size is smaller than the main header indicates! Loading
secondary header from the last sector of the disk! You should use 'v' to
verify disk integrity, and perhaps options on the experts' menu to repair
the disk.
GPT data structures destroyed! You may now partition the disk using fdisk or
other utilities.
Stderr:
Caution: invalid backup GPT header, but valid main header; regenerating
backup header from main header.

Warning! One or more CRCs don't match. You should repair the disk!
Main header: OK
Backup header: ERROR
Main partition table: OK
Backup partition table: ERROR
Invalid partition data!
```
Rerunning it failed with the same error.
I ran sgdisk --zap-all "/run/ignition/dev_aliases/dev/sda" manually. It ran successfully, and after that, the ignition-ww4 service started successfully. We need to determine why the same command failed under systemd but ran fine manually.

If WW4 ignition is creating multiple systemd services, it's preferable for it to also create mountpoints and mount the partitions automatically by default. I was able to do these manually after the service started.
```

mslacken:
Thanks for testing @jcsiadal
I think the problem here is that there might have been a problem with the disk. Perhaps it may be useful to run `sgdisk --zap-all ` before ignition is launched.

PR #792, opened 2023-04-20
Configure ONBOOT with netdev.OnBoot

### LINKED ISSUE #644: document --onboot
It appears that

```
ONBOOT=yes
```

is set statically in the interface configuration. Investigate what the `--onboot` argument is actually doing and document.

anderbubble:

`netdev.OnBoot` wasn't actually configuring ifcfg ONBOOT, which was set statically to true.

 - Fixes #644


anderbubble:
@mslacken I think this means that devices that don't have onboot set to true will default to false, as I suspect that's what EL and Suse do when it's not specified. I think that's technically correct, but it is a change, and perhaps a surprising one. Should we make the default value true in the template?

PR #781, opened 2023-04-15
Refactor package that handles warewulf.conf

anderbubble:
## Description of the Pull Request (PR):

Refactor warewulfconf for clarity, consistency, documentation, and testing.

PR #808, opened 2023-05-04
fix warewulf.conf location for wwclient

mslacken:
## fix loading of warewulf.conf for wwclient
#781 introduced a new API for reading warewulf.conf. This new API wasn't
implemented in wwclient. This PR introduces this new API to wwclient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   file is create and will executed after the `ignition-disks-ww4.service` has finished.
   Entries in `/etc/fstab` for every file system are created with the `noauto` option.
 
+- wwclient has now a commandline switch for the location of warewulf.conf
 ## [4.4.0] 2023-01-18
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,19 +40,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The primary hostname and warewulf server fqdn are now the canonical name in
   `/etc/hosts`
-- new subcommand `wwctl genconf` is available with following subcommands:
-  * `completions` which will create the files used for bash-completion. Also
-     fish an zsh completions can be generated
-  * `defaults` which will generate a valid `defaults.conf`
-  * `man` which will generate the man pages in the specified directory
-  * `reference` which will generate a reference documentation for the wwctl commands
-  * `warwulfconf print` which will print the used `warewulf.conf`. If there is no valid
-     `warewulf.conf` a valid configuration is provided, prefilled with default values
-     and an IP configuration derived from the network configuration of the host
-- All paths can now be configured in `warewulf.conf`, check the paths section of of 
-   `wwctl --emptyconf genconfig warewulfconf print` for the available paths.
-- Added experimental dnsmasq support.
+
 - Refactored `profile add` command to make it alike `node add`. #658 #659 
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,19 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-     fish an zsh completions can be generated
-  * `defaults` which will generate a valid `defaults.conf`
-  * `man` which will generate the man pages in the specified directory
-  * `reference` which will generate a reference documentation for the wwctl commands
-  * `warwulfconf print` which will print the used `warewulf.conf`. If there is no valid
-     `warewulf.conf` a valid configuration is provided, prefilled with default values
-     and an IP configuration derived from the network configuration of the host
-- All paths can now be configured in `warewulf.conf`, check the paths section of of 
-   `wwctl --emptyconf genconfig warewulfconf print` for the available paths.
-- Added experimental dnsmasq support.
```
anderbubble, line 37: Why the removals here?

CHANGELOG.md
```diff
@@ -48,25 +38,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-- new subcommand `wwctl genconf` is available with following subcommands:
-  * `completions` which will create the files used for bash-completion. Also
-     fish an zsh completions can be generated
-  * `defaults` which will generate a valid `defaults.conf`
-  * `man` which will generate the man pages in the specified directory
-  * `reference` which will generate a reference documentation for the wwctl commands
-  * `warwulfconf print` which will print the used `warewulf.conf`. If there is no valid
-     `warewulf.conf` a valid configuration is provided, prefilled with default values
-     and an IP configuration derived from the network configuration of the host
```
anderbubble, line 63: And why the removals here?

internal/app/wwclient/root.go
```diff
@@ -16,8 +16,8 @@ import (
 
 	"github.com/coreos/go-systemd/daemon"
 	"github.com/google/uuid"
-	"github.com/hpcng/warewulf/internal/pkg/pidfile"
 	warewulfconf "github.com/hpcng/warewulf/internal/pkg/config"
```
anderbubble, line 19: It looks to me like the change from #781 _is_ present here, so I'm confused by the PR description.

internal/app/wwclient/root.go
```diff
@@ -31,14 +31,16 @@ var (
 		RunE:         CobraRunE,
 		SilenceUsage: true,
 	}
-	DebugFlag bool
-	PIDFile   string
-	Webclient *http.Client
+	DebugFlag       bool
+	PIDFile         string
+	Webclient       *http.Client
+	WarewulfConfArg string
```
anderbubble, line 37: It looks to me like this PR is actually adding configuration to specify the config file, not adjusting for #781. Have I missed something?

anderbubble (commented):
Would it be possible to add at least a smoke test for wwclient so that such breakage would be caught in the future?

CHANGELOG.md
mslacken: Also duplicated entries

internal/app/wwclient/root.go
mslacken: I assume that since #781 a configuration file has to be specified, before it was specified implicitly. But this could also have been introduced by me earlier. Another fix would be to add to warewulf conf.Get() a implicit default configuration file, but as Get is only called twice, besides the tests, I think this is the better solution.

anderbubble: Agreed. Thanks for the clarification.

PR #811, opened 2023-05-09
allow rendering of host templates

### LINKED ISSUE #623: wwctl overlay show --render for host templates
`wwctl overlay show --render` supports rendering templates for "nodes" but not for the host overlay. There should be a way to do that.

mslacken:
Allow rendering of host templates where `-r host` or even `-r $(uname -n)` are now 
valid flags for `-r`.

- Closes #623


anderbubble:
Implements #623

anderbubble:
Closes #623

JasonYangShadow:
@mslacken I am testing this PR
it looks like the code will always fail.
```
NodeName != "host" || NodeName != localHostname
```
the value of  `localHostname` of my test env is `localhost.localdomain`. 
from the code, this condition will always be true, because NodeName can not equal both `host and localHostname`?

mslacken:
> @mslacken I am testing this PR it looks like the code will always fail.
> 
> ```
> NodeName != "host" || NodeName != localHostname
> ```
> 
> the value of `localHostname` of my test env is `localhost.localdomain`. from the code, this condition will always be true, because NodeName can not equal both `host and localHostname`? this will always generate the following error
> 
> ```
> ERROR  : xxx does not identify a single node
> ```

Right, I am just a physcist so I hoped this could be solved by a schrödingers inequality

mslacken:
merged via #1153

Issue #505, opened 2022-07-29
Need a template function that includes another template and evaluates its content


jmechalas:
RIght now, the various Include functions in Warewulf's template support include a file or block, but do not evaluate the included template's content: it's just brought in as text. We need an option to nest templates, so that if template A includes template B, template B gets evaluated as well.

mslacken:
Nested templates should be possible, but I never saw a sensible use case. Can you elaborate on this? I am just curious.

jmechalas:
Take /etc/fstab as an example. You can't have the same fstab on all systems because it might havepieces that are specific to a particular hardware platform. But you might also add, say, NFS volumes, and those NFS mounts are mostly the same across an environment.

So, you would start with something like this in an overlay:

```
# This file is autogenerated by warewulf 
# Host:   {{.BuildHost}}
# Time:   {{.BuildTime}}
# Source: {{.BuildSource}}
rootfs / tmpfs defaults 0 0
devpts /dev/pts devpts gid=5,mode=620 0 0
tmpfs /run/shm tmpfs defaults 0 0
sysfs /sys sysfs defaults 0 0
proc /proc proc defaults 0 0
# local storage mounts
/dev/nvme0n1p1 /local/mount ext4 defaults 1 1
# NFS mounts
server:/path/to/exports /nfs/somepath nfs defaults 0 0
```

Now, the first chunk of that applies to all Warewulf-booted nodes, and the last chunk is generic, so you can reduce it to this:

```
{{ Include /path/to/warewulf/fstab/header }}
# local storage mounts
/dev/nvme0n1p1 /local/mount ext4 defaults 1 1
{{ Include /path/to/NFS/mounts }}
```

But, what if my NFS mounts _aren't_ common to all systems? I may have two different OS's that require a different libc, which means tool directories (e.g. /usr/local/bin or whatever) are going to vary across the platforms. So now I need a different NFS mounts for each of those platforms.

Except, I don't want to maintain two copies of NFS exports because that is an accident waiting to happen. What I really want to do is have one master NFS mount list, and use the templating system so that I can choose which export I want based on the client's configuration. And that means my `Include /path/to/NFS/mounts` has to be a template which gets processed as well e.g. `/path/to/NFS/mounts.ww`.

mslacken:
Quite interesting trick.
I will look into, but most likely there will be a limit of nested templates you can call, as this might easy create endless loops/

jmechalas:
[...]

PR #814, opened 2023-05-09
render template files several times

mslacken:
This makes it possible to use something like {{ .IPAddress }} in variables,
so that e.g. the kernel commandline parameter can contain a reference to 
the ip address of the warewulf master host

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added experimental dnsmasq support.
 - Check for formal correct IP and MAC addresses for command line options and
   when reading in the configurations
-
+- render template several (up to 8 times) if template refrences like '{{ foo }}' are
+  found in a rendered template. This make it possible to use e.g. the ip address of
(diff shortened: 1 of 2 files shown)


anderbubble:
What if, in stead of rendering the whole template several times, we just process (some?) template variable values as templates themselves once before applying them to the overall template?

griznog:
I've been using this in my `ipxe` script for a while now to get the masters info into the kernel command line and it's always worked fine, not sure I understand what I'm missing out on with the way it already worked? Would this make it possible (assuming it's not already possible) to use the node ipaddr? That's actually a requirement to use the `dracut` networking to set up networking from a dracut produced `initrd` via the kernel command line. 

```
boot kernel initrd=initrd.magic rootfs=tmpfs rootfstype=tmpfs wwid={{.Id}} wwcluster={{.Cluster}} wwcontainer={{.ContainerName}} {{.KernelArgs}} 
```

mslacken:
close #505

anderbubble (changes requested):
@mslacken if you'd still like this included, can you rebase on main?

PR #815, opened 2023-05-11
download rendered overlays via http

mslacken:
Adds a simple http server on the warewulf port which is rooted at
the $OVERLAYIDR/wwroot/. Files with the .ww suffix are rendered as 
templates where the node can be set via the 
`http://${wwinit_ip}:${wwinit_port}/overlay/test?node=$NODE` 
url.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix issue that initrd fails at downloading runtime overlay with permission denied error,
   when warewulf secure option in warewulf.conf is enabled. #806
 - Allow iPXE to continue booting without runtime overlay. #806
+- Block unprivileged requests for arbitrary overlays in secure mode.
+- new subcommand `wwctl genconf` is available with following subcommands:
+  * `completions` which will create the files used for bash-completion. Also
+     fish an zsh completions can be generated
+  * `defaults` which will generate a valid `defaults.conf`
+  * `man` which will generate the man pages in the specified directory
+  * `reference` which will generate a reference documentation for the wwctl commands
+  * `warwulfconf print` which will print the used `warewulf.conf`. If there is no valid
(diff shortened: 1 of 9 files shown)


mslacken:
depends on #786 in order to compile

anderbubble:
Is there any security on this? Similar to the `secure: true` function, I would want to see a way to prevent regular users from accessing arbitrary overlays and, perhaps, to prevent nodes from seeing each other's overlays.

mslacken:
> Is there any security on this? Similar to the `secure: true` function, I would want to see a way to prevent regular users from accessing arbitrary overlays and, perhaps, to prevent nodes from seeing each other's overlays.
For the actual implementation only files which are under `$WWOVERLYADIR/wwwroot` are delivered. Also the secure port is honored.

griznog:
Curious what does this add we didn't already have by just pulling overlays from the warewulfd port directly?

mslacken:
> Curious what does this add we didn't already have by just pulling overlays from the warewulfd port directly?

At the moment we are pulling a cpio archive of all files in the overlay. This will allow us to pull a single file from an overlay which than can be used as configuration file for e.g. cloud-init.

griznog:
I think I'd like an option to serve static files through `warewulfd` but every time I think about what I'd do with it all I can come up with are bad ideas that I don't think I should do. Could the class of use cases like `cloud-init` be better solved by moving all `cloud-init` (or whatever) files into a dedicated overlay so the smaller and more focused overlay could be pulled as a complete entity as proposed in https://github.com/warewulf/warewulf/pull/1030?

anderbubble:
> This will allow us to pull a single file from an overlay which than can be used as configuration file for e.g. cloud-init.

I think we can use [NoCloud][1] to read cloud-init configuration from a local file.

[1]: https://cloudinit.readthedocs.io/en/stable/reference/datasources/nocloud.html

mslacken:
Cloud init is just an example. There are a lot of other tools which can consume cofigurations from http[s] resources.

anderbubble:
Replaced by #1659.

PR #822, opened 2023-05-19
Init rootfs fixes

elfurbe:

This PR refactors `/init` in the wwinit overlay to more properly address the rootfs option.

After a discussion with @anderbubble about tmpfs vs ramfs vs rootfs, and some light reading about [how the kernel deals with the initramfs](https://www.kernel.org/doc/html/latest/filesystems/ramfs-rootfs-initramfs.html#what-is-rootfs), and poking around in the provisioning environment with a Rocky 8 kernel (4.18.0-425.19.2.el8_7.x86_64), it is apparently true that if a kernel is configured with `CONFIG_TMPFS=y` then it will default to establishing the initial rootfs as a tmpfs implicitly. This makes the default value of `initramfs` for  the `root` property for nodes/profiles a non sequitur, as `initramfs` no longer appears to me to have any intrinsic meaning. This refactoring attempts to address that reality.

For the purposes of this implementation, I've assumed if the `root` property is set to `initramfs`, the implication is the user is indifferent and wants to use whatever the kernel chooses, or has ensured their rootfs is what they want through other means outside the scope of warewulf. I've also accounted for what seems to me to be the logical future wherein the default value of the `root` parameter would be unset, and added a catch condition that outputs messages informing the user that the `initramfs` option is deprecated and to unset the option if their intent is to trust what the kernel provides, but init still proceeds with a normal non-pivoting chain to `/warewulf/wwinit` thus remaining functionally transparent in the interim. This work only accounts for `tmpfs` and `ramfs` as possible options for the `root` property, and abuses that fact to provide the fstype option to mount using the derived $WWROOT value in `/warewulf/config`. This seemed reasonably sane to me, as there does not seem to be any practical value today to having abstracted/arbitrary values de-reference to actual filesystem types. Obviously this logic is not complicated, so that could be addressed sensibly at a later date when something like stateful provisioning may require it.

It is worth noting that if this PR is accepted the way I've written it, there would be a recommended corresponding change in the Go codebase that generates the `defaults.conf` configuration file to remove the default value for the `root` property. I am not a go programmer, so I am not undertaking to make that change, but it does appear to me from casual reading to be reasonably trivial.

It is also worth noting that these changes remove the informational message about the kernel parameter `rootfstype=ramfs` as it does not seem to be necessary if you're committed to doing the mirroring operation from the default rootfs to a new rootfs on your explicitly provided fstype. From my testing, that kernel parameter can be used to influence the default fstype used by the kernel, and if you wanted to _not_ deal with these two different ram-based filesystem options in `init`, an alternative solution would be to use kernel arguments to make the kernel's implicit behavior explicit/deterministic. It would definitely be more efficient and require less boot time for large images by avoiding the copying step, but I think handing this in the init process will best mesh with a potential future case where the new rootfs may be on local media.

 - Fixes #

There is no existing issue.

#### Before submitting a PR, make sure you have done the following:

- [x] Read the [documentation for contributing to Warewulf](https://warewulf.org/docs/development/contributing/contributing.html) 
- [x] Added changes to the [CHANGELOG](https://github.com/hpcng/warewulf/blob/development/CHANGELOG.md) if necessary
- [ ] Updated [userdocs](https://github.com/hpcng/warewulf/tree/development/userdocs) if necessary
- [x] Based this PR against the appropriate branch (typically [development](https://github.com/hpcng/warewulf/tree/development/userdocs))
- [x] Added myself as a contributor to the [Contributors File](https://github.com/hpcng/warewulf/blob/development/CONTRIBUTORS.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Return an error during `wwctl container import` if archive filename includes a colon. #1371
 - Correctly extract smbios asset key during Grub boot. #1291
+- Refactor of `wwinit/init` to more properly address rootfs options. #1098
 
 ## v4.5.7, 2024-09-11
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -38,3 +38,4 @@
 * Tobias Poschwatta <poschwatta@zib.de>
 * Josh Burks <jeburks2@asu.edu>
 * Elmar Pruesse <pruessee@njhealth.org> @epruesse
+* Adam Michel <elfurbe@furbism.com> [@elfurbe](https://github.com/elfurbe)
--- a/overlays/wwinit/rootfs/init
+++ b/overlays/wwinit/rootfs/init
@@ -47,7 +47,7 @@ if [ "${WWROOT}" != "${ROOTFSTYPE}" ]; then
         echo "Unrecognized rootfs type requested: ${WWROOT}"
         echo "Rebooting in 1 minute..."
         sleep 60
-        /sbin/reboot
+        /sbin/reboot -f
     fi
 else
     echo "Invoking /warewulf/wwinit..."
@@ -57,4 +57,4 @@ fi
 echo
 echo "There was a problem with the initial provisioning process, rebooting in 1 minute..."
 sleep 60
-echo b > /proc/sysrq-trigger || /sbin/reboot
+echo b > /proc/sysrq-trigger || /sbin/reboot -f
--- a/overlays/wwinit/rootfs/init
+++ b/overlays/wwinit/rootfs/init
@@ -43,7 +43,7 @@ if [ -n "${WWROOT}" ] && [ "${WWROOT}" != "${ROOTFSTYPE}" ]; then
     elif [ "${WWROOT}" == "initramfs" ]; then
         echo ">>> Deprecated option \"initramfs\" chosen."
         echo ">>> The 'initramfs' can be either a ramfs or a tmpfs natively, depending on kernel configuration."
-        echo ">>> If you wish to use whatever this kernel chooses by default, remove this 'root' setting from your nodes/profiles."
+        echo ">>> If you wish to use whatever this kernel chooses by default, remove the 'root' setting from your nodes/profiles."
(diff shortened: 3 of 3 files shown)


anderbubble:
@griznog I'd love to get your thoughts on this.

griznog:
Really wish I'd taken better notes when I slogged through this over a year ago. The one thing I do recall is that it is all really confusing and I think I reached the same conclusion as @elfurbe only less eloquently stated. Simplifying this would be a win, but since I'm currently still using my hacked up image that skips `wwinit` and boots straight to `system` for `init` on this initial `tmpfs` from unpacking the `initrd`,  I can't easily test it. For my use  case I do explicitly set `rootfs=tmpfs rootfstype=tmpfs` but I don't think it's strictly necessary that I do that, it's just where I was at when it started working and I walked away.

This is also tied into the long term goal (at least for me) to have each node image/container build its own small initrd with dracut and have that initrd pull the full image at boot time rather than pulling it in during the ipxe stage. That's a pretty massive change in the flow of booting a node though, like a WW5 change.

overlays/wwinit/init
```diff
@@ -26,31 +26,38 @@ mount -t tmpfs tmpfs /run
...
+    exec /warewulf/wwinit
 fi
 
 echo
-echo "There was a problem with the provisioning process, rebooting in 1 minute..."
+echo "There was a problem with the initial provisioning process, rebooting in 1 minute..."
 sleep 60
-/sbin/reboot
\ No newline at end of file
+/sbin/reboot -f
```
mslacken, line 63: `/sbin/reboot` is a softlink to `systemctl` and may working without `systemd` running, or does the `-f` switch override this. @elfurbe can you please test this? 
An alternative is `echo b > /proc/sysrq-trigger` which always works.

mslacken (changes requested):
Test the reboot command.

anderbubble:
Rebased to main.

anderbubble:
@elfurbe I'm so sorry I let this languish for so long. I've got fixing this on the current milestone, and intend to base the final fix on your changes. I might branch off to a new PR; but I want to make sure we retain your contribution.

Thank you!

anderbubble:
Fixed changlog entry.

mslacken (approved):

anderbubble:
@mslacken this PR shouldn't've been merged; please see #1401.

PR #854, opened 2023-06-14
Build fixes for API

MatthewHink:
Updated the linter from 1.50 to 1.53.2 because the previous version was falling over (likely OOM). Updated protoc build command for newer version of protobuf.

The make command below should work now.
make clean setup proto all build wwapid wwapic wwapird ; echo $?

Did not update build dependencies. We should consider updating supported golang versions. From https://go.dev/doc/devel/release
Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including critical security problems, in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

At this time we should consider supporting only go versions 1.20 and 1.19. That is not in this commit for fear of making a breaking change.


mslacken (commented):
add white space

MatthewHink:
It turns out it works with or without whitespace. I also had a chown issue on my dev box which could have been the way I installed golang.

MatthewHink:
Added whitespace

mslacken (approved):

MatthewHink:
Would like to merge this. Sorry I have been a bum and missing the meetings.  Want to help. How can I?

anderbubble:
Porting this comment from a commit to the PR:

> Did not update build dependencies. We should consider updating supported golang versions.
> From https://go.dev/doc/devel/release
> Each major Go release is supported until there are two newer major releases.
> For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release.
> We fix critical problems, including critical security problems,
> in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).
> 
> At this time we should consider supporting only go versions 1.20 and 1.19.
> That is not in this commit for fear of making a breaking change.

anderbubble:
Ordinarily I'd be concerned about the changes to `internal/pkg/api/routes/wwapiv1/routes.pb.go` et al; but I see that these are actually generated files. This leads me to thinking that these files should be removed from the repository and only be built; but since they depend on protobuf applications that are not in Rocky, I _also_ don't want to do that. So, for now, I'll leave as-is.

Issue #784, opened 2023-04-18
stop bundling ipxe with warewulf


anderbubble:
Warewulf bundles ipxe, but with recent revelations that Rocky Linux and Suse Linux each provide an iPXE images package and frequent requests for changes to the bundled iPXE build (#681, #689, #236), along with the ability to configure where Warewulf gets its iPXE images from (32ab50f) we should just stop bundling iPXE. At most, we can just bundle [ipxe-update.sh](https://github.com/hpcng/warewulf/blob/development/ipxe-update.sh) or some derivative of it.

* [ ] remove ipxe binaries from the source repository
* [ ] remove ipxe binaries from Makefile
* [ ] remove ipxe binaries from specfile(s)
* [ ] update specfile to depend on rocky/suse iPXE image packages
* [ ] update default warewulf config to point to suse/rocky iPXE image location

anderbubble:
@mslacken I'll plan to work on this; but anything I've missed in the checklist?

PR #859, opened 2023-06-21
grub netboot implementation

mslacken:
added a simple grub.cfg in order to replace iPXE


anderbubble:
I'm reviewing this now, and rebased against current origin/development to resolve conflicts.

anderbubble (changes requested):
Forgive the long-winded review here: I have a lot of things.

I don't like that this implementation makes the default profile special, and restricts what you can do with it. (To summarize: grub boot needs an EFI shim, and this implementation assumes to get it from the container assigned to the default profile.)

My first thought was "why get it from the container at all?" After all, the grub boot shim isn't that different from the ipxe bootloader: today we package that with Warewulf, but we're planning (#784) to stop doing that, in favor of using the ipxe packages included with, say, Suse or Rocky. Rocky has a similar shim package (`shim-x64`) for grub, and I presume Suse does, too; so why can't we just use that, and leave the container alone?

I _think_ the answer is secure boot: it's not enough to have a "working" shim; we need a shim that has keys to validate that the eventual grub, kernel, etc is valid. So we really do need the shim that's in the container, not just any old shim.

But, if so, I think that means that this implementation, of just using the shim from the container of the default profile, is insufficient: what if we want to be able to provision two different operating systems that have incompatible requirements for the boot shim?

If my understanding is correct, I think this needs to work pretty much exactly how kernels work today: if you're doing a grub boot, the shim should be detected at boot time and fed from the container that it's about to boot, without having to copy the shim somewhere or treat any specific container special. (I haven't looked in depth regarding how kernel auto-detection works yet, so I presume it's more complex than that; but the end result is the same: Warewulf doesn't use the kernel from the container of the default profile; it uses the kernel from the container it's booting.)

This analogy may extend to the ability to import shims _into_ warewulf, like how we support importing kernels into warewulf and specifying them explicitly; but I expect it's less necessary to support arbitrary shims per container. (i.e., it doesn't appear to be normal to have multiple versions of the shim installed in the container)

To summarize, I think the following changes should be made:

* Don't treat the default profile specially, or introduce any new limitations on in
* Read the shim dynamically from the container being booted in the same way that we read the kernel from the container

anderbubble:
@mslacken thanks for reminding me in the meeting today that the use of the default container is only for the node discovery case / unknown node case.

In this case, could we just use the shims from the host operating system?

mslacken:
Using the shim and grub of the host is also possible and maybe the better alternative.

CHANGELOG.md
```diff
@@ -100,6 +100,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   profile. #852, #876, #883, #896, #903
 - During node discovery, prefer the primary network interface, if defined. #775
 
+- default is now protected and can't be deleted
+- first container imported container is added to the default profile
```
anderbubble, line 104: If this is using the host shim now, can all the special-casing around the default profile and its container be removed?

anderbubble (changes requested):

internal/pkg/warewulfd/warewulfd.go
```diff
@@ -14,6 +15,21 @@ import (
...
+	mux http.Handler
+}
+
+/*
+Filter out the '//'
+*/
+func (h *slashFix) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.URL.Path = strings.Replace(r.URL.Path, "//", "/", -1)
+	h.mux.ServeHTTP(w, r)
+}
```
anderbubble, line 32: This is probably fine; but maybe it should only filter out for calls to `http://efiboot//`?

internal/pkg/warewulfd/sentDB.go
anderbubble, line 1: Can you add a package-level comment about what sentDB is for? 

It looks to me like this is entirely for inspection during testing. Is that accurate? If so, I do not think this is how this should be tested. That each part has been sent should be verified by the test itself by acting as an http client and inspecting what is received.

Is warewulfd multi-threaded? The lock management here would seem to indicate so; but what happens if more than one node is being served at a time? Wouldn't that corrupt this global `sentDB` hash?

internal/pkg/warewulfd/parser.go
```diff
@@ -25,16 +27,22 @@ func parseReq(req *http.Request) (parserInfo, error) {
...
+	hwaddr := ""
+	if stage != "efiboot" {
+		hwaddr = path_parts[2]
+		hwaddr = strings.ReplaceAll(hwaddr, "-", ":")
+		hwaddr = strings.ToLower(hwaddr)
+	} else if len(path_parts) > 3 {
+		ret.efifile = strings.Join(path_parts[2:], "/")
+	} else {
+		ret.efifile = path_parts[2]
+	}
```
anderbubble, line 45: What if path_parts > 3 and stage == efiboot?

At least, "unknown path components in GET" probably needs to be adjusted. (It was probably inaccurate before, too.)

anderbubble, line 45: You don't need `if len(path_parts) > 3` here. `strings.Join(path_parts[2:], "/")` produces the desired output in both cases.

internal/pkg/overlay/datastructure.go
```diff
@@ -49,19 +50,19 @@ func InitStruct(nodeInfo *node.NodeInfo) TemplateStruct {
...
+		wwlog.Warn("Problems opening nodes.conf: %s", err)
 	}
-	allNodes, err := nodeDB.FindAllNodes()
+	tstruct.AllNodes, err = nodeDB.FindAllNodes()
 	if err != nil {
-		wwlog.Error("%s", err)
-		os.Exit(1)
+		wwlog.Warn("couldn't get all nodes: %s", err)
+	}
+	tstruct.ProfileMap, err = nodeDB.MapAllProfiles()
```
anderbubble, line 59: I don't see any overlays using ProfileMap. Why is it being added as part of this PR?

internal/pkg/warewulfd/parser.go
```diff
@@ -76,7 +86,11 @@ func parseReq(req *http.Request) (parserInfo, error) {
 		return ret, errors.New("no stage encoded in GET")
 	}
 	if ret.hwaddr == "" {
-		return ret, errors.New("no hwaddr encoded in GET")
+		ret.hwaddr = ArpFind(ret.ipaddr)
```
anderbubble, line 89: ARP will only work if we're on the same Ethernet subnet as the target host. Is this usually encoded as hwaddr in GET? If not, I'm concerned by any dependence on ARP, as that would restrict Warewulf from being used behind a DHCP relay.

internal/pkg/node/constructors.go
```diff
@@ -31,6 +31,8 @@ defaultnode:
   init: /sbin/init
   root: initramfs
   ipxe template: default
+  grub template: grub.cfg.ww
+  boot method: ipxe
```
anderbubble, line 35: This is great (I was requesting a generic "boot method" or "bootloader" option, and a per-node setting, earlier; but is this actually used anywhere? I don't see `boot method` anywhere else in the code.

internal/pkg/container/shim.go
anderbubble, line 1: I haven't put too much thought into it; but it seems like shim.go and grub.go could be genericized and combined a bit.

anderbubble (changes requested):
- It looks like the commit history contains the previous implementation attempt that is later undone. I'd prefer a clean commit history; but I understand that the best way to do that is not particularly intuitive. I have a method I got from a friend; but perhaps I should get that method published before I hold people accountable to it.
- It looks like there's still some special-casing of the default profile left behind from the first implementation attempt.
- If I understand correctly that `sentDB` is entirely for the purpose of the test suite, I really think it should be removed and behavior should be validated some other way during testing. Code should be written so as to be testable; but not by introducing global state to the main application.
- Other things. (See individual comments.)

internal/app/wwctl/container/imprt/main.go
```diff
@@ -13,6 +15,9 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	if len(args) == 2 {
 		name = args[1]
 	}
+	if list, _ := container.ListSources(); len(list) == 0 {
+		SetDefault = true
+	}
```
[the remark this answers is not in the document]
mslacken: No its not obsolete as this makes it easier for newbies to get in touch with warewulf. And the fist container is *always* the default one.

internal/app/wwctl/profile/delete/main.go
```diff
@@ -13,6 +14,9 @@ import (
 
 func CobraRunE(cmd *cobra.Command, args []string) error {
 	var count int
+	if util.InSlice(args, "default") {
+		return fmt.Errorf("can't delete the `default` profile ")
+	}
```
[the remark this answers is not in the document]
mslacken: Still do the special case or we will have to remove the switch`--setdefault` for `container import`.

internal/pkg/api/profile/profile.go
```diff
@@ -24,7 +24,15 @@ func ProfileSet(set *wwapiv1.ProfileSetParameter) (err error) {
...
 		return errors.Wrap(err, "Could not open database")
 	}
-	return apinode.DbSave(&nodeDB)
+
+	var pConf node.NodeConf
+	err = yaml.Unmarshal([]byte(set.NodeConfYaml), &pConf)
+	if err != nil {
+		wwlog.Error(fmt.Sprintf("%v", err.Error()))
+		return
+	}
```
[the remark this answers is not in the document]
mslacken: No debugging but it checks that we can unmarshall the var `set.NodeYaml`, but the check is duplicated so we can remove.

mslacken:
> * It looks like the commit history contains the previous implementation attempt that is later undone. I'd prefer a clean commit history; but I understand that the best way to do that is not particularly intuitive. I have a method I got from a friend; but perhaps I should get that method published before I hold people accountable to it.
>

I know having a clean commit history looks good and professional, but leaving the dead ends in the commits also shows that the authors tries alternative methods and that this methods failed. A null result is also a result. Or are you a secret member of the TVA?
 
>     * It looks like there's still some special-casing of the default profile left behind from the first implementation attempt.
> 
We have special casing for the default profile which can't removed right now, e.g. `container import --setdefault`. The code added here just makes this more obvious. 
>     * If I understand correctly that `sentDB` is entirely for the purpose of the test suite, I really think it should be removed and behavior should be validated some other way during testing. Code should be written so as to be testable; but not by introducing global state to the main application.
> 
As explained below, the sentDB isn't used right now, but will be used in the future for measured boot. I want to leave it int he code base right, as if it resides not in development branch the work may get lost.
>     * Other things. (See individual comments.)
I wlll update the PR

internal/pkg/warewulfd/sentDB.go
```diff
@@ -0,0 +1,144 @@
...
+values can be used for measured boot in combination with
+TPM devicces
+*/
+var sentDB map[string]*SentFiles
+
+// mutex for locking the map
+var mu sync.Mutex
+
+func init() {
+	sentDB = map[string]*SentFiles{}
```
JasonYangShadow, line 40: not sure the current use case falls into the usage of sync.map
https://pkg.go.dev/sync#Map
 (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or 
 (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

internal/pkg/warewulfd/util.go
```diff
@@ -36,7 +38,14 @@ func sendFile(
 		filename,
 		stat.ModTime(),
 		fd)
-
+	// seek back
+	_, err = fd.Seek(0, 0)
```
JasonYangShadow, line 42: might use fd.Seek(0, io.SeekStart) for a better understanding here.

JasonYangShadow:
lgtm! I do not have more comments. Thanks for the hard work.

mslacken:
@JasonYangShadow  can you just approve, so that I can merge?

JasonYangShadow (approved):

JasonYangShadow:
@anderbubble I do not have more comments. It's up to other community members to approve if it's okay to be merged or not. cc @mslacken

anderbubble:
Rebased to main to continue testing and see if the checks clear.

CHANGELOG.md
```diff
@@ -125,6 +125,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support importing containers with symlinked `/bin/sh` #797
 - Don't panic on malformed passwd #527
 
+- first container imported container is added to the default profile
+- grub in combination can now be set as boot method with `warewulf.grubboot: true` in
```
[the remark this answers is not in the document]
anderbubble: I set this in `warewulf.conf`

```
warewulf:
  port: 9873
  secure: false
  update interval: 60
  autobuild overlays: true
  host overlay: true
  syslog: false
  grubboot: true
```

But I still got an iPXE boot. What did I do wrong? (I also restarted warewulfd.)

anderbubble (commented):
I haven't been able to boot my node with Grub yet. What am I doing wrong?

CHANGELOG.md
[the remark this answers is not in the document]
anderbubble: I realized that I hadn't rebuilt the dhcp config. Now I've done that and am at least getting an error. Investigating.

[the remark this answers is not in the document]
anderbubble: I fixed a couple errors in the dhcp template, but I still don't have grub booting successfully.

Screenshot 2023-12-23 at 10 16 21 PM

I don't think my "fixes" are the cause of the issue; but let me know if you see what I'm doing wrong. @mslacken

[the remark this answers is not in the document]
anderbubble: I didn't have grub or shim installed in the container, but only the host. But after installing that, still no change.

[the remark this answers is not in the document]
anderbubble: The documented grub package to be installed was incorrect: should be `grub2-efi-x64`.

I'm further now: still not a successful boot, though.

Screenshot 2023-12-23 at 10 59 22 PM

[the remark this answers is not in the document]
anderbubble: Screenshot 2023-12-23 at 11 08 30 PM

mslacken:
@anderbubble I have now tested the code again and found that there is a timeout which grub can't handle, if 
1. the overlays are not readily available
2. the container image isn't built
as then warewuld isn't sending out the images in the right timeframe. I think we don't see this behavior for iPXE as there, if the compressed couldn't be fetched, iPXE makes a failover to uncompressed images, which gives warewulfd enough time to build them.
So in order to use the grub boot you will have to run `wwctl overlay /container build` before boot.

I was also able to boot the rocky image from a openSUSE ww4-host.

CHANGELOG.md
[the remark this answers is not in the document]
JasonYangShadow: I have a similar scrrenshot
image
image

JasonYangShadow:
oh, confirmed that 
https://github.com/hpcng/warewulf/pull/859#discussion_r1450008563
issue listed here, can be temporally resolved by
```
wwctl container build --force NAME
wwctl overlay build
wwctl server restart
```
my local compute nodes can successfully pass `kernel panic` issue and continue.

anderbubble:
@mslacken I made a couple changes that ordinarily I'd want to ask you about; but I think by this point you're more interested in getting this merged than anything, so I'm going to go ahead and do that. Of course, if you disagree with any of my changes we can do a different (presumably smaller, easier) PR.

anderbubble (approved):

PR #863, opened 2023-06-26
Fix lsmod wrong spelling

ArnaudLcm:
lsmod command is misspelled in the Vagrant Documentation part.


mslacken (approved):
Thanks, good catch

PR #871, opened 2023-06-27
reboot also without systemd

mslacken:
## reboot also without systemd

/sbin/reboot needs systemd to be running, so reboot the hard^W easy way

### DIFF
--- a/overlays/wwinit/init
+++ b/overlays/wwinit/init
@@ -11,7 +11,7 @@ if test -f "/warewulf/config"; then
 else
     echo "ERROR: Warewulf configuration file not found... rebooting in 1 minute"
     sleep 60
-    /sbin/reboot
+    echo b > /proc/sysrq-trigger || /sbin/reboot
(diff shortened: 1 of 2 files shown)


overlays/wwinit/init
```diff
@@ -53,4 +53,4 @@ fi
 echo
 echo "There was a problem with the provisioning process, rebooting in 1 minute..."
 sleep 60
-/sbin/reboot
\ No newline at end of file
+echo b > /proc/sysrq-trigger || /sbin/reboot
```
anderbubble, line 56: Shouldn't it be the other way around? Try to reboot with /sbin/reboot or, if that fails, use sysrq?

mslacken:
As we don't have systemd at this time, `/sbin/reboot` will fail, so the `sysrq-trigger` does *always* the job.
It should never ever fail and if yes, trying a `/sbin/reboot` doesn't make things worse.

PR #880, opened 2023-07-25
griznog fix build errors

griznog:
- More aggressive cleaning based on contents of .gitignore.
- Replace deprecated io.ioutil functions with new os versions.

Fix assorted build errors I encountered building on Rocky 9.2. Not sure teh OS
matters, but that is it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New documentation for development environment (Vagrant)
 
 ### Fixed
-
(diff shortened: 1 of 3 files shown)


griznog:
Commit e05438f fixes this error by pointing the command at the `warewulf.conf` from the build:
```
./wwctl genconfig completions > /etc/bash_completion.d/wwctl
ERROR: Couldn't parse IP address: invalid CIDR address: 100.64.15.254
make: *** [Makefile:206: files] Error 255
make install failed.
```
But 100.64.15.254 is a valid IP address, so that might also need to be addressed in `wwctl`?

Issue #852, opened 2023-06-12
simpler overlay names


anderbubble:
I suggest that Warewulf should build per-node overlays using static names like "system" or "wwinit" or "runtime" (within their existing per-node directories).

The current form of naming built overlays as a hypenated list of all the contained overlays doesn't scale well when many overlays are used simultaneously.

griznog:
@mslacken is the one who knows for sure, but I think the long name one is just the merged version of all the individual overlays which is pulled during boot. The individual overlays  can be built individually with `wwctl overlay build -O overlayname` if they aren't there already.

It gets messy in WWPROVISION location if you add/remove overlays often, as those merged ones stack up, but that's just a part of the larger issue of `wwctl` not doing a lot of housekeeping anywhere for old versions of things. 

It'd be neat if it just built the individuals, then merged them on the fly when provisioning. I played around a bit with having them come in individually in the ipxe script, but that also doesn't scale well.

anderbubble:
> I think the long name one is just the merged version of all the individual overlays which is pulled during boot.

Agreed. But this long name is what shows up in `wwctl status` for example, and it gets unwieldy.

anderbubble:
See also #876.

anderbubble:
Fixed in #883, though there's a bug. See #896

PR #883, opened 2023-07-27
Improved overlay image naming

### LINKED ISSUE #876: 'File name too long' when building overlay for a node with a lot of overlays.
### Version of Warewulf

4.3.0-1.git_235c23

### Expected behavior

Overlays will be built.

### Actual behavior
```
DEBUG  : cpio: Cannot open /datapool/warewulf/provision/overlays/cpu-vm-1/autofs-backups-chrony-common-common-bhn-cron-dnsmasq-duo-fail2ban-firewalld-gdm-globus-gpu-httpd-libvirt-lustre-mlocate-motd-munge-namespaces-networking-nhc-nomachine-packages-pam-postfix-profile.d-root-rsyslog-samba-sbgrid-selinux-slurm-sshd-storage-setup-timezone-zfs-znapzend.img: File name too long
```

Add a lot of overlays to a node config.

```sh
# cat /etc/os-release 
NAME="Rocky Linux"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
```

Built from source.

@mslacken I found the limit of the current approach to naming the merged overlay for provisioning :D

griznog:

This addresses the system/runtime overlay image naming by adding an optional
argument to OverlayImage() to specify the context in which an image name is
being constructed. If set to "legacy" then the old name joining method is used.
Any other value passed is used to construct a short image name. For example,
the node config:

```
nodeprofiles:
  default:
    container name: rocky-9
    runtime overlay:
      - generic
      - debug
    system overlay:
       - wwinit
    kernel:
      override: rocky-9
      args: selinux=0 rootfs=tmpfs consoleblank=600
nodes:
  cpu-a-1:
    profiles:
    - default
    network devices:
      clusternet:
        device: eth0
        hwaddr: 28:80:23:e8:7b:f0
        ipaddr: 100.68.0.66
        primary: "true"
```

Produces the overlay images:

```
[root@micropterus overlays]# tree
.
├── cpu-a-1
│   ├── debug.img
│   ├── debug.img.gz
│   ├── generic.img
│   ├── generic.img.gz
│   ├── __RUNTIME__.img
│   ├── __RUNTIME__.img.gz
│   ├── __SYSTEM__.img
│   └── __SYSTEM__.img.gz
```

Functions that use OverlayImage() have been updated to pass the context along
where applicable.

BuildOverlay() gets an optional argument so it can pass the context through.

BuildSpecificOverlays() has been changed to loop over the overlay list, calling
BuildOverlay() for a single overlay at a time and defaulting to the legacy
naming, which for a single overlay just names the image OVERLAYNAME.img.

ProvisionSend() conditionally sends `legacy` for the image if the querystring
specifed an `overlay=` value, otherwise the stage is used, which results in the
`__RUNTIME__.img` or `__SYSTEM__.img` image names.

Because OverlayImage uses the passed img_context to construct the name, it's
open-ended for constructing merged overlays images with short, specific names.

 - Fixes #876 #852

### DIFF
--- a/internal/app/wwctl/overlay/imprt/main_test.go
+++ b/internal/app/wwctl/overlay/imprt/main_test.go
@@ -2,7 +2,6 @@ package imprt
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"testing"
 
@@ -13,7 +12,7 @@ import (
 )
 
 func Test_List(t *testing.T) {
-	tmpdir, err := ioutil.TempDir(os.TempDir(), "warewulf")
+	tmpdir, err := os.MkdirTemp(os.TempDir(), "warewulf")
 	if err != nil {
 		t.Errorf("Could not create temp folder: %v", err)
 		t.FailNow()
@@ -34,7 +33,7 @@ func Test_List(t *testing.T) {
 		t.FailNow()
 	}
 
-	file, err := ioutil.TempFile(tmpdir, "file")
+	file, err := os.CreateTemp(tmpdir, "file")
 	if err != nil {
 		t.Errorf("Could not create tempfile")
 		t.FailNow()
--- a/internal/pkg/overlay/overlay.go
+++ b/internal/pkg/overlay/overlay.go
@@ -47,13 +47,13 @@ func BuildAllOverlays(nodes []node.NodeInfo) error {
 
 		sysOverlays := n.SystemOverlay.GetSlice()
 		wwlog.Info("Building system overlays for %s: [%s]", n.Id.Get(), strings.Join(sysOverlays, ", "))
-		err := BuildOverlay(n, sysOverlays)
+		err := BuildOverlay(n, sysOverlays, "system")
 		if err != nil {
 			return errors.Wrapf(err, "could not build system overlays %v for node %s", sysOverlays, n.Id.Get())
 		}
 		runOverlays := n.RuntimeOverlay.GetSlice()
 		wwlog.Info("Building runtime overlays for %s: [%s]", n.Id.Get(), strings.Join(runOverlays, ", "))
-		err = BuildOverlay(n, runOverlays)
+		err = BuildOverlay(n, runOverlays, "runtime")
 		if err != nil {
 			return errors.Wrapf(err, "could not build runtime overlays %v for node %s", runOverlays, n.Id.Get())
 		}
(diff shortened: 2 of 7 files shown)


griznog:
I guess I still don't know how to pull those commits fixing my build errors out of other PRs, but I can't build/test without them so I guess lumping them in isn't totally awful.

internal/pkg/overlay/overlay.go
```diff
@@ -139,10 +140,18 @@ func OverlayInit(overlayName string) error {
...
 Build the given overlays for a node and create a Image for them
 */
-func BuildOverlay(nodeInfo node.NodeInfo, overlayNames []string) error {
+func BuildOverlay(nodeInfo node.NodeInfo, overlayNames []string, img_context ...string) error {
+	var context string
+	/* Check optional context argument. If missing, default to legacy. */
+	if len(img_context) == 0 {
+		context = "legacy"
+	} else {
+		context = img_context[0]
```
JasonYangShadow, line 149: if we are only using first item of img_context, shall we directly declare img_context as `string` ?
we just need to check whether img_context is set or not.

internal/app/wwctl/overlay/imprt/main_test.go
```diff
@@ -34,7 +33,7 @@ func Test_List(t *testing.T) {
 		t.FailNow()
 	}
 
-	file, err := ioutil.TempFile(tmpdir, "file")
+	file, err := os.CreateTemp(tmpdir, "file")
```
JasonYangShadow, line 36: can we also add test cases for the verification of overlay names?
to check if it is the format of `__RUNTIME__**` and `__SYSTEM__**`.

internal/pkg/overlay/overlay.go
griznog: My extremely limited understanding is that the optional arguments in a variadic function will always be in an array as there's no way to know how many things will be passed to it.

JasonYangShadow (approved):

JasonYangShadow:
@anderbubble this PR looks good to me.

mslacken (approved):
Thanks @griznog and thanks for the review @JasonYangShadow

PR #889, opened 2023-08-09
Restore %config(noreplace) to nodes.conf

anderbubble:
The previous PR #888 erroneously removed the %config(noreplace) macro from the nodes.conf file. This PR restores it.


mslacken:
lgtm

PR #890, opened 2023-08-12
Makefile refactor

anderbubble:
I've found the warewulf Makefile and related build system somewhat difficult to follow for a while. This PR represents my review of `Makefile`, `warewulf.spec`, et al, along with some edits aimed at making it easier to understand and maintain.


Makefile
```diff
@@ -203,117 +102,103 @@ files: all
...
+#    Extract the contents and change in the directory
+#    ./configure
+#    make
+#    make check
+#    sudo make install
+#    sudo ldconfig # refresh shared library cache.
+# To setup protoc-gen-grpc-gateway, see https://github.com/grpc-ecosystem/grpc-gateway
+.PHONY: proto
+proto:
 	protoc -I /usr/include -I internal/pkg/api/routes/v1 -I=. \
```
mslacken, line 177: protoc needs following two binaries to run, `protoc-gen-go`  and `protoc-gen-go-grpc`. They can be installed with 
```
go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
```
Please add an appropriate rule, like there is one for the linter.

mslacken (changes requested):

mslacken:
Nice work, btw.

Makefile
anderbubble: @mslacken I captured all four, in fact. Let me know what you think.

PR #894, opened 2023-08-16
calling the overlay build functions direclty

mslacken:
The warewulfd process called `wwctl build -O [system|runtime] node`
if autobuild was set. This created [system|runtime].img which isn't
supported with the build model. Calling now BuildOverlay directly

### DIFF
--- a/internal/pkg/warewulfd/util.go
+++ b/internal/pkg/warewulfd/util.go
@@ -1,9 +1,11 @@
 package warewulfd
 
 import (
+	"fmt"
 	"net/http"
 	"os"
 
+	"github.com/hpcng/warewulf/internal/pkg/node"
 	nodepkg "github.com/hpcng/warewulf/internal/pkg/node"
 	"github.com/hpcng/warewulf/internal/pkg/overlay"
 	"github.com/hpcng/warewulf/internal/pkg/util"
@@ -61,21 +63,22 @@ func getOverlayFile(
 	}
 
 	if build {
-		wwlog.Serv("BUILD %15s, overlays %v", nodeId, stage_overlays)
-
-		args := []string{"overlay", "build"}
-
-		for _, overlayname := range stage_overlays {
-			args = append(args, "-O", overlayname)
+		nodeDB, errNested := node.New()
+		if err != nil {
+			return stage_file, errNested
 		}
-
-		args = append(args, nodeId)
-
-		out, err := util.RunWWCTL(args...)
-
+		myNode, errNested := nodeDB.FindAllNodes()
+		if err != nil {
+			return stage_file, errNested
+		}
+		myNode = node.FilterByName(myNode, []string{nodeId})
+		if len(myNode) != 1 {
+			return stage_file, fmt.Errorf("couldn't find node %s", nodeId)
+		}
+		err = overlay.BuildOverlay(myNode[0], context, stage_overlays)
 		if err != nil {
 			wwlog.Error("Failed to build overlay: %s, %s, %s\n%s",
-				nodeId, stage_overlays, stage_file, string(out))
+				nodeId, stage_overlays, stage_file, err)
 		}
 	}
 
--- a/internal/pkg/warewulfd/util.go
+++ b/internal/pkg/warewulfd/util.go
@@ -1,7 +1,6 @@
 package warewulfd
 
 import (
-	"fmt"
 	"net/http"
 	"os"
 
@@ -44,16 +43,15 @@ func sendFile(
 }
 
 func getOverlayFile(
-	nodeId string,
+	n node.NodeInfo,
 	context string,
 	stage_overlays []string,
 	autobuild bool) (stage_file string, err error) {
 
-	stage_file = overlay.OverlayImage(nodeId, context, stage_overlays)
+	stage_file = overlay.OverlayImage(n.Id.Get(), context, stage_overlays)
 	err = nil
-
 	build := !util.IsFile(stage_file)
-
+	wwlog.Verbose("stage file: %s", stage_file)
 	if !build && autobuild {
 		build = util.PathIsNewer(stage_file, nodepkg.ConfigFile)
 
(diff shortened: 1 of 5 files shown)


anderbubble:
@mslacken I'd like to rebase this after #903, but I expect that PR will require some consideration. Let me know what you think.

anderbubble:
#903 was merged. @mslacken can you rebase on current development?

mslacken:
@anderbubble can you please recheck again , especially `util_test.go` as before refactoring the error code of the wwctl call wasn't checked.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Container file gids are now updated properly during syncuser. #840
 - Fix build for API.
 - Bindd mounts for `wwctl exec` are now on the same fs
+- Fix automatic build of overlays
```
anderbubble, line 42: Should indicate the context in which automatic build was failing (e.g., when fetching an overlay with warewulfd).

internal/pkg/warewulfd/util_test.go
```diff
@@ -2,34 +2,40 @@ package warewulfd
...
+	{"specific overlay", "node1", "", []string{"o1"}, "p/overlays/node1/o1.img", false},
+	{"multiple specific overlays", "node1", "", []string{"o1", "o2"}, "p/overlays/node1/o1-o2.img", false},
+	{"multiple specific overlays with context", "node1", "system", []string{"o1", "o2"}, "p/overlays/node1/__SYSTEM__.img", false},
 }
 
 func Test_getOverlayFile(t *testing.T) {
+	wwlog.SetLogLevel(wwlog.VERBOSE)
 	conf := warewulfconf.Get()
 	conf.Paths.WWProvisiondir = "p"
+	var n node.NodeInfo
```
anderbubble, line 34: This should be defined within the `for` loop to ensure we don't leak data from one test to another.

internal/pkg/warewulfd/util.go
```diff
@@ -34,24 +35,23 @@ func sendFile(
...
+		fd)
 
 	wwlog.Send("%15s: %s", sendto, filename)
 
 	return nil
 }
 
 func getOverlayFile(
-	nodeId string,
+	n node.NodeInfo,
```
anderbubble, line 46: I'd really rather this be something other than `n`. It takes me so much longer to understand what some of the Warewulf code is doing because of all these single-character variables. (I've tended to use `nodeInfo` in similar circumstances.)

internal/pkg/warewulfd/util_test.go
```diff
@@ -2,34 +2,40 @@ package warewulfd
...
-	node string
-	context string
-	overlays []string
-	result string
-	succeed bool
+	node        string
+	context     string
+	overlays    []string
+	result      string
+	succeed     bool // getOverlayFile will always fail if no overlay dir is defined!
```
anderbubble, line 18: Why did this behavior change?

internal/pkg/warewulfd/util_test.go
```diff
@@ -2,34 +2,40 @@ package warewulfd
...
-	{"multiple specific overlays", "node1", "", []string{"o1", "o2"}, "p/overlays/node1/o1-o2.img", true},
-	{"multiple specific overlays with context", "node1", "system", []string{"o1", "o2"}, "p/overlays/node1/__SYSTEM__.img", true},
+	{"empty", "", "", nil, "", false},
+	{"empty node", "node1", "", nil, "", false},
+	{"specific overlays without node", "", "", []string{"o1", "o2"}, "p/overlays/node1/o1-o2.img", false},
+	{"system overlay", "node1", "system", nil, "p/overlays/node1/__SYSTEM__.img", false},
+	{"runtime overlay", "node1", "runtime", nil, "p/overlays/node1/__RUNTIME__.img", false},
+	{"specific overlay", "node1", "", []string{"o1"}, "p/overlays/node1/o1.img", false},
+	{"multiple specific overlays", "node1", "", []string{"o1", "o2"}, "p/overlays/node1/o1-o2.img", false},
+	{"multiple specific overlays with context", "node1", "system", []string{"o1", "o2"}, "p/overlays/node1/__SYSTEM__.img", false},
```
anderbubble, line 27: How is this supposed to work if none of these succeed any more?

anderbubble (changes requested):
It looks like all the tests were adjusted to just say "these are broken now and that's ok" which _isn't_ right. Are you asking me to investigate why they're broken now? I'll do that; just want to be clear on your expectations.

CHANGELOG.md
mslacken: Its always hard to put something sensible in the CHANGELOG if only a refactoring is happerning

internal/pkg/warewulfd/util.go
mslacken: right, short names can be confusing.

mslacken:
> It looks like all the tests were adjusted to just say "these are broken now and that's ok" which _isn't_ right. Are you asking me to investigate why they're broken now? I'll do that; just want to be clear on your expectations.
> 

The core problem is that the return status of the `wwctl overlay build` wasn't checked, so the error was there, but wasn't checked.
> edit: I think I understand now. These tests didn't do as much as I thought they did: apparently I developed the more advanced techniques after this set. I'll work on getting these updated, and check back with you for any questions.
Yes, thats the point. The tests didn't really test what they pretended to test. For real tests we would need the stuff which is already in `provision_test.go`. We also have then rearrange the tests here in order to be sensible.

internal/pkg/warewulfd/util_test.go
mslacken: They couldn't fail in the previous version.

internal/pkg/warewulfd/util_test.go
mslacken: see the other comments

mslacken:
@anderbubble I have now updated the tests also the existing ones, as we direct calls the behavior changed, especially errors are now coming back, but also on more cases overlays can be built.

anderbubble (approved):

PR #897, opened 2023-08-17
Epehermal container mounts on same fs as chroot

mslacken:
Previously ephemeral container mounts were performed in /tmp. Now these
mounts are overlayed on the same fs as is used for chroots as fs overlays
need to be on the same fs.

Signed-off-by: Christian Goll <cgoll@suse.de>


anderbubble:
@mslacken is this ready to go, or do you need to make another change ([as indicated in PR #786](https://github.com/hpcng/warewulf/pull/786#issuecomment-1681073189))?

PR #898, opened 2023-08-17
Disable per-netdev primary flag

mslacken:
Primary network devices are now identified by a per-node attribute,
rather than a per-netdev boolean. While the legacy data can still be
parsed, the argument for setting the per-netdev boolean at the
command-line is no longer available.
Correct behavior is tested.


anderbubble:
@mslacken is this ready to go, or do you need to make another change ([as indicated in PR #786](https://github.com/hpcng/warewulf/pull/786#issuecomment-1681073189))?

griznog:
How do we get to this new tag in a template?

PR #900, opened 2023-08-22
Updated Makefile to account for vendor/ in dist

anderbubble:
The previous PR #890 introduced a `Defaults.mk` target for defining default variables for future calls to `make`. This functionality was already present in a different form, but this new "real" target had an unintended side-effect of causing the target to be called automatically at reference, even the optional reference `-include Defaults.mk`. This is undesirable as it causes the generation of defaults even during a `make clean`.

This commit resolves these issues and further refactors the Makefile, building on the work in #890, but with additional understanding:

* Convert the `Defaults.mk` target to the "phony" `defaults` target, preventing automatic generation
* Remove `make clean` from `warewulf.spec`, expecting the contents of the dist to be appropriate
* Refactor and split `clean` targets
* Update `cleanvendor` to reflect `OFFLINE_BUILD`, similar to the `vendor` target
* Move "tools" targets into `Tools.mk`
* Add `ARCH` to `Variables.mk` for architecture-specific actions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ COPY . /warewulf-src
 RUN cd /warewulf-src &&\
   make contclean Defaults.mk \
     PREFIX=/usr \
-    BINDIRa=/usr/bin \
+    BINDIR=/usr/bin \
     SYSCONFDIR=/etc \
     DATADIR=/usr/share \
     LOCALSTATEDIR=/var/lib \
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,9 @@ RUN zypper  -n install --no-recommends git go1.18 libgpgme-devel &&\
 COPY . /warewulf-src
 
 RUN cd /warewulf-src &&\
-  make contclean Defaults.mk \
+  make clean \
+    OFFLINE_BUILD=1 &&\
+  make defaults \
     PREFIX=/usr \
     BINDIR=/usr/bin \
     SYSCONFDIR=/etc \
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,10 @@ include Variables.mk
 include Tools.mk
 
 .PHONY: build
-build: wwctl wwclient man_pages wwapid wwapic wwapird etc/defaults.conf etc/bash_completion.d/wwctl
+build: wwctl wwclient wwapid wwapic wwapird etc/defaults.conf etc/bash_completion.d/wwctl
+
+.PHONY: docs
+docs: man_pages reference
 
 vendor:
 ifndef OFFLINE_BUILD
@@ -81,7 +84,7 @@ test-cover: $(config)
 	go tool cover -html=coverage/cover.out -o=coverage/cover.html
 
 .PHONY: install
-install: all
+install: build docs
 	install -d -m 0755 $(DESTDIR)$(BINDIR)
 	install -d -m 0755 $(DESTDIR)$(WWCHROOTDIR)
 	install -d -m 0755 $(DESTDIR)$(WWPROVISIONDIR)
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: all
 all: build
-	
+
 include Variables.mk
 include Tools.mk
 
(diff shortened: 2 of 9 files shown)


mslacken:
This branch includes also all the changes of DiskIgnition, which was merged in #786 
Also a simple 
```
make
```
doesn't build the `all` target and
```
make wwctl
```
will always rebuild `wwctl` as the config target is always updated.

anderbubble:
> This branch includes also all the changes of DiskIgnition, which was merged in https://github.com/hpcng/warewulf/pull/786

I had this branched from the wrong commit. Sorry about that. Fixed.

> make doesn't build the all target

Fixed. It was picking up a target from the include and using that as the default. I moved `ALL` above the includes so that it would be picked up by default. We could also specify `.DEFAULT_GOAL := all`, which is more explicit and probably less fragile; but I haven't done that for now.

> make wwctl will always rebuild wwctl as the config target is always updated.

This actually got me to learn a new practice / technique: thanks! I converted the dependency list for `config` to a variable and used that within the makefile rather than the "PHONY" `config` target, which produces the expected behavior. (I left the `config` target in place as well, since it's referenced in scripts and documentation. Over time, though, it should become unnecessary. In fact, it probably is even now.)

mslacken:
Looks much better now, but still the target `man_pages` is always rebuilt, if here was a change or not.
Can you please move the `man_pages` target to the `install` section, as  I assume as man pages are only sensible when installed on a system.

mslacken (changes requested):
move `man_pages` to install

anderbubble:
> I assume as man pages are only sensible when installed on a system.

I disagree. Generating the man pages locally is still useful, because you can point `man` directly at a man page file; so those could be built, inspected, and references even without `install`.

> still the target man_pages is always rebuilt, if here was a change or not

I don't think there's a reasonable way around this. Since man pages are generated by `wwctl genconfig man`, the list of files generated isn't known to the Makefile. As such, it doesn't know what files might be rebuilt, so it can't know what files to check to see they need to be rebuilt.

We could "fix" this by including the list of files generated in the Makefile. Something like this:

```
man_pages = \
	defaults.conf.5 \
	nodes.conf.5 \
	warewulf.conf.5 \
	[...]
.PHONY: man_pages
man_pages: $(man_pages)
```

But (a) that's a long list (76 entries in that `[...]` from `wwctl`) and (b) I think it'd be more troublesome to maintain the Makefile that way.

I _think_ what you're primarily after is to just have `make` (and, by extension, `make all`) not do repeated builds; so what I've done in stead is create a `docs` target, moved `man_pages` to that (and out of `build) and also added `reference` to it. `install` now depends on `build` and `docs` individually.

Let me know what you think.

mslacken:
When testing this a two subsequent runs of `make proto` created following eroor:
```
internal/pkg/api/routes/wwapiv1/routes.pb.go:13:2: cannot find module providing package github.com/golang/protobuf/ptypes/empty: import lookup disabled by -mod=vendor
        (Go version in go.mod is at least 1.14 and vendor directory exists.)
internal/pkg/api/routes/wwapiv1/routes.pb.go:13:2: cannot find module providing package github.com/golang/protobuf/ptypes/empty: import lookup disabled by -mod=vendor
        (Go version in go.mod is at least 1.14 and vendor directory exists.)
internal/pkg/api/routes/wwapiv1/routes.pb.go:13:2: cannot find module providing package github.com/golang/protobuf/ptypes/empty: import lookup disabled by -mod=vendor
        (Go version in go.mod is at least 1.14 and vendor directory exists.)
internal/pkg/api/routes/wwapiv1/routes.pb.go:13:2: cannot find module providing package github.com/golang/protobuf/ptypes/empty: import lookup disabled by -mod=vendor
        (Go version in go.mod is at least 1.14 and vendor directory exists.)
GOOS=linux go build -mod vendor -tags "containers_image_openpgp containers_image_ostree" -o wwctl cmd/wwctl/main.go
internal/pkg/api/routes/wwapiv1/routes.pb.go:13:2: cannot find module providing package github.com/golang/protobuf/ptypes/empty: import lookup disabled by -mod=vendor
make: *** [Makefile:32: wwctl] Error 1
```
Is this introduced with this PR?

mslacken:
Can you also add the following, so that a change of `rules.proto`  recreates the go sources?
```
protcmd="PATH=$(TOOLS_BIN):$(PATH) $(PROTOC) -I /usr/include -I internal/pkg/api/routes/v1 -I=. \
		--grpc-gateway_out=. \
		--grpc-gateway_opt logtostderr=true \
		--go_out=. \
		--go-grpc_out=. \
		routes.proto"

.PHONY: proto
proto: $(PROTOC) $(PROTOC_GEN_GRPC_GATEWAY) $(PROTOC_GEN_GO) $(PROTOC_GEN_GO_GRPC)
	$(protcmd)

internal/pkg/api/routes/vwwapiv1/%.go: proto internal/pkg/api/routes/v1/routes.proto
	$(protcmd)
```

anderbubble:
@mslacken I'm not seeing the error you're getting. I'm seeing something similar:

```
[janderson@rocky fix-build]$ make cleanproto
rm -f internal/pkg/api/routes/wwapiv1/routes.pb.go internal/pkg/api/routes/wwapiv1/routes.pb.gw.go internal/pkg/api/routes/wwapiv1/routes_grpc.pb.go
[janderson@rocky fix-build]$ make proto
internal/pkg/api/container/container.go:12:2: no required module provides package github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1; to add it:
	go get github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1
internal/pkg/api/container/container.go:12:2: no required module provides package github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1; to add it:
	go get github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1
internal/app/api/wwapic/wwapic.go:14:2: no required module provides package github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1; to add it:
	go get github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1
internal/app/api/wwapird/wwapird.go:23:2: no required module provides package github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1; to add it:
	go get github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1
PATH=.tools/bin:/home/janderson/.local/bin:/home/janderson/bin:/usr/share/Modules/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin .tools/bin/protoc \
	-I /usr/include -I internal/pkg/api/routes/v1 -I=. \
	--grpc-gateway_opt logtostderr=true \
	--go_out=. \
	--go-grpc_out=. \
	--grpc-gateway_out=. \
	routes.proto
```

but this doesn't actually produce an error, and I think it's misleading. (I'm not an expert, but I think https://github.com/golang/go/issues/44961 is relevant.)

I also cleaned up a few things to improve the idempotency of the proto target. Not quite the implementation you suggested, but I think it addresses your request.

anderbubble:

PR #902, opened 2023-08-24
Update overlay edit with tempfile and mtime

anderbubble:
The recent `overlay edit` command reimplementation included some incorrect ordering that led to spurious error messages for files that did not yet exist. This reimplementation reorders operations to avoid such error messages and clarifies the implementation by editing in a temporary file and using mtime to discern whether an edit has occurred.


mslacken (changes requested):
Please remove th `\n")` artifacts.

anderbubble:
Those artifacts appear to have been erroneously retained in 12efa0cc6801131bbb0665458f3896b97638b664 when this was converted from a `fmt.Fprintf` construction.

Removing them now and updating the PR.

anderbubble:

PR #903, opened 2023-08-24
Fix warewulfd for new contextual overlay names

### LINKED ISSUE #896: new overlay filenames aren't working for autobuild
#883 improved overlay image naming, but this improvement isn't working for autobuilt overlays, which is still using the old format.

Ensure that all areas of the code are using the new format.

anderbubble:
The recent PR #883 replaced the standard overlay naming convention with static names for each of the runtime and wwinit/system contexts; but the method by which it left the "legacy" naming convention in place led to a bug where warewulfd was still attempting to serve overlays with the previous naming convention, though this name wasn't being built.

This PR refactors contextual overlay image naming to be, I think, a more typical golang pattern, and updates affected code to match.

This PR also included tests for all functions modified by this change.

Fixes #896

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,6 +79,7 @@ require (
 	github.com/prometheus/procfs v0.6.0 // indirect
 	github.com/rootless-containers/proto v0.1.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
+	github.com/sassoftware/go-rpmutils v0.2.0 // indirect
 	github.com/sirupsen/logrus v1.8.1 // indirect
 	github.com/spf13/pflag v1.0.6-0.20210604193023-d5e0c0615ace // indirect
 	github.com/stefanberger/go-pkcs11uri v0.0.0-20201008174630-78d3cae3a980 // indirect
--- a/go.sum
+++ b/go.sum
@@ -409,6 +409,7 @@ github.com/Azure/go-autorest/tracing v0.6.0/go.mod h1:+vhtPC754Xsa23ID7GlGsrdKBp
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
+github.com/DataDog/zstd v1.4.5/go.mod h1:1jcaCB/ufaK+sKp1NBhlGmpz41jOoPQ35bpF36t7BBo=
 github.com/Microsoft/go-winio v0.4.11/go.mod h1:VhR8bwka0BXejwEJY73c50VrPtXAaKcyvVC4A4RozmA=
 github.com/Microsoft/go-winio v0.4.14/go.mod h1:qXqCSQ3Xa7+6tgxaGTIe4Kpcdsi+P8jBhyzoq1bpyYA=
 github.com/Microsoft/go-winio v0.4.15-0.20190919025122-fc70bd9a86b5/go.mod h1:tTuCMEN+UleMWgg9dVx4Hu52b1bJo+59jBh3ajtinzw=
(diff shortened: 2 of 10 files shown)


anderbubble:
@griznog I'd be interested in your thoughts on this as well, mostly to check that you don't have some esoteric use case for `wwctl overlay build --overlay`. ^_^

griznog:
Of course I use it if it's esoteric :grin: 

The most important use is that because I push overlays manually and individually rather than use `wwclient` to pull the entire runtime bundle on a schedule, I need all the individual overlays to be built. I really think it'd  be better to build the merged bundle dynamically from individual overlay when the node requests it, as it's just a `cat` operation and doesn't require a lot CPU to accomplish. As a long term RFE I'd rather see  `wwclient` have access to a flag for each overlay to control how/when that overlay is pulled/applied post-boot and be able to be called manually to pull a single overlay, which would also have the useful side effect of eliminating the need for a system vs. runtime overlay distinction.

I also use `--output` quite a bit when testing overlays, although that may be a legacy thing as I picked up that habit before there was a render command, so that just may be a place where I need to change muscle memory. Still it's something that could be useful for debugging.

While on the topic of overlays, `wwctl overlay build` also doesn't build overlays based on the nodes config and profiles, it builds all overlays for all nodes even if they aren't needed/wanted. I consider that a low impact security hole as it gives nodes the opportunity to download individual overlays they should never see based on the node config. 

Related to that issue is that `wwctl` does not clean up after itself. If/when a node config changes to remove an overlay, the old copies of that in the `WWPROVSION` directory should also be cleaned up.

And all those things tie in with the fact that it's difficult to detect when an overlay has changed and needs to be rebuilt, so the entire concept of auto-building is flawed. It can certainly be triggered automatically when something known to impact overlays happens, but since `nodes.conf` and overlays can be hand edited and it's not obvious when a tag is added which (if any) overlays might use it without a lot of processing to figure that out, it'd make more sense to me to just require a `wwctl overlay build` manually if `wwctl overlay build` will remain a thing.

OR ... go all in with autobuilding and build overlays individually every time they are requested and don't cache them in `WWPROVISION` at all, removing `wwctl overlay build` as an option. This would have the nice side effect of implicitly parallelizing the building of overlays in the most efficient and simplest way possible. Debugging could be accomplished by pretending to be the node and requesting the specific overlay from `warewulfd`. 

All of that is probably better suited for a new issue around a complete re-think/re-factor of overlay handling, which includes pre/post hooks when an overlay is pulled/applied, revision control for overlays, etc.

griznog:
I put my current wrapper for building overlays here: https://gist.github.com/griznog/0405b1d864d9e74ddb642f56f137a4d1 as it encapsulates my crude attempt at versioning overlays. But it's still broken as to know for sure when to rebuild an overlay it'd need to check all tags used in all templates against any changes to those tags in `nodes.conf`. The more I ponder it the more I think caching built overlays in any way without leaving corner case bugs that cause them to be stale is a really hard problem and just having them always build on request to `warewulfd` solves that pretty elegantly.

mslacken:
So if its still in use, we could mark the flag as hidden?

anderbubble:
I like that idea, @mslacken. I did talk with @griznog about this a bit, and my intent was to re-work this PR to restore the functionality I had removed; but marking them hidden would still help reduce the confusion I constantly see about this feature on Slack.

@griznog would you object to making the ability to build specific named overlays hidden / undocumented?

griznog:
I'm generally not a fan of undocumented features so I could live with that but I think it's a bad idea just based on my aversion to undocumented features. 

Assuming that the rendering command/option can be used for all template debugging (e.g. I don't want to have to build the entire overlay stack each time I want to debug a template) then a better solution from my perspective would be to completely eliminate the `build` command from `wwctl` and instead have `warewulfd` build the individual or merged overlays whenever they are requested. Benefits to that are:

1.  No need to store them on-disk at all, overlays could be eliminated from the provisioning directory
2.  No need to worry about cleaning up the provisioning staging area when overlay configs/names change
3.  If I want to get actual overlay image file(s) for some reason, I can just pretend to be a given node and `wget` it from `localhost:WWPORT` on the master. That's a simple script to write locally.
4. `warewulfd` can be the enforcer/gateway which ensures an overlay never gets built for a node unless it is assigned to that nodes config. 
5. Always building on request eliminates the need for complex change detection for overlays or the node config.

Potential problems are:
1. Thundering herd issues with a cluster booting all at once and/or updating a nodes config at the same time. (Maybe `warewulfd` already has a mechanism for putting request on hold at some threshold?)
2. [Insert terrible awful thing I'm not thinking of here.]

griznog:
Again specific to my environment, it'd be nice if the recent idea I had about using a dummy node to define overlays that get used by the container when `chroot`'ing to it didn't break as a result of changes to `wwctl overlay build`. Slack thread here: https://hpcng.slack.com/archives/C019X4EUFFW/p1693754511135419. That trick has worked out great for getting templated files/directories bind mounted into the `chroot` when managing the container.

anderbubble:
@griznog I've re-written this PR to retain the ability to build and fetch individual overlays like you said you need. But I'd be very interested to know if it introduces any problems for you.

anderbubble:
@mslacken I'd be interested in your review of this updated PR. I'm particularly interested in your feedback on the new tests.

griznog:
@anderbubble  I tested this and it seems to work fine, looks good to me.

mslacken (approved):
I really like the big test coverage your have created.
I will create similar tests for the grub intergration.

anderbubble:
@mslacken I'm glad you like it. I feel like this technique for stubbing out the necessary directories has finally solved the remaining problems I've had writing successful tests for Warewulf, so I foresee good things for our ability to test Warewulf in the future!

Thanks for the merge!

PR #925, opened 2023-09-20
do not download protobuf stuff on offline builds

mslacken:
All the proto stuff will be downloaded so just disable all protobuff rules and targets on OFFLINE_BUILD.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,10 +11,8 @@ build: wwctl wwclient wwapid wwapic wwapird etc/defaults.conf etc/bash_completio
 docs: man_pages reference
 
 vendor:
-ifndef OFFLINE_BUILD
 	go mod tidy -v
 	go mod vendor
-endif
 
 config = etc/wwapic.conf \
 	etc/wwapid.conf \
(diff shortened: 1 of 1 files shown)


anderbubble:
Was this actually causing a problem? Since the built protobuf files are already included in the repo, my expectation was that this would only be an issue if one first ran `make cleanproto`.

To make sure I understand, this is just about the tools having to be downloaded, right?

There's a few small changes I'd make (myself) to this before merging, but I want to make sure I understand the overall goal and need first.

mslacken:
> Was this actually causing a problem? Since the built protobuf files are already included in the repo, my expectation was that this would only be an issue if one first ran `make cleanproto`.
> 
The generated files depend on on e.g. $(PROTOC) so make is building $(PROTOC) and so downloading the stuff depsite the fact that it never uses $(PROTOC)
> To make sure I understand, this is just about the tools having to be downloaded, right?
> 
Right, checkout and the first things it does is to download the protoc compiler
> There's a few small changes I'd make (myself) to this before merging, but I want to make sure I understand the overall goal and need first.

anderbubble:
It was still trying to pull down PROTOC; so I adjusted the strategy for OFFLINE_BUILD to simplify it. Now mock can build packages again.

PR #930, opened 2023-09-22
bash completion for all overlay commands

mslacken:
added the list command for the overlay files for all `wwctl overlay` commands. Preivous commits fell under the floor.


mslacken:
no needed any more

Issue #756, opened 2023-03-27
`wwctl configure -a` and `wwctl configure tftp` not creating `/var/lib/tftpboot/warewulf`


pedroalvesbatista:
### Version of Warewulf

What version of Warewulf are you using? Run

```
wwctl version
wwctl version:	 4.4.0-1.git_dd3b34a
rpc version: apiPrefix:"rc1" apiVersion:"1" warewulfVersion:"4.4.0-1.git_dd3b34a"
```

### Expected behavior

Running `wwctl configure -a` and `wwctl configure tftp` should create `/var/lib/tftpboot/warewulf`.

### Actual behavior

An error message appears after running `wwctl configure tftp` as after all configurations done, compute nodes doesn't boot due to not finding any PXE file to load in :  

```
Writing PXE files to: /var/lib/tftpboot/warewulf
ERROR  : Could not open source file /usr/local/share/x86_64.efi: open /usr/local/share/x86_64.efi: no such file or directory
WARN   : ipxe binary could not be copied, booting may not work: open /usr/local/share/x86_64.efi: no such file or directory
ERROR  : Could not open source file /usr/local/share/x86_64.kpxe: open /usr/local/share/x86_64.kpxe: no such file or directory
WARN   : ipxe binary could not be copied, booting may not work: open /usr/local/share/x86_64.kpxe: no such file or directory
ERROR  : Could not open source file /usr/local/share/arm64.efi: open /usr/local/share/arm64.efi: no such file or directory
WARN   : ipxe binary could not be copied, booting may not work: open /usr/local/share/arm64.efi: no such file or directory
Enabling and restarting the TFTP services
```

### Steps to reproduce this behavior

* Run `wwctl configure -a` and `wwctl configure tftp`

### What OS/distro are you running

```sh
$ cat /etc/os-release
NAME="Rocky Linux"
VERSION="8.6 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.6"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.6 (Green Obsidian)"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:rocky:rocky:8:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
ROCKY_SUPPORT_PRODUCT="Rocky Linux"
[...]

PR #931, opened 2023-09-26
seperate the path for the iPXE binaries

mslacken:
As well the `defaults.conf` as the ipxe binaires were installed
in the same location. This doen't make any problems if the binaries
are installed fromt he warewulf sources but makes it impossible 
to configure the path for system ipxe binaries


mslacken:
close #756

mslacken:
#784 needs this PR, so that the iPXE paths can be specified

PR #932, opened 2023-09-27
warn and don't show scary error on syncuids

mslacken:
As warewulf should be warm and welcoming, don't show a scary error if there is
uid/gid mismatch. Logically this is an error but scares the hell out of
inexperienced users.


anderbubble (approved):
Rebased. Looks good to me.

PR #935, opened 2023-09-29
added check if commas can be set

### LINKED ISSUE #769: Using special characters in profile and node tags
```
wwctl node|profile set --tag 
```
Doesn't support `,` and possibly other special characters, but by accident I discovered that if I edit `nodes.conf` directly and add tags with `,` in the values they work fine.  I faintly recall discussing this with @mslacken but have forgotten why it doesn't work for `wwctl` and also since that discussion have assumed that `,` couldn't be used at all in values. So creating this issue to hopefully have this fixed for `wwctl` down the road and also as a note for anyone searching for how to get `,` into a value so they don't assume as I did that it doesn't work at all.

griznog

mslacken:
close #769


anderbubble:
Rebased to main.

anderbubble (approved):

PR #937, opened 2023-10-05
node set needs to have a node range set

### LINKED ISSUE #502: `wwctl node set` wants to modify nodes even though no other arguments
Observed behavior:

```
[I am root!@frankie:(bare)::~]# wwctl node set 

*** WARNING: This command will modify all nodes! ***

? Are you sure you want to modify 55 nodes(s)? [y/N] █
```

Expected behavior:

An error or usage message.

mslacken:
`wwctl node set` needs now a node range as if a value should be set 
for all nodes still this node range can be specified via argument

close #502


mslacken:
branched wrongly

PR #940, opened 2023-10-05
don't update ipmi if there is no password

### LINKED ISSUE #638: 50-ipmi wwinit script hangs when ipmi password blank
In Warewulf 3.x, a blank ipmi password would be replaced with a random string in the configuration for each node. In WW4, it appears that a blank password is passed to the node as-is. The current 50-ipmi script flags the password as undefined, but continues trying to set up the ipmi user. At least on the nodes I've tested, the script will wait indefinitely with:

IPMI command failed: Invalid data field in request
Password for user 2: _

  I think this check isn't working: https://github.com/hpcng/warewulf/blob/d99717f28802a683aea218829284fbdf9fcfb629/overlays/wwinit/warewulf/init.d/50-ipmi#L62-L64

mslacken:
- Closes #638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   not have a device specified. #1154
 - Return non-zero exit code on overlay sub-commands #1423
 - Simplify passing of arguments to commands through `wwctl container exec`. #253
+- Don't update IPMI if password isn't set. #638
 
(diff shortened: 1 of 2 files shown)


anderbubble:
@mslacken I pushed an additional commit that... well, it re-writes 50-ipmi pretty much entirely.

I was reviewing your PR and thought it didn't quite handle #638; but, when I started trying to offer an addendum, I got into refactoring mode and this came out.

Let me know what you think.

anderbubble (approved):

PR #956, opened 2023-10-11
Update el9.rst

Yiltan:

Updated el9.rst

In the previous step we download the rocky-9 container. We should also use that container when we set the default.

### DIFF
--- a/userdocs/quickstart/el9.rst
+++ b/userdocs/quickstart/el9.rst
@@ -137,7 +137,7 @@ the ``default`` node profile:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


anderbubble:
Nice catch. Thanks.

anderbubble:
@Yiltan can you update this commit with `--signoff`? You should be able to do something like

```
git commit --amend --signoff
git push --force
```

Thanks again!

anderbubble (changes requested):
Commit needs to be signed off.

anderbubble:
Change has already been applied in a different PR.

PR #959, opened 2023-10-16
allow empty commits on doc workflow

mslacken:
Hopefully removes following error message after a merge
```
Cloning into '/home/runner/warewulf-web'...
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -68,5 +68,5 @@ jobs:
           cp -r userdocs/_build/html ~/warewulf-web/static/docs/${GITHUB_REF##*/}
           cd ~/warewulf-web
           git add static/docs/${GITHUB_REF##*/}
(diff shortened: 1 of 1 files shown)


anderbubble (commented):
@mslacken can you look at #978 as an alternative? The solution here would lead to unnecessary empty commits at warewulf-web. I _think_ the solution at #978 would avoid them (and also let us start publishing docs per-release).

brianclemens:
Yep, this bit of #978 is the better option, since it skips docs publishing entirely when unnecessary. Will have to be removed if we ever start dynamically generating docs from the code like Apptainer, but for now it's optimal.
```
    paths:
      - 'userdocs/**'
```

anderbubble:
Closing in favor of #978.

PR #960, opened 2023-10-16
propagate paths to overlays and use for wwclient

mslacken:
Paths were not propagated to overlays
WWClientdir set set in Variables.mk and is now also used in the 
template `/warewulf/init.d/80-wwclient.ww`


anderbubble (approved):
I updated the changelog, and fixed a typo that was in the original overlay file. Otherwise, lgtm!

PR #961, opened 2023-10-16
moved initialization of ConfigFile to New()

mslacken:
setting the ConfigFile (location of nodes.conf) in init()
led to the situation that a configured Paths.Sysconfdir is ignored.


anderbubble (approved):
Looks good to me!

PR #965, opened 2023-10-25
update include/exclude during dist creation

mslacken:
For the github actions this shouldn't matter, but calling `make dist` on
development systems creates nasty build artefacts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - added general test framework which creates temporary directories on the fly
 - only write IPMI if write is true
 - Don't show an error if image files for containers can't be found. #933
+- Clean dir before calling dist target
 
 ## [4.4.0] 2023-01-18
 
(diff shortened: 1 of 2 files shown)


anderbubble:
@mslacken if I'm following you correctly, I think the better way to handle this would be to restrict what's copied in by the rsync. Running `clean` would actually break the build, because we need, for example, the `vendor/` directory, which `clean` removes.

mslacken:
Yup, we can close this one and have a more extensive exclude list. `buildconfig.go` broke my neck this time

anderbubble:
I know what you mean. At one point I was accidentally nesting the dist, so each dist had the previous version within it. 😅

mslacken:
I am opening this again, not for implementation but I found follow bug.
The `--exclude .*` also excludes the `.ssh` dir in the generic template and is so disabling ssh logins. This also needs to be addressed.

anderbubble:
> The --exclude .* also excludes the .ssh dir in the generic template and is so disabling ssh logins.

I find that very surprising, as I don't think this rsync line has changed recently and ssh login has been working so far, hasn't it?

anderbubble:
@mslacken is the issue of excluding the `.ssh` directory no longer an issue? Or has it been reported elsewhere?

Issue #967, opened 2023-10-26
Tags set in profiles are ignored


griznog:
Pasting in from Slack:

@mslacken  not sure if this is worth creating an issue (edit: it is), hopefully you can spot what I'm doing wrong here. This is in development based on a fresh git pull a couple of hours ago.
I have a template that has this line in it:
```
{{ if .Tags.caddy_email -}}email {{ .Tags.caddy_email }}{{- end }}
```

Then I have set in a profile:
```
  tags:
    caddy_email: 'john.hanks@czbiohub.org'
```

When I then render that template I get no hit/match there and the line is empty for nodes which include this profile, for example:

```
[I am root!@frankie:(bare)::~]# wwctl node list -a svc-vm-globus | grep caddy_email | sed  's/[[:space:]]*$//g' 
  svc-vm-globus  Tags[caddy_email]                     tags-common                   john.hanks@czbiohub.org
```

Clearly the tag is defined via the tags-common profile, but when I render the template:

```
[I am root!@frankie:(bare):wwovl(service-caddy):caddy]# wwctl overlay show -r svc-vm-globus service-caddy /etc/caddy/Caddyfile.ww | grep john.hanks
backupFile: true
writeFile: true
Filename: /etc/caddy/Caddyfile
```

But if I define that tag in the node config, for example:

```[I am root!@frankie:(bare)::~]# wwctl node list -a svc-vm-webgw | grep caddy_email | sed  's/[[:space:]]*$//g' 
  svc-vm-webgw  Tags[caddy_email]                               --                            john.hanks@czbiohub.org
```

The node there svc-vm-webgw overrides the tag with a local value. This renders correctly:
```
[I am root!@frankie:(bare):wwovl(service-caddy):caddy]# wwctl overlay show -r svc-vm-webgw service-caddy /etc/caddy/Caddyfile.ww | grep john.hanks
backupFile: true
writeFile: true
Filename: /etc/caddy/Caddyfile
email john.hanks@czbiohub.org
```

Any idea why it's ignoring the tag if the tag is set via profile instead of directly in the node config? I'm thinking that when I updated to the latest development this morning I broke all my templates that are using a profile to set the tags.

anderbubble:
@mslacken this appears to have been introduced by https://github.com/hpcng/warewulf/pull/949

Can you investigate?

mslacken:
I can confirm this now, as the tags aren't correctly merged back.

griznog:
[...]

PR #971, opened 2023-10-26
added test for tag from profile

mslacken:
test for issue #967, but seems to work correctly.
Still an addtional test for the overlay tempalting has to be added.


mslacken:
@anderbubble The last commit does the fix, if you checkout and reset to HEAD~1 you will see one of the tests failing

anderbubble (approved):
I removed WriteFileAbs to preserve the safety of `testenv`; but otherwise lgtm.

PR #983, opened 2023-11-08
removed trailing space ww4-disks.target.ww

mslacken:
Makefile also didn't include soft links on packaging


mslacken:
Real nasty whitespace error

anderbubble (approved):
I split the commit to separate Makefile changes and the overlay rename, and I clarified the commit messages for each.

I removed the changelog entry because this shouldn't be new behavior since the last release: it just fixes a regression from development/main (as I understand it).

PR #1004, opened 2023-12-06
Use templates for power calls

mslacken:
Refactor the `wwctl power` calls in such way that `ipmitool` is not hardcoded,
but is deducted from the ipmitool.tmpl which is located in the $DATADIR
This empowers users to have other memthods for the power managemnt like
freeipmi or redfish without touching the warewulf code itself. As an
example a template for power management without bmc is included.

1. @anderbubble don't hit the merge button, because test coverage has
   increased a lot, but using the template is one of my best ideas in 
   the last time
2. @griznog we can use the template approach also for `wwctl container copy`
   so that on zfs pools the directory is cloned and on btrfs a `cp --reflink=auto`
   is used.


griznog:
This is an awesome approach! I am very much in love with this idea and would like to see as many `wwctl` commands as possible have the option to be calls to a template.

anderbubble (changes requested):
@mslacken if you'd still like to get this in, please rebase on current main.

mslacken:
@anderbubble looks more massive as it is. Most power command are copy and paste code. So you just have to check one carefully. 
Also I have added a lot of tests, so test coverage will go high with this PR.

anderbubble:
rebased to main

anderbubble:
Consolidated down some weirdness with CHANGELOG.

anderbubble:
Fixed Makefile.

anderbubble:
Rebased on main.

PR #1015, opened 2023-12-12
fix a bug when replacing existing containers will generate an error

### LINKED ISSUE #474: container import --force to replace an existing container deletes, generates an error
I import a container like this:

```
# wwctl container import docker://docker.io/anderbubble/openhpc-sms:latest
Setting VNFS name: openhpc-sms:latest
[...]
Updating the container's /etc/resolv.conf
WARN   : user: munge:997:994 not present on host
WARN   : user: slurm:202:202 not present on host
WARN   : uid(997) collision for host: munge and container: libstoragemgmt
ERROR  : Error in user sync, fix error and run 'syncuser' manually: user munge only present in container has same uid(997) as user libstoragemgmt on host,
add this user to /etc/passwd on host

# wwctl container ls
openhpc-sms:latest        0      4.18.0-348.23.1.el8_5.x86_64
```

Then, presuming updates at the registry, I try to import the container again:

```
# wwctl container import --force docker://docker.io/anderbubble/openhpc-sms:latest
Setting VNFS name: openhpc-sms:latest
Overwriting existing VNFS
Updating the container's /etc/resolv.conf
ERROR  : Could not create destination file /var/lib/warewulf/chroots/openhpc-sms:latest/rootfs/etc/resolv.conf: open /var/lib/warewulf/chroots/openhpc-sms:latest/rootfs/etc/resolv.conf: no such file or directory
WARN   : Could not copy /etc/resolv.conf into container: open /var/lib/warewulf/chroots/openhpc-sms:latest/rootfs/etc/resolv.conf: no such file or directory
ERROR  : Could not open /var/lib/warewulf/chroots/openhpc-sms:latest/rootfs/etc/passwd
ERROR  : Error in user sync, fix error and run 'syncuser' manually: open /var/lib/warewulf/chroots/openhpc-sms:latest/rootfs/etc/passwd: no such file or directory

# wwctl container ls

# ls -l /var/lib/warewulf/chroots/
total 0
```

The container is removed, but no new container is imported.

Re-running the import (with or without force) succeeds.

JasonYangShadow:

Fix a bug when replacing existing containers will generate an error

 - Fixes #474

## Test
Before:
image

After:
image

### DIFF
--- a/internal/pkg/api/container/container.go
+++ b/internal/pkg/api/container/container.go
@@ -175,21 +175,23 @@ func ContainerImport(cip *wwapiv1.ContainerImportParameter) (containerName strin
 	containerName = cip.Name
 	fullPath := container.SourceDir(cip.Name)
 
+	// container already exists and should be removed first
+	if util.IsDir(fullPath) && cip.Force {
+		wwlog.Info("Overwriting existing VNFS")
+		err = os.RemoveAll(fullPath)
+		if err != nil {
+			wwlog.ErrorExc(err, "")
+			return
+		}
+	}
+
 	if util.IsDir(fullPath) {
-		if cip.Force {
-			wwlog.Info("Overwriting existing VNFS")
-			err = os.RemoveAll(fullPath)
-			if err != nil {
-				wwlog.ErrorExc(err, "")
-				return
-			}
-		} else if cip.Update {
-			wwlog.Info("Updating existing VNFS")
-		} else {
+		if !cip.Update {
 			err = fmt.Errorf("VNFS Name exists, specify --force, --update, or choose a different name: %s", cip.Name)
 			wwlog.Error(err.Error())
 			return
 		}
+		wwlog.Info("Updating existing VNFS")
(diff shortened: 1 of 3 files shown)


internal/pkg/api/container/container.go
mslacken, line 1: Can you please explain, why this fixes this bug?

JasonYangShadow: @mslacken 
sorry that I misunderstood the `--update` flag. When container exists and `--force` flag is passed in, it'll remove existing container content, and then do a check again, so that another image pull will occur. Otherwise, `--update` won't trigger pull but only do syncuids. 

I updated the code, as it can receive both `--force` and `--update` flags at the same time. So I also added a warn message showing that when both flags are used, `--force` will remove existing container content and `--update` will be ignored. It'll rebuild forcely. 

Besides, I found that `--build` flag is actually not used in the current impl, I enabled it in this PR as well.

mslacken (approved):
lgtm

PR #1022, opened 2023-12-21
add container rename support

### LINKED ISSUE #583: Container copy and rename
In version 4.3 there is no way to copy a container and or rename it. I need this functionality so I can make a copy of my current container, rename the copy, and start experimenting with this renamed container. Therefore, I'm making a feature request.

Maybe something like

$ sudo wwctl container copy <container1> <container2>

This allows you to copy it and rename it at the same time. Without "container2" it just makes a copy with the name "container1-copy". If container2 exists, a message is written to stdout and the operation is cancelled.

To rename a container,

$ sudo wwctl container rename <container1_name> <container1_new_name>

There would likely need to be some limits on what you can name a container. Perhaps the command checks for that. If there is an illegal character, and error is written to stdout and the operation is cancelled.

Thanks!

Jeff

JasonYangShadow:

add container rename support

 - Fixes #583 


internal/app/wwctl/container/rename/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+	t.Run("Container list", func(t *testing.T) {
+		verifyContainerListOutput(t, "test-container-rename")
+	})
+}
+
+func verifyContainerListOutput(t *testing.T, content string) {
+	baseCmd := containerList.GetCommand()
+	baseCmd.SetOut(nil)
+	baseCmd.SetErr(nil)
+	stdoutR, stdoutW, _ := os.Pipe()
```
mslacken, line 47: You could simply use `wwlog.SetLogWriter(buf)` instead of creatinga pipe and as stream

JasonYangShadow: > You could simply use `wwlog.SetLogWriter(buf)` instead of creatinga pipe and as stream

not sure if I understand correctly. Here we need to retrieve the output of `container list`, it is generated by https://github.com/JasonYangShadow/warewulf/blob/issue/583/internal/app/wwctl/container/list/main.go#L37, PrintHelper will directly output to os.Stdout.

anderbubble:
Rebased onto main.

internal/app/wwctl/container/rename/main.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/hpcng/warewulf/internal/pkg/api/routes/wwapiv1"
+	"github.com/hpcng/warewulf/internal/pkg/container"
+	"github.com/hpcng/warewulf/internal/pkg/wwlog"
+	"github.com/spf13/cobra"
+)
+
+func CobraRunE(cmd *cobra.Command, args []string) (err error) {
+	if len(args) > 2 {
+		wwlog.Warn("rename only requires 2 arguments but you provided %d arguments. Hence, they will be ignored.", len(args))
+	}
```
anderbubble, line 16: This should quit out with an error if the incorrect number of arguments are provided.

internal/app/wwctl/container/rename/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"path"
+	"testing"
+
+	containerList "github.com/hpcng/warewulf/internal/app/wwctl/container/list"
+	"github.com/hpcng/warewulf/internal/pkg/testenv"
+	"github.com/hpcng/warewulf/internal/pkg/warewulfd"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Renname(t *testing.T) {
```
anderbubble, line 16: This is not important, but this is misspelled.

internal/app/wwctl/container/rename/root.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"github.com/hpcng/warewulf/internal/pkg/container"
+	"github.com/spf13/cobra"
+)
+
+var baseCmd = &cobra.Command{
+	DisableFlagsInUseLine: true,
+	Use:                   "rename CONTAINER NEW_NAME",
+	Aliases:               []string{"rn"},
```
anderbubble, line 11: The alias should be `mv` not `rn`.

internal/pkg/api/container/container.go
```diff
@@ -404,6 +404,69 @@ func ContainerShow(csp *wwapiv1.ContainerShowParameter) (response *wwapiv1.Conta
 	return
 }
 
+func ContainerRename(crp *wwapiv1.ContainerRenameParameter) (err error) {
```
anderbubble, line 407: I wish this wasn't being done as part of the api; but it does match the existing pattern, and I expect we hadn't talked about that yet when you did this.

Assuming this all works (I haven't actually tested it yet) I don't think we should refactor it right now; but this is one of those things I'll want to simplify in v4.6.0.

anderbubble (changes requested):
A few small changes; I'll try to do them myself.

internal/app/wwctl/container/rename/root.go
```diff
@@ -0,0 +1,35 @@
...
+		list, _ := container.ListSources()
+		return list, cobra.ShellCompDirectiveNoFileComp
+	},
+}
+
+var SetBuild bool
+
+func init() {
+	// Nothing to do here
+	baseCmd.PersistentFlags().BoolVarP(&SetBuild, "build", "b", false, "Build container when after pulling")
```
anderbubble, line 29: Should be "Build container after rename."

internal/pkg/api/container/container.go
```diff
@@ -404,6 +404,69 @@ func ContainerShow(csp *wwapiv1.ContainerShowParameter) (response *wwapiv1.Conta
...
+	// rename the container source folder
+	sourceDir := container.SourceDir(crp.ContainerName)
+	destDir := container.SourceDir(crp.TargetName)
+	err = os.Rename(sourceDir, destDir)
```
anderbubble, line 411: I think we should probably also either remove the built containers from the provision dir or move them, too.

anderbubble (changes requested):

anderbubble:
@JasonYangShadow I made a few changes that I want you to have a chance to respond to before merge. If they look good, I'll go ahead and merge; but I left them in separate commits so that they could be changed or removed if you disagree.

Let me know what you think. Thanks!

JasonYangShadow:
> @JasonYangShadow I made a few changes that I want you to have a chance to respond to before merge. If they look good, I'll go ahead and merge; but I left them in separate commits so that they could be changed or removed if you disagree.
> 
> Let me know what you think. Thanks!

Thanks for the changes, it looks totally okay to me.

anderbubble (approved):

PR #1030, opened 2023-12-27
Add string functions from sprig for use in templates

griznog:
- Making additional string functions available to templates.
- Updating CHANGELOG.md

fixing or enhancing.

 - Fixes #


griznog:
I think the only potential thing this breaks is making `replace` map directly to string.Replace rather than wrapping it and hard-coding the arg for number of replacements. I wanted to be able to use the function the same way it works normally and since `tr` was wrapping strings.Replace exactly the same way letting `replace` revert to the expected behavior seemed like the best approach.

anderbubble:
@griznog can you switch this to use https://github.com/Masterminds/sprig?

griznog:
@anderbubble adding the sprig package required me to `go get github.com/Masterminds/sprig/v3` to build, which is also killing the linter checks. Is there some place I should have added that so this would automatically happen?

Aside from that, this PR now adds the `sprig` functions for use in templates and is lightly tested :) 

griznog

henkela:
I'd like to add another function: extract the last character of a string in my case. But maybe extract n-characters (from the end) of a string more generally.

griznog:
> I'd like to add another function: extract the last character of a string in my case. But maybe extract n-characters (from the end) of a string more generally.

It's a bit convoluted, but it can be done with sprig functions. Given this tag for example:

```
# wwctl node set --tagadd=mystring=thisisalongstring node-kvm-1
```

The template:

```
{{ $mystring := .Tags.mystring }}
{{ $mystart := int (sub (len $mystring) 2) }}
MYSTRING={{ $mystring }}
LASTTWO={{ substr $mystart -1 $mystring }}
```

produces:

```
# wwctl overlay show -r node-kvm-1 debug /warewulf/substring.ww
backupFile: true
writeFile: true
Filename: /warewulf/substring


MYSTRING=thisisalongstring
LASTTWO=ng

```

anderbubble:
Converting to a draft until @griznog figured out an error he's getting.

```
/home/griznog/rpmbuild/BUILD/warewulf-4.5.x/internal/pkg/warewulfd/provision.go:243
template: /etc/warewulf/ipxe/default.ipxe: "/etc/warewulf/ipxe/default.ipxe" is an incomplete or empty template
```

griznog:
I'm pretty sure this is working and doesn't break anything. I'm not sure at all that I pulled off the rebase successfully.

internal/pkg/overlay/overlay.go
```diff
@@ -350,8 +352,12 @@ func RenderTemplateFile(fileName string, data TemplateStruct) (
...
 	backupFile = true
 	writeFile = true
-	tmpl, err := template.New(path.Base(fileName)).Option("missingkey=default").Funcs(template.FuncMap{
-		// TODO: Fix for missingkey=zero
+
+	// Build our FuncMap
+	funcMap := template.FuncMap{
+		"tr": func(source, old, new string) string {
+			return strings.Replace(source, old, new, -1)
+		},
```
mslacken, line 360: Although I like the `tr` on the command line, I would remove it from the templates and rely on the sprig functions only. This closes the doors for requests to add special functions because they are useful for this special use case.

griznog:
@mslacken the only occurrences of `tr` I found were in the ignition related templates, which I don't have a local test for other than to say there was no error when building the overlays. Please double-check that I got the arguments in the correct order when I switched these to `replace`.

mslacken (approved):

PR #1034, opened 2024-01-12
move dnsmasq template into etc

mslacken:
the template for dnsmasq was located direclty under `overlays/host` which 
created the `dnsmasq.d` dirctory directly und the root dir on the host. 
Moving the template to `etc` to that the config is created in the right dir.


anderbubble:
Rebased onto main.

anderbubble (approved):

PR #1037, opened 2024-01-15
removed double entries from paths

mslacken:
tftproot and datadir were in paths and in the tftp and warewulf 
sections. This is completely misleading and there is so no sane
way to configure warewulf.
Also it was possibleto set version and rc in the configuration 
which is also nuts.


mslacken:
@anderbubble As you should have a free time slot today, can you please merge and tag the rc1 release?

anderbubble:
@mslacken to split into multiple PRs.

mslacken:
@anderbubble only the path split now

anderbubble (approved):
lgtm. Thanks for splitting!

PR #1032, opened 2024-01-03
[WIP] Add output [yaml | json | csv | text] for list subcommand

JasonYangShadow:
## Description of the Pull Request (PR):

adding multiple output format support

this PR will add multiple output format support for wwctl list sub-commands.

example:

`container list`
```
root@ubuntu:/home/vagrant/warewulf# wwctl container list
  CONTAINER NAME  NODES  KERNEL VERSION                CREATION TIME        MODIFICATION TIME    SIZE     
  rocky8          0      4.18.0-513.9.1.el8_9.aarch64  02 Jan 24 07:23 UTC  01 Jan 70 00:00 UTC  1.2 GiB  
root@ubuntu:/home/vagrant/warewulf# wwctl container list -o yaml
containers:
    - createDate: "1704180191"
      kernelVersion: 4.18.0-513.9.1.el8_9.aarch64
      name: rocky8
      size: "1266908641"
root@ubuntu:/home/vagrant/warewulf# wwctl container list -o json
{"containers":[{"name":"rocky8","kernelVersion":"4.18.0-513.9.1.el8_9.aarch64","createDate":1704180191,"size":1266908641}]}
root@ubuntu:/home/vagrant/warewulf# wwctl container list -o csv
CONTAINER NAME,NODES,KERNEL VERSION,CREATION TIME,MODIFICATION TIME,SIZE
rocky8,0,4.18.0-513.9.1.el8_9.aarch64,02 Jan 24 07:23 UTC,01 Jan 70 00:00 UTC,1.2 GiB
root@ubuntu:/home/vagrant/warewulf# wwctl container list -o text
  CONTAINER NAME  NODES  KERNEL VERSION                CREATION TIME        MODIFICATION TIME    SIZE     
  rocky8          0      4.18.0-513.9.1.el8_9.aarch64  02 Jan 24 07:23 UTC  01 Jan 70 00:00 UTC  1.2 GiB  
```

`profile list`
```
root@ubuntu:/home/vagrant/warewulf# wwctl profile list
  PROFILE NAME  COMMENT/DESCRIPTION                                   
  default       This profile is automatically included for each node  
root@ubuntu:/home/vagrant/warewulf# wwctl profile list -o yaml
profiles:
    - profileSimple:
        comment: This profile is automatically included for each node
        profileName: default
root@ubuntu:/home/vagrant/warewulf# wwctl profile list -o json
[...]

PR #1050, opened 2024-01-22
simple yaml output for node list

mslacken:
simple output of a yaml for node list only


anderbubble:
Isn't this redundant with the work going on in #1032? Can you combine efforts there?

mslacken:
> Isn't this redundant with the work going on in #1032? Can you combine efforts there?

This Pr is just meant as a example on how to implement #1032

PR #1055, opened 2024-01-24
added {{ .Overlay }} to templates

### LINKED ISSUE #1052: Make current overlay name visible to templates
### Summary

Make the current overlay being rendered's name visible to templates, e.g. `{{ .Overlay }}` would resolve to `wwinit` when rendering things in the `wwinit` overlay.

### Rationale

It would be nice to have the current overlay name available in templates. For instance, if an overlay sets up a systemd service, then it might use the overlay name to build the ExecStart line:
```
[Service]
ExecStart=/warewulf/sbin/service.{{ .Overlay }}.sh
```
This is a step towards having a basic systemd service template that can be easily reused without needing to go and edit each file to update things that include the overlay name. 

### Description

`{{ .Overlay }}`  should resolve in templates to the current overlay name being rendered.

### Additional information

_No response_

### General information

mslacken:
{{ .Overlay }} contains the overlay name in which the template is located, otherwise
its values is "UNDEF" if the RenderTempalte is called outside the warewulf
overlay tree

* Closes #1052

### DIFF
--- a/internal/pkg/overlay/overlay.go
+++ b/internal/pkg/overlay/overlay.go
@@ -352,6 +352,18 @@ func RenderTemplateFile(fileName string, data TemplateStruct) (
 	err error) {
 	backupFile = true
 	writeFile = true
+	// parse the overlay name out of the absolute path
+	overlayPath, _ := filepath.Rel(OverlaySourceTopDir(), fileName)
+	withoutRootfs, err := filepath.Rel("rootfs", overlayPath)
+	if err != nil {
+		overlayPath = withoutRootfs
+	}
+	overlayPathElem := strings.Split(overlayPath, "/")
+	if len(overlayPathElem) > 1 {
+		data.Overlay = overlayPathElem[0]
+	} else {
+		data.Overlay = ""
+	}
 
 	// Build our FuncMap
 	funcMap := template.FuncMap{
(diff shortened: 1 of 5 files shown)


internal/pkg/overlay/overlay.go
```diff
@@ -352,6 +352,18 @@ func RenderTemplateFile(fileName string, data TemplateStruct) (
...
+	overlayPath, _ := filepath.Rel(OverlaySourceTopDir(), fileName)
+	withoutRootfs, err := filepath.Rel("rootfs", overlayPath)
+	if err != nil {
+		overlayPath = withoutRootfs
+	}
+	overlayPathElem := strings.Split(overlayPath, "/")
+	if len(overlayPathElem) > 1 {
+		data.Overlay = overlayPathElem[0]
+	} else {
+		data.Overlay = "UNDEF"
```
anderbubble, line 365: I think an empty string would be more appropriate here, rather than hard-coding a string value "UNDEF". Any objection?

anderbubble (changes requested):

internal/pkg/overlay/overlay.go
griznog: An empty string avoids the problem of an overlay named `UNDEF`. I can't think of a case where I'd want this to return anything other than an empty string if I called it outside the context of an overlay.

anderbubble (approved):

PR #1036, opened 2024-01-15
Fix error when editing node

JasonYangShadow:
## Description of the Pull Request (PR):

Fix showing error when editing node.

## This fixes or addresses the following GitHub issues:

 - Fixes #1024
 - Fixes #1057

## Before submitting a PR, make sure you have done the following:

- [ ] Signed off on all commits (e.g., using `git commit --signoff`) in agreement to the [DCO](DCO.txt)
- [ ] Read the [documentation for contributing to Warewulf](https://warewulf.org/docs/main/contributing/contributing.html)
- [ ] Added changes to the [CHANGELOG](https://github.com/hpcng/warewulf/blob/main/CHANGELOG.md) if necessary
- [ ] Updated [userdocs](https://github.com/hpcng/warewulf/tree/main/userdocs) if necessary
- [ ] Based this PR against the appropriate branch (typically [main](https://github.com/hpcng/warewulf/tree/main/userdocs))
- [ ] Added myself as a contributor to the [Contributors File](https://github.com/hpcng/warewulf/blob/main/CONTRIBUTORS.md)

PR #1064, opened 2024-01-30
check real path of `nodes.conf`

mslacken:
as init() isn't call any more on 8323cb8aa781d52ccca7c2b121ef5b91b0c0f0f9 the test of nodes.conf failed


JasonYangShadow (approved):
lgtm!

anderbubble:
I think this was already fixed in #1036. I'm closing for now; but @mslacken let me know if you think it's still an issue.

PR #1068, opened 2024-02-01
Container management functionality

### LINKED ISSUE #1067: Advanced container handling
### Summary

The interaction of `wwctl` with container registries is quite limited. Following things are missing:
* signature check
* cache cleanup 
* reimport from the same container source
* regsitry information of the imported container
* push of a container

### Rationale

_No response_

### Description

_No response_

### Additional information

_No response_

### General information

mslacken:
Add a more advanved container management functionality to warewulf.
close #1067


mslacken:
@anderbubble @griznog This PR will add a
```
wwctl container reimport
```
command which will recreate a container from its original source. I think this useful, when you have messed your container installation process and want a fresh start. What do you think?
I implemented this more by accident as the cache is already in warewulf, so a reimport just pulls the old references.

anderbubble:
I like the idea in theory; but how does it know what container to reimport from? What happens if the cache is cleared?

mslacken:
> I like the idea in theory; but how does it know what container to reimport from?
At import time additional information is written to a meta file, so that the metainformation of the container is preserved
> What happens if the cache is cleared?
An error! As a cache clearing mechanism isn't implemented, this is not really a problem.

My idea is, to also add the possibility to add new layers to cache, if a container is modified. This would make it possible to push/export containers back to a registry.

anderbubble:
In anticipation of the upcoming v4.6.0 release, I am reviewing open PRs that are not currently associated with a release milestone.

mslacken:
too much to rebase

PR #1076, opened 2024-02-08
Update metadata for 4.5.0rc1 release

mslacken:
Updated the metadata for the 4.5.0rc1 release. Its a quite big
patch as
* for the config files the confversion 45 must be set seperately
* tests get the version by the compiled in confversion

Also Version and Release are now const strings to make obivous
that this values can't be changed, as this might be suggested by
unexported variables.

@anderbubble must be rebased after #1037 is merged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [4.5.0] 2024-02-08
 
 ### Added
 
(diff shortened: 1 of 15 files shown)


anderbubble:
@mslacken I did the rebase, and made two changes. But I wanted to give you a chance to object before I did.

Rebase with changes is at https://github.com/anderbubble/warewulf/tree/45Release

* Changed version to 4.5.0rc1 (v4.5.0-rc1 not allowed in RPM)
* Reverted tests to use static version string

I think it's important to keep tests as static as possible. If the version changes, we _should_ have to update the tests. If we pull values from the code into the tests, then we risk not catching accidental changes in behavior.

At least, that's my opinion. But if you object, I'll redo the rebase.

mslacken:
The more important thing is that we have the version string inside the code now.
Having it static in tests sounds reasonable, so go ahead and merge.

anderbubble (approved):

anderbubble:
Great! Thanks for the consultation.

PR #1089, opened 2024-02-17
Fix sphinx warnings

anderbubble:


mslacken:
@anderbubble I would have recommended to leave the el7 in the documentation. So imagine that you are an junior sysadmin and so you are only allowed to manage the old cluster which has only el7 installed. In such a case you would still want to have the old documentation in place.

anderbubble:
@mslacken fair enough. We had previously decided (though perhaps a bit informally) to stop supporting el7 after EOL; but it's early for that. I was just trying to minimize the amount of work needed, and the el7 quickstart needs to at least be updated.

I'll plan to restore it prior to release.

mslacken:
I would suggest not to remove the quickstart, but mark it deprecated with
* the deprecation date 
* deprecation version 
in the documentation so that the user knows, that it has worked, but doesn't get updated any more.

PR #1091, opened 2024-02-19
Prevent Networkmanager from trying to obtain IP address via DHCP

riederd:
Prevent Networkmanager from trying to obtain IP address via DHCP on unused/unmanaged network interfaces. Only act on interfaces. configured via wwctl otherwise systemd NetworkManager-wait-online.service. runs into a time out.


anderbubble:
@riederd thanks for the contribution! I need to review in a bit more detail, cause right now I'm not following how this gets interfaces that Warewulf doesn't know about or, if you do have to enumerate them, how it differentiates from configured interfaces.

In the mean time, though, could you update the commit with a DCO signoff?

riederd:
> @riederd thanks for the contribution! I need to review in a bit more detail, cause right now I'm not following how this gets interfaces that Warewulf doesn't know about or, if you do have to enumerate them, how it differentiates from configured interfaces.

The key is the `except:` prefix, this tells the Networkmanager that all devices which are not listed are unmanaged.
e.g.

```
unmanaged-devices=except:interface-name:ceph-client,except:interface-name:ens8f0np0,except:interface-name:ens8f1np1,except:interface-name:ens1f0
```


> In the mean time, though, could you update the commit with a DCO signoff?

Done

PR #1107, opened 2024-02-26
removed static if clause for dhcpd.conf.ww

mslacken:
`static` isn't a keyword in `warewulf.conf` any more, so removed the use in the template


anderbubble:
Static isn't a keyword; it's a value for "template"; and you can't just remove the if statements; if you want to remove it, you have to remove the contents of the clause, too.

But since people are actively using the static template, I don't think this should be removed. What is the goal here?

mslacken:
Right, I thought we just use a boolean to distinguish between static and non static environments. So my mistake.

Issue #1111, opened 2024-02-27
Nodes that boot outside iPXE may not have `/proc/cmdline` set


MiddelkoopT:
### Steps to reproduce

Boot a node outside iPXE and wwid will not get set causing wwclient to abort.  Wwclient gets wwid from /proc/cmdline set as wwid=mac:address.  In case of the Raspberry pi a solution would require modifying `/proc/cmdline` which is readonly (a solution would be to bind-remount a file to /proc/cmdline) or creating a directory/file on the tftpserver for each node.  This would require a custom script as templating does not support creating arbitrary files from a single template.

### Error message

```
Aug 22 00:02:43 c0 wwclient[695]: ERROR  : 'wwid' is not defined in /proc/cmdline
```

### Information on your system

Raspberry Pi booting via custom tftboot.

```
[root@head warewulf]# wwctl version
wwctl version:	 4.5.x-1.git_c036f658
rpc version: apiPrefix:"rc1" apiVersion:"1" warewulfVersion:"4.5.x-1.git_c036f658"
```

### General information

- [X] I have run `wwctl version` and reported the contents of `/etc/os-release`
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have captured and reported relevant error messages and logs

PR #1109, opened 2024-02-26
Add support for machines that do not have a SMBIOS

MiddelkoopT:
## Description of the Pull Request (PR):

Allow machines without a SMBIOS to run wwclieint.  The Raspberry Pi is one such machine.  The PR will try to read from "/sys/firmware/devicetree/base/serial-number" if there is no SMBIOS (Raspberry Pi Serial Number) and create a UUID from that.

## This fixes or addresses the following GitHub issues:

 - Fixes warewulf/warewulf#1108

## Before submitting a PR, make sure you have done the following:

- [x] Signed off on all commits (e.g., using `git commit --signoff`) in agreement to the [DCO](DCO.txt)
- [x] Read the [documentation for contributing to Warewulf](https://warewulf.org/docs/main/contributing/contributing.html)
- [x] Added changes to the [CHANGELOG](https://github.com/warewulf/warewulf/blob/main/CHANGELOG.md) if necessary
- [ ] Updated [userdocs](https://github.com/warewulf/warewulf/tree/main/userdocs) if necessary
- [ ] Based this PR against the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))
- [x] Added myself as a contributor to the [Contributors File](https://github.com/warewulf/warewulf/blob/main/CONTRIBUTORS.md)

PR #1112, opened 2024-02-27
Get wwid from interface not /proc/cmdline when wwid=[interface] is set

MiddelkoopT:

Add code to detect the new format of `wwid=[interface]`, where interface is the interfaceid for which the wwid mac address is taken from, and replace it with `wwid=mac:addres` where `mac:address` is the mac address of the interface..  The code will read the mac address form this interface and use it as wwid in the remainder of the code.

This code also fixes a bug where if the wwid is the last argument in `/proc/cmdline` and `/proc/cmdline` has a newline at the end, the newline will show up in wwid ("mac:address\n") (first line of the patch)

Contributors addition is in PR #1109

 - Fixes Issue #1111

- [X] Signed off on all commits (e.g., using `git commit --signoff`) in agreement to the [DCO](DCO.txt)
- [X] Read the [documentation for contributing to Warewulf](https://warewulf.org/docs/main/contributing/contributing.html)
- [X] Based this PR against the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))


mslacken (changes requested):
One sensible reason for `wwclient` to rely on `wwid` to be set via `/proc/cmdline` is, that this way you can be sure, that `wwclient` only modifies files, if the node was booted via warewulf. I don't know if this behavior was intentional or by accident.  [*]
For that reason can you please explain, what is the use case, that `/proc/cmdline` can't be read, but `wwclient` should run?
Can you also add a Changelog entry?

MiddelkoopT:
The wwid=[eth0] still needs to be present to boot, it just replaces the [eth0] with the mac address.  In this way traditional booting will only modify files if booted via warewulf.  

This patch is designed to get a Raspberry Pi to work with warewulf.  The Raspberry pi boots via it's own custom built-in bootloader (half in EEPROM and half loaded via tftp for the Pi 4) and loading of the config/kernel/and initrd for booting is done tftp and does not use iPXE to load the kernel.  The kernel and the kernel command line is loaded via tftp from a separate file (cmdline.txt) in a folder that is either the serial number (default) or the mac address if different cmdline.txt's are required per node.  This means that every node has to have a separate directory with the cmdline.txt in it (with their own mac addresses) and possibly need to know the serial number.  This requires additional overhead to create these folders for each node.  From what I understand warewulf templating cannot create files in a single tempalte.

Given the small amount of RAM on each node (2-8GB), and a 62MB limit on initrd size, I have developed an initrd that writes the container and overlays to disk in Dracut and Dracut will pivot root to the newly mounted filesystem.  This method with this patch only requires a single cmdline.txt for all nodes.  (This method also could be adapted to use a ram-disk if desired.) One workaround is to bind-mount a file on top of /proc/cmdline (untested - I was worried about pivot root for this method) with the wwid in it.  Another workaround would be to use a u-boot/UEFI/iPXE boot path but it's fragile, not well supported, and unlikely to support the Raspberry Pi 5 soon. It also has some initrd size limits (the limits are unknown at this time, I did not explore this path too far due to the complexity of running 4 bootloaders).

Changelog incoming.

mslacken (approved):
Thanks @MiddelkoopT  great work

PR #1119, opened 2024-03-05
Fixed profile list bug

yqin:
Fixed a bug with profile list when kernelargs are set with 'var=val' format. This also removed the unnecessary '--' for the source profile since it's never used.

If kernelargs are set, `profile list -a` format is messed up due to profile uses '=' as separator, e.g.,

```
# wwctl.orig profile list -a default
  PROFILE  FIELD                        PROFILE  VALUE
  default  Id                           --       default
  default  Comment                      --       This profile is automatically included for each node
  default  Kernel.Args                  --       crashkernel                                           no vga  791 net.naming-scheme  v238
```

This fix standardized the field separator to ':=:' which is what node uses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Systems with no SMBIOS (Raspberry Pi) will create a UUID from
   `/sys/firmware/devicetree/base/serial-number`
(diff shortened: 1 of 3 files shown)


anderbubble:
Thanks for the PR, @yqin!

I had thought this was already fixed; but we'll review and look at getting this into v4.5.1.

mslacken (approved):
Thanks @yqin

PR #1182, opened 2024-04-10
Fix overlay directory location/creation

### LINKED ISSUE #830: container modification lock during shell/exec
### Version of Warewulf

What version of Warewulf are you using? 
```
```

### Expected behavior

"wwctl container shell containername" should block if containername already in use

### Actual behavior

Containers can be open multiple times simultaneously

```
NAME="Rocky Linux"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
```

RPMs built from source

### Additional Notes:
I accidentally had the same container open in two different shells, and may have made conflicting minor edits. This could happen non-accidentally with multiple administrators. Can "wwctl container shell" set and check for a per-container lock file at start? Lock removal should happen automatically on exit, but there should be an option to override it.

### LINKED ISSUE #1180: `/var/lib/warewulf/chroots/overlay` directory should be somewhere else.
### Steps to reproduce

```
[root@micropterus chroots]# ls
rocky-9  rocky-9-minimal
[root@micropterus ~]# wwctl container shell rocky-9
[rocky-9] Warewulf> exit
exit
+ LANG=C
+ LC_CTYPE=C
+ export LANG LC_CTYPE
+ dnf clean all
0 files removed
Rebuilding container...
Created image for VNFS container rocky-9: /var/lib/warewulf/provision/container/rocky-9.img
Compressed image for VNFS container rocky-9: /var/lib/warewulf/provision/container/rocky-9.img.gz
[root@micropterus chroots]# ls 
overlays  rocky-9  rocky-9-minimal
[root@micropterus chroots]# tree overlays/
overlays/
├── lower
├── nodeoverlay
└── work

3 directories, 0 files
```

Not sure what this `overlay` directory does or what triggers it to be created, but it breaks the assumption that everything in `/var/lib/warewulf/chroots` will be a node image `chroot`.

mslacken:
- Close #1180
- Close #830
- using the right overlay directory and remove it
- use warewulf.conf from parent on child

The directories for the overlays needed for the bind mount are now created under `conf.Paths.WWChrootdir/$CONTAINERNAME/run`. This directory can be used as lock, so that there can't be congruent shell/exec calls to the same container.

### DIFF
--- a/internal/app/wwctl/container/exec/child/main.go
+++ b/internal/app/wwctl/container/exec/child/main.go
@@ -21,7 +21,7 @@ import (
 	"github.com/warewulf/warewulf/internal/pkg/wwlog"
 )
 
-const exitEval = `$(VALU="$?" ; if [ $VALU == 0 ]; then echo write; else echo discard; fi)`
+const exitEval = `$(VALU="$?" ; if [ $VALU == 0 ]; then echo create new image; else echo no new image; fi)`
 
 func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	if os.Getpid() != 1 {
--- a/internal/app/wwctl/container/exec/child/main.go
+++ b/internal/app/wwctl/container/exec/child/main.go
@@ -21,8 +21,6 @@ import (
 	"github.com/warewulf/warewulf/internal/pkg/wwlog"
 )
 
-const exitEval = `$(VALU="$?" ; if [ $VALU == 0 ]; then echo create new image; else echo no new image; fi)`
-
 func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	if os.Getpid() != 1 {
 		wwlog.Error("PID is not 1: %d", os.Getpid())
@@ -36,8 +34,9 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 		os.Exit(1)
 	}
 	conf := warewulfconf.Get()
-	if overlayDir == "" {
-		overlayDir = path.Join(conf.Paths.WWChrootdir, "overlays")
+	runDir := container.RunDir(containerName)
+	if _, err := os.Stat(runDir); os.IsNotExist(err) {
+		return errors.Wrap(err, "container run directory does not exist")
 	}
 	mountPts := conf.MountsContainer
 	mountPts = append(container.InitMountPnts(binds), mountPts...)
@@ -45,11 +44,16 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	lowerObjects := checkMountPoints(containerName, mountPts)
 	// need to create a overlay, where the lower layer contains
 	// the missing mount points
-	wwlog.Verbose("for ephermal mount use tempdir %s", overlayDir)
-	// ignore errors as we are doomed if a tmp dir couldn't be written
-	_ = os.MkdirAll(path.Join(overlayDir, "work"), os.ModePerm)
-	_ = os.MkdirAll(path.Join(overlayDir, "lower"), os.ModePerm)
-	_ = os.MkdirAll(path.Join(overlayDir, "nodeoverlay"), os.ModePerm)
+	wwlog.Verbose("for ephermal mount use tempdir %s", runDir)
+	if err = os.Mkdir(path.Join(runDir, "work"), os.ModePerm); err != nil {
+		return err
+	}
+	if err = os.Mkdir(path.Join(runDir, "lower"), os.ModePerm); err != nil {
+		return err
+	}
+	if err = os.Mkdir(path.Join(runDir, "nodeoverlay"), os.ModePerm); err != nil {
+		return err
+	}
 	// handle all lower object, have some extra logic if the object is a file
 	for _, obj := range lowerObjects {
 		newFile := ""
@@ -57,12 +61,12 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 			newFile = filepath.Base(obj)
 			obj = filepath.Dir(obj)
 		}
-		err = os.MkdirAll(filepath.Join(overlayDir, "lower", obj), os.ModePerm)
+		err = os.Mkdir(filepath.Join(runDir, "lower", obj), os.ModePerm)
 		if err != nil {
 			wwlog.Warn("couldn't create directory for mounts: %s", err)
 		}
 		if newFile != "" {
-			desc, err := os.Create(filepath.Join(overlayDir, "lower", obj, newFile))
+			desc, err := os.Create(filepath.Join(runDir, "lower", obj, newFile))
 			if err != nil {
 				wwlog.Warn("couldn't create directory for mounts: %s", err)
 			}
@@ -75,10 +79,10 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	if err != nil {
 		return errors.Wrap(err, "failed to mount")
 	}
-	ps1Str := fmt.Sprintf("[%s|%s] Warewulf> ", exitEval, containerName)
+	ps1Str := fmt.Sprintf("[%s] Warewulf> ", containerName)
 	if len(lowerObjects) != 0 && nodename == "" {
 		options := fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s",
-			path.Join(overlayDir, "lower"), containerPath, path.Join(overlayDir, "work"))
+			path.Join(runDir, "lower"), containerPath, path.Join(runDir, "work"))
 		wwlog.Debug("overlay options: %s", options)
 		err = syscall.Mount("overlay", containerPath, "overlay", 0, options)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


anderbubble:
Rebased on main and updated changelog.

internal/app/wwctl/container/exec/child/main.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/warewulf/warewulf/internal/pkg/wwlog"
 )
 
+const exitEval = `$(VALU="$?" ; if [ $VALU == 0 ]; then echo write; else echo discard; fi)`
```
anderbubble, line 24: "write" and "discard" are a bit misleading here, aren't they? Changes stay resident in the chroot; they just aren't built into a new image.

Maybe "build" and "skip"? Then again, if I understand correctly, I think this change is out-of-scope for this PR, and should be split into a different PR, probably reserved for v4.6.0.

internal/app/wwctl/container/exec/root.go
```diff
@@ -25,17 +25,17 @@ var (
...
+	overlayDir string
+	nodeName   string
 )
 
 func init() {
 	baseCmd.AddCommand(child.GetCommand())
 	baseCmd.PersistentFlags().StringArrayVarP(&binds, "bind", "b", []string{}, "Bind a local path into the container (must exist)")
 	baseCmd.PersistentFlags().BoolVar(&SyncUser, "syncuser", false, "Synchronize UIDs/GIDs from host to container")
-	baseCmd.PersistentFlags().StringVar(&tempDir, "tempdir", "", "Use tempdir for constructing the overlay fs (only used if mount points don't exist in container)")
+	baseCmd.PersistentFlags().StringVar(&overlayDir, "overlaydir", "", "Use tempdir for constructing the overlay fs (only used if mount points don't exist in container)")
```
anderbubble, line 38: I'm concerned that the use of the term "overlay" here is confusing, given that we already use the term for a primary Warewulf feature. It took me a while to remember what's going on here myself.

internal/app/wwctl/container/exec/child/main.go
```diff
@@ -34,18 +36,21 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 		os.Exit(1)
 	}
 	conf := warewulfconf.Get()
+	if overlayDir == "" {
+		overlayDir = path.Join(conf.Paths.WWChrootdir, "overlays")
```
anderbubble, line 40: Isn't this the original complaint? My understanding of #1180 was that there shouldn't be anything created in `conf.Paths.WWChrootdir` that isn't a chroot.

internal/app/wwctl/container/exec/child/main.go
```diff
@@ -65,11 +70,12 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 		}
 	}
 	containerPath := container.RootFsDir(containerName)
+	// running in a private PID space, so also make / private, so that nothing gets out from here
 	err = syscall.Mount("", "/", "", syscall.MS_PRIVATE|syscall.MS_REC, "")
 	if err != nil {
 		return errors.Wrap(err, "failed to mount")
 	}
-	ps1Str := fmt.Sprintf("[%s] Warewulf> ", containerName)
+	ps1Str := fmt.Sprintf("[%s|%s] Warewulf> ", exitEval, containerName)
```
anderbubble, line 78: How does this change relate to fixing #1180?

internal/app/wwctl/container/exec/main.go
```diff
@@ -20,23 +24,22 @@ import (
...
-			if err != nil {
-				wwlog.Warn("Couldn't remove temp dir for ephermal mounts:", err)
-			}
-		}()
+func runContainedCmd(args []string) (err error) {
+	conf := warewulfconf.Get()
+	if matches, _ := filepath.Glob(path.Join(conf.Paths.WWChrootdir, args[0], args[0]) + "-run-*"); len(matches) > 0 {
+		return fmt.Errorf("found lock directories for container: %v", matches)
+	}
+	overlayDir, err = os.MkdirTemp(path.Join(conf.Paths.WWChrootdir, args[0]), args[0]+"-run-")
```
anderbubble, line 32: This isn't a true lock: there's a race condition between the checking of the existence of a directory for another process and the creation of one for this process.

If you want this to be a lock, don't use `os.MkdirTemp`--you only need that to prevent collisions with other processes. Since you intentionally want to create and detect collisions, just use `os.Mkdir` and use the error handling as the lock detection.

internal/app/wwctl/container/exec/child/main.go
anderbubble: Or have I misunderstood? Is it actually the case that changes are discarded on nonzero exit now?

_goes to test_

anderbubble: No, the files stay resident.

```
[root@wwctl2 ~]# wwctl container shell rocky-8
[rocky-8] Warewulf> touch /testfile
[rocky-8] Warewulf> exit -1
exit
Command exited non-zero, not rebuilding/updating VNFS image

[root@wwctl2 ~]# stat /var/lib/warewulf/chroots/rocky-8/rootfs/testfile
  File: /var/lib/warewulf/chroots/rocky-8/rootfs/testfile
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: fd00h/64768d	Inode: 51694332    Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Context: unconfined_u:object_r:var_lib_t:s0
Access: 2024-04-20 01:41:46.232917500 -0400
Modify: 2024-04-20 01:41:46.232917500 -0400
Change: 2024-04-20 01:41:46.232917500 -0400
 Birth: 2024-04-20 01:41:46.232917500 -0400
```

anderbubble (changes requested):

internal/app/wwctl/container/exec/root.go
mslacken: Right, as command line option we can find another name,

anderbubble:
After initial testing I confirm that the `overlays` directory is no longer created in `chroots` and the equivalent directory under the specific chroot is cleaned up on exit.

I still think we need to find another name for the "overlay" dir, especially since this introduces `wwctl container exec --overlaydir`. We can't overload the term "overlay" like that in the user interface. (I'd prefer to avoid overloading it in the code, too.)

internal/app/wwctl/container/exec/child/main.go
```diff
@@ -34,18 +36,21 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
...
 	// check for valid mount points
 	lowerObjects := checkMountPoints(containerName, mountPts)
-	overlayDir := conf.Paths.WWChrootdir + "/overlays"
 	// need to create a overlay, where the lower layer contains
 	// the missing mount points
 	wwlog.Verbose("for ephermal mount use tempdir %s", overlayDir)
 	// ignore errors as we are doomed if a tmp dir couldn't be written
 	_ = os.MkdirAll(path.Join(overlayDir, "work"), os.ModePerm)
 	_ = os.MkdirAll(path.Join(overlayDir, "lower"), os.ModePerm)
 	_ = os.MkdirAll(path.Join(overlayDir, "nodeoverlay"), os.ModePerm)
```
anderbubble, line 52: If we're doomed, maybe we shouldn't ignore the errors.

anderbubble:
@mslacken I made some adjustments and pushed to https://github.com/anderbubble/warewulf/tree/CleanupAfterShell. Let me know what you think.

mslacken:
@anderbubble can you make a PR from  https://github.com/anderbubble/warewulf/tree/CleanupAfterShell
I am closing this one.

anderbubble:
I'll just push to here if that's alright. I just wanted to give you a chance to see it before I made such a change to your branch.

mslacken:
So 👍 for a merge

internal/app/wwctl/container/exec/child/main.go
```diff
@@ -165,11 +172,8 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	os.Setenv("PATH", "/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin")
 	os.Setenv("HISTFILE", "/dev/null")
 
-	_ = syscall.Exec(args[1], args[1:], os.Environ())
-	/*
-		Exec replaces the actual program, so nothing to do here afterwards
-	*/
-	return nil
+	wwlog.Debug("Exec: %s %s", args[1], args[1:])
+	return syscall.Exec(args[1], args[1:], os.Environ())
```
anderbubble, line 176: @mslacken I remembered you made this change in the dracut branch but it didn't get merged there; so I brought it in here.

anderbubble (approved):
I think it looks good now. I'll merge in a bit if there are no objections by then.

anderbubble:
Actually, now I remember that I wanted to improve the error messages in my change a bit; so I'll do that before merge.

PR #1186, opened 2024-04-12
initial support for two stage initramfs built by dracut

### LINKED ISSUE #1115: Add a bootstrap stage to allow a two stage node boot process
### Summary

Request is for `warewulfd` to serve a 

* stage=bootstrap_container which would download a container to be specified by `BootStrapContainer` in the node config
* stage=bootstrap_kernel which would download a kernel to be specified by `BootStrapKernel` in the node config
* stage=bootstrap_system which would download the system overlay for the `BootStrapSystemOverlay`
* stage=bootstrap_runtime which would download the runtime overlay for the `BootStrapRuntimeOverlay` 

This would allow optionally booting nodes with an arbitrary kernel and container, which could then perform whatever actions are needed before downloading the final kernel/container/overlays and pivot/switch-root-ing  and/or doing a `kexec()`. 

### Rationale

This enhancement would allow easily testing using a shim bootstrap, whether that is a dracut initrd, custom bootstrap image or just a minimal OS container.  Potential applications/benefits are:

* Making large containers possible with Legacy BIOS by loading them from a shim minimal OS
* Allowing for a full OS environment to set up disks, `grub` and any other steps required to statefully provision nodes
* Allow stateless provisioning to be direct to disk rather than using a `tmpfs` and depending on swapping to free memory

### Description

* Add methods to set the values of `BootStrap [Container|Kernel|SystemOverlay|RuntimeOverlay]` in the node config
* Add necessary `stage_*` requests to `warewulfd`

The BootStrap containers, kernels and overlays can be managed just like any other containers, kernels and overlays. 

### Additional information

Not sure how this would look for `ipxe` or `grub` or any other boot methods, so leaving those as an exercise for the admin after the bootstrrap functionality is available. 

### General information

JasonYangShadow:

Initialized code base for the two stage initramfs dracut build support

 - Fixes #1115 

- [ ] Based this PR against the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))

## Test

1. Added a new option `--initramfs` for `container build` sub-command, which will use dracut inside container to build an initramfs image
image
2. Compute node now can make a call to `http://HEAD_NODE/provision/${mac}?stage=initramfs` to download initramfs image built in step 1
image
Verified logs in `/var/log/warewulfd.log`
image


mslacken:
We will have to discuss this at the next meeting. I would prefer another solution and leverage `util.‎BuildFsImage` instead. The advantage would be, that all the modules which are already in the kernel image.

internal/app/wwctl/container/build/main.go
```diff
@@ -12,6 +18,51 @@ func CobraRunE(cmd *cobra.Command, args []string) error {
...
+		rootfsDir := pkgcontianer.RootFsDir(c)
+		kver, err = kernel.FindKernelVersion(rootfsDir)
+		if err != nil {
+			return fmt.Errorf("failed to locate container kernel version: %s", err)
+		}
+
+		err = cexec.CobraRunE(cmd, []string{c, "/usr/bin/dracut --no-hostonly --force --verbose --kver " + kver})
+		if err != nil {
+			return
+		}
```
anderbubble, line 60: I think that this is running dracut on the host, right? If we go down the dracut path, we have to run dracut within the container.

JasonYangShadow: Nope, it should run inside the container. This utilizes the 'container exec' sub-command. Check this file 'internal/app/wwctl/container/exec/child/main.go'

internal/pkg/container/config.go
```diff
@@ -27,3 +27,9 @@ func ImageParentDir() string {
 func ImageFile(name string) string {
 	return path.Join(ImageParentDir(), name+".img")
 }
+
+// InitramfsBootPath returns the dracut built initramfs path, as dracut built initramfs inside container
+// the function returns host path of the built file
+func InitramfsBootPath(image, kver string) string {
+	return path.Join(RootFsDir(image), "boot", "initramfs-"+kver+".img")
```
anderbubble, line 34: Can you make sure that Suse and Debian both use this same path and naming scheme for dracut images? If not, we should make sure we can detect those, too.

anderbubble, line 34: I'm not certain how I feel about explicitly only detecting an initramfs that matches the kver to be booted. I want to see if we can overlay a kernel override onto the dracut image and boot that successfully. If so, then we should be able to boot any arbitrary initramfs and not worry about the kernel matching; but if not then it probably makes sense to be strict about it as you have here.

JasonYangShadow:
I will close this PR because of in favor of this PR
https://github.com/warewulf/warewulf/pull/1191

PR #1187, opened 2024-04-12
allow the user to specify ssh key types

### LINKED ISSUE #1185: Better FIPS Support in SSH key selection
### Summary

We are trying to use Warewulf 4 on a FIPS enabled system. When we run a wwctl configure --all there is a failure due to DSA being disabled in FIPS.

Can this be changed to a soft warning or remove DSA?

Specifically:
https://github.com/warewulf/warewulf/blob/main/internal/pkg/configure/ssh.go#L26

### Rationale

When we run a wwctl configure --all there is a failure due to DSA being disabled in FIPS.

wwctl configure --all
Building overlay for warewulf-001.local: host
Enabling and restarting the DHCP services
Building overlay for warewulf-001.local: host
Enabling and restarting the NFS services
Updating system keys
Skipping, key already exists: ssh_host_rsa_key
Setting up key: ssh_host_dsa_key
DSA keys are not allowed in FIPS mode
ERROR  : Failed to exec ssh-keygen: exit status 255
ERROR  : failed to exec ssh-keygen command: exit status 255

### Description

Can this be changed to a soft warning or remove DSA?
Perhaps replace the "exit 255" with "reply_warn"?

### Additional information

https://github.com/warewulf/warewulf/blob/main/internal/pkg/configure/ssh.go#L26
as well as a few other places I believe.

### General information

mslacken:
Was a low hanging fruit

- Close #1185

### DIFF
--- a/internal/app/wwctl/configure/ssh/root.go
+++ b/internal/app/wwctl/configure/ssh/root.go
@@ -12,6 +12,7 @@ var (
 			"keys.",
 		RunE: CobraRunE,
 	}
+	keyTypes []string
 )
 
 func init() {
@@ -19,5 +20,6 @@ func init() {
 
 // GetRootCommand returns the root cobra.Command for the application.
 func GetCommand() *cobra.Command {
+	baseCmd.PersistentFlags().StringArrayVarP(&keyTypes, "keytypes", "t", []string{"rsa", "dsa", "ecdsa", "ed25519"}, "ssh key types to be created")
 	return baseCmd
 }
--- a/internal/app/wwctl/configure/ssh/root.go
+++ b/internal/app/wwctl/configure/ssh/root.go
@@ -1,6 +1,8 @@
 package ssh
 
-import "github.com/spf13/cobra"
+import (
+	"github.com/spf13/cobra"
+)
 
 var (
 	baseCmd = &cobra.Command{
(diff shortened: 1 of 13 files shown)


JasonYangShadow (approved):
lgtm! feel free to merge.

internal/app/wwctl/configure/ssh/root.go
```diff
@@ -12,12 +12,14 @@ var (
...
 	}
+	keyTypes []string
 )
 
 func init() {
 }
 
 // GetRootCommand returns the root cobra.Command for the application.
 func GetCommand() *cobra.Command {
+	baseCmd.PersistentFlags().StringArrayVarP(&keyTypes, "keytypes", "t", []string{"rsa", "dsa", "ecdsa", "ed25519"}, "ssh key types to be created")
```
anderbubble, line 23: Can this value also be specified in `warewulf.conf`? If not, I don't think this could be done with `wwctl configure -a`. And wouldn't it need to be specified every time you run `wwctl configure`?

anderbubble (changes requested):

internal/app/wwctl/configure/ssh/root.go
anderbubble: Also, please add documentation for it. :)

mslacken: Its now in the delivered `warewulf.conf` which is also the documentation for all this stuff.

anderbubble:
Rebased on main.

anderbubble:
Rebased to main.

anderbubble:
@mslacken I think I'm likely to go ahead and merge this; but we should probably plan to add support for specifying the user key type as well.

https://github.com/warewulf/warewulf/pull/1187/files#diff-4bec3007d09da934bb849ebf67fe39762b6191ea05f3e36c955e5c463d24a839L50-L66

anderbubble:
@mslacken I had an idea, so I went ahead and added using this for authentication ssh keys, too. It just uses the first key type in the list.

Let me know what you think.

anderbubble (approved):

anderbubble:
@mslacken I just found myself wondering what actually happens to the hostkey overlays if some of the keys don’t get built: does ssh still start up ok on the compute node?

PR #1191, opened 2024-04-16
Dracut initramfs boot

### LINKED ISSUE #1115: Add a bootstrap stage to allow a two stage node boot process

anderbubble:

Support booting a dracut image from a container and using it to bootstrap a large container.

### To-do

- [ ] GRUB boot support
- [ ] Kernel override support

 - Closes #1115

### DIFF
--- a/dracut/modules.d/90wwinit/parse-wwinit.sh
+++ b/dracut/modules.d/90wwinit/parse-wwinit.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+# root=wwinit
+
+[ -z "$root" ] && root=$(getarg root=)
+
+if [ "${root}" = "wwinit" ]
+then
+    info "root=${root}"
+    export wwinit_container=$(getarg wwinit.container=); info "wwinit.container=${wwinit_container}"
+    export wwinit_system=$(getarg wwinit.system=); info "wwinit.system=${wwinit_system}"
+    export wwinit_runtime=$(getarg wwinit.runtime=); info "wwinit.runtime=${wwinit_runtime}"
+    export wwinit_kmods=$(getarg wwinit.kmods=); info "wwinit.kmods=${wwinit_kmods}"
+
+    wwinit_tmpfs_size=$(getarg wwinit.tmpfs.size=)
+    if [ -n "$wwinit_tmpfs_size" ]
+    then
+        info "wwinit.tmpfs.size=${wwinit_tmpfs_size}"
+        export wwinit_tmpfs_size_option="-o size=${wwinit_tmpfs_size}"
+    fi
+
+    if [ -n "${wwinit_container}" ]
+    then
+        info "Found root=${root} and a Warewulf container image. Will boot from Warewulf."
+        rootok=1
+    else
+        die "Found root=${root} but no container image. Cannot boot from Warewulf."
+    fi
+fi
--- a/internal/pkg/container/config.go
+++ b/internal/pkg/container/config.go
@@ -1,9 +1,23 @@
 package container
 
 import (
+	"fmt"
 	"path"
 
+	"github.com/warewulf/warewulf/internal/pkg/util"
+	"github.com/warewulf/warewulf/internal/pkg/wwlog"
+
 	warewulfconf "github.com/warewulf/warewulf/internal/pkg/config"
+)
+
+var (
+	initramfsSearchPaths = []string{
+		// This is a printf format where the %s will be the kernel version
+		"boot/initramfs-%s",
+		"boot/initramfs-%s.img",
+		"boot/initrd-%s",
+		"boot/initrd-%s.img",
+	}
 )
 
 func SourceParentDir() string {
@@ -26,4 +40,17 @@ func ImageParentDir() string {
 
 func ImageFile(name string) string {
 	return path.Join(ImageParentDir(), name+".img")
+}
+
+// InitramfsBootPath returns the dracut built initramfs path, as dracut built initramfs inside container
+// the function returns host path of the built file
+func InitramfsBootPath(image, kver string) (string, error) {
+	for _, searchPath := range initramfsSearchPaths {
+		initramfs_path := path.Join(RootFsDir(image), fmt.Sprintf(searchPath, kver))
+		wwlog.Debug("Looking for initramfs at: %s", initramfs_path)
+		if util.IsFile(initramfs_path) {
+			return initramfs_path, nil
+		}
+	}
+	return "", fmt.Errorf("Failed to find a target kernel version initramfs")
 }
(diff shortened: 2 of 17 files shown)


anderbubble:
@JasonYangShadow I refactored our work here, squashing redundant commits and making sure everything had a clear progression and commit messages. Let me know what you think.

internal/pkg/warewulfd/provision.go
```diff
@@ -103,7 +107,8 @@ func ProvisionSend(w http.ResponseWriter, req *http.Request) {
 			Hwaddr:         rinfo.hwaddr,
 			ContainerName:  node.ContainerName.Get(),
 			KernelArgs:     node.Kernel.Args.Get(),
-			KernelOverride: node.Kernel.Override.Get()}
+			KernelOverride: node.Kernel.Override.Get(),
+			NetDevs:        tstruct.NetDevs}
```
anderbubble, line 111: @JasonYangShadow can I get you to add a test for my changes here?

internal/pkg/container/config.go
```diff
@@ -27,3 +28,19 @@ func ImageParentDir() string {
...
+		return image_path
+	}
+
+	image_path = path.Join(RootFsDir(image), "boot", "initrd-"+kver)
+	if util.IsFile(image_path) {
+		return image_path
+	}
+
+	return ""
+}
```
anderbubble, line 46: @JasonYangShadow I had to add an additional search pattern for Suse. I'm not super happy with this implementation, so feel free to improve it.

I squashed this into your commit, so it's still attributed to you. Let me know if you have any concerns.

internal/pkg/warewulfd/provision.go
JasonYangShadow: @anderbubble sure will do that

dracut/modules.d/90wwinit/parse-wwinit.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
JasonYangShadow, line 1: is it /bin/bash here?

internal/pkg/warewulfd/provision.go
JasonYangShadow: I've pushed test cases in this commit
https://github.com/warewulf/warewulf/pull/1191/commits/75c70fe3e2893808f9bff97049a1cb1b434c55b2

internal/pkg/container/config.go
JasonYangShadow: I've made some changes in this commit
https://github.com/warewulf/warewulf/pull/1191/commits/75c70fe3e2893808f9bff97049a1cb1b434c55b2

internal/pkg/warewulfd/provision.go
anderbubble: Thanks!

anderbubble:
Rebased on main.

dracut/modules.d/90wwinit/parse-wwinit.sh
anderbubble: The other modules I looked at use `/bin/sh`. I kept this same pattern thinking that a POSIX shell may be all that is available in the initramfs environment.

internal/pkg/container/config.go
anderbubble: @JasonYangShadow I changed it a bit against to more closely match what's in `internal/pkg/kernel/kernel.go`.

I also removed a test case that dependendent on the order of files being returned, as I'd consider that undefined. (I know the golang Walk makes it deterministic, but I'd rather that not be an assumption.

anderbubble:
@mslacken @jcsiadal and @griznog I'd still like to add Grub support and see if I can get kernel overrides working for this; but I'd love to get your thoughts on getting this merged in time for v4.5.2. Let me know what you think.

Thanks!

mslacken:
If we link this together with #815 we have state full install. Steps for this would be:
* add ignition to dracut
* fetch the igniton.json from the warewulf host, which was the purpose of #815
* make sure that $NEWROOT is correctly set

This would allow a dump of the container to the rootfs. What is till missing would be checks, if the rootfs needs to be updated or reinstalled. I think this could achieved by placing configurations in `/warewulf` on the node.

mslacken (approved):

mslacken:
@anderbubble my commit is WIP, some args get lost

mslacken:
@anderbubble @JasonYangShadow Buidling the dracut works now on `wwctl container build` feel free to test it.

anderbubble:
@mslacken I'm torn on your contribution here. For example, I definitely appreciate this fix:

https://github.com/warewulf/warewulf/pull/1191/commits/9f1d378bf8a6a26a30bc914f6b711c466f7296ef#diff-56ccdd3cf2e2cb4ea3726e8587dbf75472b67d8ba3c25b47b3b7d9a80135ef3dL167-R174

But, ultimately, I do not think that `container build` should presumptively build a dracut image just because it's installed. More likely I think dracut should get called by `/etc/warewulf/container_exit.sh`.

For now, I want to omit this change from the PR; but I'll try to make sure it doesn't get lost so that we can still discuss it.

anderbubble:
> I'll try to make sure it doesn't get lost so that we can still discuss it.

Captured at https://github.com/anderbubble/warewulf/tree/dracut-initramfs-autobuild

PR #1220, opened 2024-05-10
Ipv6 support for provisioning nodes

dacianstremtan:

fixing or enhancing.

 - Fixes #

- [ ] Based this PR against the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))

### DIFF
--- a/internal/pkg/config/root.go
+++ b/internal/pkg/config/root.go
@@ -184,10 +184,10 @@ func (conf *RootConf) SetDynamicDefaults() (err error) {
 	}
 	// check validity of ipv6 net
 	if conf.Ipaddr6 != "" {
-        if conf.Ipv6net == "" {
-            wwlog.Error("Ipv6 network has not been set in warewulf.conf: ipv6net")
-            return errors.New("Invalid Ipv6 network")
-        }
+		if conf.Ipv6net == "" {
+			wwlog.Error("Ipv6 network has not been set in warewulf.conf: ipv6net")
+			return errors.New("Invalid Ipv6 network")
+		}
 		_, ipv6net, err := net.ParseCIDR(conf.Ipv6net)
 		if err != nil {
 			wwlog.Error("Invalid Ipv6 address specified, must be CIDR notation: %s", conf.Ipv6net)
--- a/internal/pkg/config/root.go
+++ b/internal/pkg/config/root.go
@@ -29,10 +29,10 @@ type RootConf struct {
 	WWInternal      int           `yaml:"WW_INTERNAL"`
 	Comment         string        `yaml:"comment,omitempty"`
 	Ipaddr          string        `yaml:"ipaddr"`
-	Ipaddr6         string        `yaml:"ipaddr6,omitempty"`
+	Ipaddr6         string        `yaml:"Ipaddr6,omitempty"`
 	Netmask         string        `yaml:"netmask"`
 	Network         string        `yaml:"network,omitempty"`
-	Ipv6net         string        `yaml:"ipv6net,omitempty"`
+	Ipv6net         string        `yaml:"Ipv6net,omitempty"`
 	Fqdn            string        `yaml:"fqdn,omitempty"`
 	Warewulf        *WarewulfConf `yaml:"warewulf"`
 	DHCP            *DHCPConf     `yaml:"dhcp"`
@@ -184,14 +184,18 @@ func (conf *RootConf) SetDynamicDefaults() (err error) {
 	}
 	// check validity of ipv6 net
 	if conf.Ipaddr6 != "" {
-		_, ipv6net, err := net.ParseCIDR(conf.Ipaddr6)
+        if conf.Ipv6net == "" {
+            wwlog.Error("Ipv6 network has not been set: Ipv6net")
+            return errors.New("Invalid Ipv6 network")
+        }
+		_, ipv6net, err := net.ParseCIDR(conf.Ipv6net)
 		if err != nil {
-			wwlog.Error("Invalid ipv6 address specified, mut be CIDR notation: %s", conf.Ipaddr6)
-			return errors.New("invalid ipv6 network")
+			wwlog.Error("Invalid Ipv6 address specified, must be CIDR notation: %s", conf.Ipv6net)
+			return errors.New("Invalid Ipv6 network")
 		}
 		if msize, _ := ipv6net.Mask.Size(); msize > 64 {
 			wwlog.Error("ipv6 mask size must be smaller than 64")
-			return errors.New("invalid ipv6 network size")
+			return errors.New("Invalid Ipv6 network size")
 		}
 	}
 	return
--- a/internal/pkg/config/root.go
+++ b/internal/pkg/config/root.go
@@ -29,10 +29,10 @@ type RootConf struct {
 	WWInternal      int           `yaml:"WW_INTERNAL"`
 	Comment         string        `yaml:"comment,omitempty"`
 	Ipaddr          string        `yaml:"ipaddr"`
-	Ipaddr6         string        `yaml:"Ipaddr6,omitempty"`
+	Ipaddr6         string        `yaml:"ipaddr6,omitempty"`
 	Netmask         string        `yaml:"netmask"`
 	Network         string        `yaml:"network,omitempty"`
-	Ipv6net         string        `yaml:"Ipv6net,omitempty"`
+	Ipv6net         string        `yaml:"ipv6net,omitempty"`
 	Fqdn            string        `yaml:"fqdn,omitempty"`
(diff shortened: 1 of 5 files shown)


anderbubble:
Thanks for the PR, @dacianstremtan! I look forward to reviewing it.

internal/pkg/config/root.go
```diff
@@ -29,10 +29,10 @@ type RootConf struct {
 	WWInternal      int           `yaml:"WW_INTERNAL"`
 	Comment         string        `yaml:"comment,omitempty"`
 	Ipaddr          string        `yaml:"ipaddr"`
-	Ipaddr6         string        `yaml:"ipaddr6,omitempty"`
+	Ipaddr6         string        `yaml:"Ipaddr6,omitempty"`
 	Netmask         string        `yaml:"netmask"`
```
mslacken, line 33: I would like to keep `ipaddr6` in the `nodes.conf` yaml file, as we have already `ipaddr` without capitalization.

dacianstremtan: Unfortunately, when it's using lower case "ipaddr6", the script is not picking up the Ipaddr6 declaration from the warewulf.conf file. It's being skipped. I did not know where this was declared to make sure that the case is ignored.

anderbubble: > the script

Which script? Or do you mean that a configured ipaddr6 value isn't actually available within Warewulf?

dacianstremtan: The value ipaddr6 for the Warewulf controller node is not being picked up from the warewulf.conf

mslacken: I think we have the following situation, in `warewulf.conf` we use `ipaddr6`, but in all the templates (file ending with `.ww` it's `Ipaddr6` as this is an exported variable in golang which *must* be capitalized.
We have similar problems for the variables in `nodes.conf`, but we might need to sync the variable names in the long run.

dacianstremtan: I updated the Ipaddr6 to use lower case. The ipaddr6 parameter in the /etc/warewulf/warewulf.conf should be with lower cased and all works fine:
e.g. 
ipaddr6: 2606:69c0:9120:786f:44:6468:6370:2
ipv6net: 2606:69c0:9120:786f::/64

Also we should be declaring an ipv6net instead of adding the network mask to the IPv6 address. When using IPv6 with network mask it breaks downloading the container/overlays files from ipxe via the IPv6 IP(/etc/warewulf/ipxe/default.ipxe):

```
set uri_base http://{{.Ipaddr6}}:{{.Port}}/provision/{{.Hwaddr}}?assetkey=${asset}&uuid=${uuid}
echo Warewulf Controller: {{.Ipaddr6}}
```

mslacken: @anderbubble we should merge this quite fast, as when testing this pr I realized that the actual ipv6 implementation is broken. This is also a bug fix.

dacianstremtan:
I don't understand the formatting errors.

mslacken:
@dacianstremtan I have tried to boot some nodes with ipv6, but it seems that templates for dhpc6 are missing. If you have these templates can you please add them to this PR? 
Also can you drop some lines of documentation of how to deploy nodes with ipv6?

dacianstremtan:
@mslacken I don't know which templates you are referring to.
The dhcpd6.conf configuration was done manually. I can upload the configuration file here if you would like to see it. I can also schedule a time to go over and demo the booting process.

mslacken:
Can you attach your `dhcpd6.conf` to this PR? From there on we can make a template out of it.

dacianstremtan:
Here are the configuration files:

The iPXE environment: ipxe-snponly-x86_64.efi does not support setting the IPv6 via DHCP only via router advertisements using SLAAC, therefore I "tricked" it by adding the filename option to the dhcpd.conf to know where to get the kernel/container/overlay files. Both dhcpd and dhcpd6 must be running to successfully boot via IPv6. 

[warewulf.conf.txt](https://github.com/user-attachments/files/15817381/warewulf.conf.txt)
[dhcpd.conf.txt](https://github.com/user-attachments/files/15817384/dhcpd.conf.txt)
[dhcpd6.conf.txt](https://github.com/user-attachments/files/15817385/dhcpd6.conf.txt)
[default.ipxe.txt](https://github.com/user-attachments/files/15817452/default.ipxe.txt)

anderbubble:
Rebased to main.

internal/pkg/warewulfd/parser.go
```diff
@@ -44,8 +47,17 @@ func parseReq(req *http.Request) (parserInfo, error) {
...
+    ipaddrtemp = ""
+    ipaddr = strings.Split(req.RemoteAddr, ":")
+    // IPv6 addresses hextets are split on ":"
+    if len(ipaddr) > 2 {
+       for i := 0; i < len(ipaddr)-1; i++ {
+            ipaddrtemp += ipaddr[i] + ":"
+       }
+       ret.ipaddr6 = strings.TrimSuffix(ipaddrtemp, ":")
+    }
 	ret.ipaddr = strings.Split(req.RemoteAddr, ":")[0]
```
anderbubble, line 59: Should `ret.ipaddr` not be set if `len(ipaddr) > 2`? What effect would this have elsewhere? What effect would it have for `ret.ipaddr` to be set to the first part of an ipv6 address (which seems to be what happens otherwise)?

Maybe something like

```
parts := string.Split(req.RemoteAddr, ":")
if len(parts) > 2 {
    ret.ipaddr6 = strings.Join(parts[:len(parts)-1], ":")
} else {
    ret.ipaddr = parts[0]
}
```

internal/pkg/warewulfd/parser.go
```diff
@@ -44,8 +47,17 @@ func parseReq(req *http.Request) (parserInfo, error) {
...
 	ret.hwaddr = hwaddr
+    ipaddrtemp = ""
+    ipaddr = strings.Split(req.RemoteAddr, ":")
+    // IPv6 addresses hextets are split on ":"
+    if len(ipaddr) > 2 {
+       for i := 0; i < len(ipaddr)-1; i++ {
+            ipaddrtemp += ipaddr[i] + ":"
+       }
+       ret.ipaddr6 = strings.TrimSuffix(ipaddrtemp, ":")
+    }
```
anderbubble, line 58: I think this would be simpler with something like

```
parts := string.Split(req.RemoteAddr, ":")
if len(parts) > 2 {
    ret.ipaddr6 = strings.Join(parts[:len(parts)-1], ":")
}
```

internal/pkg/warewulfd/parser.go
```diff
@@ -23,7 +24,9 @@ type parserInfo struct {
 
 func parseReq(req *http.Request) (parserInfo, error) {
 	var ret parserInfo
-
+    var ipaddr []string
+    var ipaddrtemp string
```
anderbubble, line 28: I suggest that these variable declarations should be made implicitly with `:=` when they are used below.

internal/pkg/warewulfd/parser.go
```diff
@@ -44,8 +47,17 @@ func parseReq(req *http.Request) (parserInfo, error) {
...
+    // IPv6 addresses hextets are split on ":"
+    if len(ipaddr) > 2 {
+       for i := 0; i < len(ipaddr)-1; i++ {
+            ipaddrtemp += ipaddr[i] + ":"
+       }
+       ret.ipaddr6 = strings.TrimSuffix(ipaddrtemp, ":")
+    }
 	ret.ipaddr = strings.Split(req.RemoteAddr, ":")[0]
-	ret.remoteport, _ = strconv.Atoi(strings.Split(req.RemoteAddr, ":")[1])
+	ret.remoteport, _ = strconv.Atoi(strings.Split(req.RemoteAddr, ":")[len(ipaddr)-1])
```
anderbubble, line 60: I think it's awkward to use `ipaddr` as a general-purpose indicator for the length of `strings.Split(req.RemoteAddr, ":")`. `strings.Split(req.RemoteAddr, ":")` could just be replaced here with `ipaddr`; but I think `parts` from my example above is clearer:

```
parts := string.Split(req.RemoteAddr, ":")
if len(parts) > 2 {
    ret.ipaddr6 = strings.Join(parts[:len(parts)-1], ":")
} else {
    ret.ipaddr = parts[0]
}
ret.remoteport, _ = strconv.Atoi(parts[len(parts)-1])
```

anderbubble (changes requested):
Thanks for the PR! And sorry it took me so long to provide a review. I won't be merging it in v4.5.5; but I'll move it to v4.5.6 in hopes of merging it then.

Let me know if you have any questions, comments, or disagreements with my suggestions.

anderbubble:
Resolved some formatting errors.

anderbubble:

PR #1221, opened 2024-05-13
Fix CVE-2024-3727

mslacken:
@anderbubble can we release a 4.5.2 with this fix as fast as possible?


anderbubble:
@mslacken I'm doing this now. I can't push into your branch to update it, so I'm making updates at anderbubble/FixCVE-2024-3727.

anderbubble:
Closed by #1222.

PR #1229, opened 2024-05-24
Simplify `wwctl server` command and logging

### LINKED ISSUE #503: warewulfd log to stdout
One step further from #152, I propose that warewulfd, by default, especially when running in the foreground, should log to stdout, as indicated at https://12factor.net/logs

The warewulf log file path, which is currently a hard-coded const at https://github.com/hpcng/warewulf/blob/development/internal/pkg/warewulfd/daemon.go#L22, should be migrated to a configuration setting. If this setting is set, then logs go to a file; but if it is left undefined, or set to an empty string, then logs should be written unbuffered to stdout. This would allow, e.g., the systemd journal to pick them up.

### LINKED ISSUE #508: do not daemonize or write pid files
I've been reading [the twelve-factor app docs](https://12factor.net/concurrency) and specifically this part:

> Twelve-factor app processes [should never daemonize](http://dustin.github.com/2010/02/28/running-processes.html) or write PID files. Instead, rely on the operating system’s process manager (such as [systemd](https://www.freedesktop.org/wiki/Software/systemd/), a distributed process manager on a cloud platform, or a tool like [Foreman](http://blog.daviddollar.org/2011/05/06/introducing-foreman.html) in development) to manage [output streams](https://12factor.net/logs), respond to crashed processes, and handle user-initiated restarts and shutdowns.

As such, I propose the following changes:

- remove `wwctl server [restart|stop|status]`
- remove `wwctl server start --foreground` (it should always be foreground)
- stop writing pidfiles
- update warewulfd.service to use new foreground semantics (and maybe rename it to warewulf-server, since there isn't actually a warewulfd process)

But I recognize this is a broad and high-impact suggestion: hence the "rfc" tag for this issue. I'd love to hear thoughts on the viability of a proposal like this.

### LINKED ISSUE #725: the warewulf server should be able to omit send/recv log messages
### Version of Warewulf

4.4

### Expected behavior

It should be possible, either through a setting in warewulf.conf or a command-line argument to `wwctl server start`, to decrease the verbosity of the warewulf log such that send and recv messages are not written.

### Actual behavior

send/recv messages are always written, and verbosity can only be increased.

anderbubble:

- Always run in the foreground
- Always log to stdout/err (unless syslog is requested)
- Update systemd to reload with a direct signal
- Omit any pidfile for "warewulfd"

## Issues affected

- Closes #508 
- Closes #503 
- Closes #725


anderbubble:
@mslacken and @griznog I've rebased this to main. I tested it before, but I haven't re-tested it yet, pending #1474.

Go ahead and review; but I don't think it should be merged until after #1474.

griznog (approved):
LGTM

PR #1233, opened 2024-05-29
Adjusts services open file limit to match systemd precedence

JellyKid:

If you have 1000s of nodes and many files you may run into the issue of hitting the warewulfd service's open file limits. This small change to the systemd unit file will effectively double the number of open files that the service can have open and is in line with systemd precedence of setting services to a high file limit of 512*1024. 

Reference:
[Systemd high nofile limit](https://github.com/search?q=repo%3Asystemd%2Fsystemd%20HIGH_RLIMIT_NOFILE&type=code)
[How to set limits for services in RHEL and systemd](https://access.redhat.com/solutions/1257953)

 - Had a discussion about this on the Slack channel but I'd be happy to open an issue about it.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -53,3 +53,4 @@
 - Karker Said <said.karker@gmail.com> [@kosmolito](https://github.com/kosmolito)
 - Jason Scott <jsco2t@outlook.com> @jsco2t
 - Sujeev Uthayakumar <sujeev.uthayakumar@gmail.com> [@Sujeev-Uthayakumar](https://github.com/Sujeev-Uthayakumar)
+- Jacob Sommerville <jsommerville@dow.com>
--- a/include/systemd/warewulfd.service.in
+++ b/include/systemd/warewulfd.service.in
@@ -11,6 +11,7 @@ User=root
 Group=root
 ExecStart=@BINDIR@/wwctl server $OPTIONS
 ExecReload=/bin/kill -HUP "$MAINPID"
+LimitNOFILE=524288
 Restart=always
 
 [Install]


anderbubble:
Thanks for the PR, @JellyKid! I'll review and let you know if I need anything else.

anderbubble:
Looking at my test system, I see a prior limit of 524288 on a Rocky 9 system.

```
[janderson@rocky v4.5.x]$ sudo cat /proc/$(pidof wwctl)/limits 
Limit                     Soft Limit           Hard Limit           Units     
[...]
Max open files            524288               524288               files     
[...]
```

@JellyKid were you really hitting this limit already? I admit that I'm surprised, even with multiple thousands of nodes.

My initial thought is that if we're going to increase the limit so high, why not just set it to unlimited / infinity? Then that further made me wonder why not just consider this something that a local cluster admin might need to adjust themselves for their site, if they know their cluster really is that large?

JellyKid:
@anderbubble This was on a Rocky 8 system. Perhaps there was a change between 8 and 9. I was trying to figure out if it was a good/bad idea to increase it to infinity and saw a few threads about oddities when increasing it to that depending on distro. I was going based on what Systemd already had as a precedence. I have a test warewulf head node running Rocky 8 as well, which shows the reduced Hard Limits
image

I tried increasing the fs.file-max kernel param and increasing the ulimits. The only thing that fixed issue was increasing the unit file. We've been running without issue since I made the change last week, when previously we would get the open file limit error after a few hours.

mslacken:
We should make this an issue as with 1000s of nodes  there might a problem with too many open files due to the serving the overlays to `wwclient`.
A mitigation in this case, could be to increase the time between overlay updated. Also in the long run, we should only transfer the overlays if something has changed.

anderbubble:

mslacken:
@JellyKid can you please rebase, I think we should merge it.

mslacken (approved):

PR #1262, opened 2024-06-13
support for search in resolv.conf

mslacken:
Add support for tag SEARCH for resolv.conf


anderbubble:
Looks like a good start! But I think we need to configure it for the different network providers, too, since resolv.conf is subject to being rewritten.

anderbubble:
Rebased on main.

anderbubble:
Got an error with the change to slice:

```
ERROR: could not execute template: template: resolv.conf.ww:10:17: executing "resolv.conf.ww" at <slice $tk 0 3>: error calling slice: list should be type of slice or array but string
```

anderbubble:

PR #1275, opened 2024-06-21
Check for empty container var

mslacken:
- Check if container is defined during boot process and give proper error otherwise


anderbubble (changes requested):
@mslacken this PR removes dracut support in Grub. Can you rebase and restore it?

mslacken:
Can't be rebased properly, so reimplementing

PR #1277, opened 2024-06-21
better udev net rules

mslacken:
- Renamed 70-ww4-netname.rules.ww
  to 70-persistent-net.rules.ww as this seems to be quasi standard file
- Better handling of InfiniBand udev net naming


e4t:
There is probably no need to rename  `70-ww4-netname.rules.ww` to `70-persistent-net.rules.ww`. `/etc/udev/rules.d/70-persistent-net.rules`  is the name of the file where udev (or better `/usr/lib/udev/write_net_rules` was looking for persistent nework device names and was editing it on older udev versions. Apparently, this is gone on later versions of udev. 
To disable automatic device naming of unconfigured devices on older udev versions, one may set the kernel parameter `net.ifnames=1`.

e4t:
This commit is better. 
The `net.ifnames=1` setting is the system default, still on SLE & Leap it requires to be set explicitly to muffle some backward compatibility handling which is a bit quirky. Since this setting is the default (ie. when nothing is set) it should not make any difference - except on SLE & Leap where it is required to get rid of the backward compatibility handling.

internal/pkg/node/constructors.go
```diff
@@ -27,7 +27,7 @@ defaultnode:
   system overlay:
   - wwinit
   kernel:
-    args: quiet crashkernel=no vga=791 net.naming-scheme=v238
+    args: quiet crashkernel=no net.ifnames=1
```
anderbubble, line 30: I'm in favor of removing `vga=791` and `net.naming-scheme=v238`; but isn't `net.ifnames=1` equivalent to just not specifying the parameter? If so, I suggest just removing these (and maybe the _other_ default kernel arguments, as well. Why does it need to default to `quiet` and `crashkernel=no`?

In any case, though, I think that this is too large a change for v4.5.x; so I'm moving it to v4.6.0.

anderbubble (changes requested):

griznog:
It'll be a week or possibly more before I cat test this, but since my previous attempts at renaming infiniband devices have all been abject failures I'd say merge this into `main` and I'll update and test in the real world as soon as I can. At least from my perspective it can't make things any worse :smile:

griznog (approved):
Merge it and I'll try it :smile:

anderbubble (approved):

PR #1283, opened 2024-06-26
enhanced container list

### LINKED ISSUE #954: Improve the wwctl container list command
### Summary

Improve the output of the `wwctl container list` command to show a more usefull size value by spliting it into several values. E.g. IMG (GZ) SIZE, for the uncompressed image size (compressed maybe into parenthesis) and TOTAL USE SIZE, for the total use in disk (as it is shown currently).

### Rationale

There is a discrepancy in the size shown by list command and the du and ls Linux commands and also "size" value can be confusing.

### Description

While using `wwctl container list` command the size of the image is shown as the addition of the chroot subdirectory size, and the .img and .img.gz files. It can be confusing since sizes are calculated with Go, and differs, in some cases significatively, from Linux commands (du and ls).

Also it is not really useful in terms of booting a node since not the whole space of a container in disk is loaded to a node but the img file size, which can be determinant, specially with Legacy boot.

Not really related, but columns tabs/spaces could be improved too, in order to have the values properly aligned to the headers.

### Additional information

Suggestion of output:
```
CONTAINER NAME   NODES  KERNEL VERSION   CREATION TIME        MODIFICATION TIME    IMG (GZ) SIZE   TOTAL USE SIZE
rocky-8-base   0      4.18.0-477.15.1.el8_8.x86_64 08 Aug 23 12:25 CEST 09 Aug 23 13:39 CEST 1.7 GiB (774 MiB) 4.2 GiB
```

Suggested changes:
- Two more size vars in `internal/pkg/api/container/container.go`
- New column(s) in `internal/app/wwctl/container/list/main.go`

### General information

### LINKED ISSUE #1117: fast container list by default
### Summary

Adjust `wwctl container list` to only report the size of the image by default. Additional arguments may be introduced to query the size of the compressed archive (`--compressed-size`) and/or the chroot directory (`--chroot-size`), probably mutually exclusively.

An alternative to #954.

### Rationale

`wwctl container list` takes longer to run than it should. This is because the container list reports size as the combination of the image, the compressed image, and the chroot directory, with sizing the chroot directory taking the most time. This is both confusing and unnecessary.

### Description

_No response_

### Additional information

_No response_

### General information

mslacken:
Container list has now options for full, kernel and size listing. Without any
arguments only available containers are listed. Size is no more fine granular
showing also size of the image, chroot and compressed image

- Closes #1117 
- Closes #954


anderbubble (changes requested):
My suggestions:

* `-f` and `--full` should be replaced with `-l` and `--long`, which I think is more expected. (`-f` is more associated with "force")
* I think size should be consolidated back down to a single value, which should default to the size of the uncompressed image (since that's the size of the image as-deployed).
* Optionally, additional values `--compressed` and `--chroot` could be added to gather and report those value _in stead of_ the default value.

See also #1117.

anderbubble:
I made a small functional change: specifying `--chroot` or `--compressed` now implies `--size`.

anderbubble (approved):

PR #1286, opened 2024-07-02
ignore +:: style lines in passwd/group

mslacken:
The lines +::::: or +::: are used in passwd/group as
placeholer for users added form the network via ldap/nis/sssd.
These lines should not trigger a warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix the issue that new files created with wwctl overlay edit have 755 permissions. #1236
 - Mount `/sys` and `/run` during `wwctl container exec`. #1287
 
+### Changed
+
+- Ignore +:::: lines in passwd/group during syncuser. #1286
+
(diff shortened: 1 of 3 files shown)


anderbubble:
@mslacken I did some reading about this, and found that this handling should probably also be extended to

* entries that _start_ with `+`
* entries that start with (or are entirely) `-`

Since these lines _are_ being omitted from the final sync'd file, I think it's still appropriate to log something indicating that they're being skipped. I've added that new warning, as well as handling for `-` and more expansive handling for `+`, to a new commit in this branch; but I haven't squashed them together so that you can evaluate and let me know what you think. (For example, I could be persuaded that it should be a debug entry, rather than a warning, if you prefer it not be shown during a regular syncuser.)

Let me know what you think.

mslacken:
It's still a warning which is confusing and messing up the output.
We should just give an info log about this 'correct' entry, I would even prefer to make the message only shown on verbose level.
Warning always means that you should do something to make the warning go away, what is definitely no true in this case.

anderbubble (approved):

anderbubble (approved):

PR #1287, opened 2024-07-02
mount /sys on exec

mslacken:
SUSEConnect need `/sys` to be present in the container
for telemetry reasons. As `/dev/` is mounted we can also
mount `/sys`


anderbubble:
@mslacken after our conversation on Slack, I updated this as seen at https://github.com/anderbubble/warewulf/pull/new/MountSys.

Let me know what you think. It appears to work in my environment.

mslacken:
@anderbubble your approach looks better, can you make a PR?

PR #1291, opened 2024-07-03
Fix grub asset key handling

epruesse:

This fixes asset key retrieval in the rendered grub.cfg and adds it to the chainload grub.cfg, allowing asset key based grub boot to work. 

- `--type1` should be `--type 1`
- The `chassis-asset-tag` (as used elsewhere), is not in the type 1 tabke, but in the type 3 table. 
- If an asset tag is listed in the node config, it is checked by wwctl server on all requests, so everything using `(http,<ip:port)` must also use `?assetkey=${assetkey}`, including fetching the templated grub.cfg. 

Note that this is not a full fix. The grub.efi files supplied with Rocky 9.4 do not include the `smbios` command! This has been the case for many years, apparently. I have used a Debian box to run `grub2-makenetdir`, generating a working grub efi.

With that, it is now working.

For testing, you can add an asset key to a KVM VM like this:

```
qm set $VMID --args "-smbios type=3,asset=$KEY"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   on unused/unmanaged network interfaces.
 - Systems with no SMBIOS (Raspberry Pi) will create a UUID from
   `/sys/firmware/devicetree/base/serial-number`
+- Grub chainload and main configs correctly extract smbios asset key
+  for node authentication
 - Replace slice in templates with sprig substr. #1093
 - Fix an invalid format issue for the GitHub nightly build action. #1258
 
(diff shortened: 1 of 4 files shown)


epruesse:
@anderbubble Do you want me to add a comment to the docs about having to use the Debian Grub? IDK how else to fix this. If you go into the grub shell and type "smbios", it'll tell you the function isn't there with the regular Rocky 9.4 grub package. There is no error shown for this when running the config. It just leads to an empty $assetkey variable.

anderbubble:
@epruesse I haven't dug into this myself yet, so I'm not certain, and I'd be interested to hear @mslacken's perspective especially, since he developed this originally.

I don't think we'd want to be dependent on a specific OS's grub.

epruesse:
Well, depending on OS packages to have the correct features compiled in can't really be avoided. Do you have influence with the Rocky maintainers or better yet upstream with RHEL?

I don't see many alternatives for having a secret shared with a diskless server other than values like this. Although I suppose the entire security concept could be reevaluated.

anderbubble:
@mslacken can you review this?

mslacken (approved):
thx

Issue #1060, opened 2024-01-28
Create a symlink or collection of symlinks from a template


griznog:
### Summary

Add the ability to create a symlink from a template, similar to how the `/etc/sysconfig/network-scripts/ifcfg.ww` template works using `file $filename`. 

### Rationale

I've been enabling/disabling services via overlay by adding an `ExecStartPre=/path/to/script` to each service I want to control, then having that `script` use a templated value like `true`/`false` to either exit with 0 to allow the service to start or 1 to have the service go into `Failed` state. This works, more or less, but makes for an ugly boot with all the things disabled on the node failing and doesn't really work for mount/automount units. 

A better solution would be to be able to conditionally create the symlinks that enable the service based on the node config. 

### Description

SImilar to `file` with
```
{{- file $filename }}
```
producing a collection of files based on the template, the symlink command would be something like:

```
{{- symlink $targetpath $linkname}}
```

producing symlink(s) in the current directory e.g. doing `ln -s $targetpath $linkname`

So a template like this in `/etc/systemd/system/multi-user.target.wants/windows-mounts.ww`:
{{{
{{- $volumeLabels := split .Tags.windows_volume_labels "," }}
{{- range $index, $value := $volumeLabels }}
{{- $targetpath := print "../" $value ".automount" }}
{{- $linkname := print $value ".automount" }}
{{-  symlink $targetpath }}
}}}

Would result in a symlink that enables each automount service as configured from the comma separated list in `.Tags.windows_volume_labels `,

### General information

- [X] I have searched the issues of this repo and believe this is not a duplicate

PR #1303, opened 2024-07-12
create soft links from overlays

mslacken:
Add the the function softlink to the template rendering, so that
{{ softlink: "/bin/bash" }} for the file `/bin/run.ww` will create
a soft link to `/bin/bash`
It's also possible to "import" soft links from the host, e.g. the
template
{{ ImportLink "/etc/localtime" }}
set the timezone from the host.

### DIFF
--- a/internal/pkg/overlay/funcmap.go
+++ b/internal/pkg/overlay/funcmap.go
@@ -125,10 +125,14 @@ func createIgnitionJson(node *node.Node) string {
 }
 
 func importSoftlink(lnk string) string {
-	path, err := filepath.EvalSymlinks(lnk)
+	target, err := filepath.EvalSymlinks(lnk)
 	if err != nil {
 		return "abort"
 	}
-	wwlog.Debug("importing softlink pointing to: %s", path)
-	return fmt.Sprintf("{{ /* softlink \"%s\" */ }}", path)
+	wwlog.Debug("importing softlink pointing to: %s", target)
+	return softlink(target)
+}
+
+func softlink(target string) string {
+	return fmt.Sprintf("{{ /* softlink \"%s\" */ }}", target)
 }
--- a/internal/pkg/overlay/funcmap.go
+++ b/internal/pkg/overlay/funcmap.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	warewulfconf "github.com/warewulf/warewulf/internal/pkg/config"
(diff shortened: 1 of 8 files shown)


overlays/wwinit/rootfs/etc/localtime.ww
```diff
@@ -0,0 +1 @@
+{{ ImportLink "/etc/localtime" }}
```
anderbubble, line 1: Can you move this to a new overlay? Something like `localtime` maybe, unless there's another one that makes sense.

internal/pkg/overlay/funcmap.go
```diff
@@ -122,3 +123,12 @@ func createIgnitionJson(node *node.NodeConf) string {
...
 }
+
+func importSoftlink(lnk string) string {
+	path, err := filepath.EvalSymlinks(lnk)
+	if err != nil {
+		return "abort"
+	}
+	wwlog.Debug("importing softlink pointing to: %s", path)
+	return fmt.Sprintf("{{ /* softlink \"%s\" */ }}", path)
+}
```
anderbubble, line 134: Could this be replaced with a value that reads the target of a symlink, which could then be passed to the other symlink function? That justifies both functions (by using them both in the example localtime use case) and is more flexible for other templates later.

internal/pkg/overlay/overlay.go
```diff
@@ -238,12 +238,17 @@ func BuildOverlayIndir(nodeData node.NodeConf, overlayNames []string, outputDir
...
 					for fileScanner.Scan() {
 						line := fileScanner.Text()
-						filenameFromTemplate := reg.FindAllStringSubmatch(line, -1)
-						if len(filenameFromTemplate) != 0 {
+						filenameFromTemplate := regFile.FindAllStringSubmatch(line, -1)
+						softlinkFromTemplate := regLink.FindAllStringSubmatch(line, -1)
+						if len(softlinkFromTemplate) != 0 {
+							wwlog.Debug("Creating soft link %s -> %s", destFileName, softlinkFromTemplate[0][1])
+							return os.Symlink(softlinkFromTemplate[0][1], path.Join(outputDir, destFileName))
+						} else if len(filenameFromTemplate) != 0 {
```
anderbubble, line 251: This implicitly uses the symlink value if it exists, ignoring any `file` directive. I think specifying both `symlink` and `file` should be an error. Frankly, if there's a `symlink` directive and there's anything else generated that _isn't_ the `symlink` directive, I think that should probably be an error.

anderbubble (changes requested):
I _finally_ think I understand what this is for.

I left some comments in the requested changes. Let me know what you think.

anderbubble:
Rebased to main.

mslacken:
rebased

anderbubble:
Rebased to main.

anderbubble (approved):
@mslacken I added tests for the "localtime" overlay and made it configurable. If you approve, I'll go ahead and merge.

anderbubble:
Merging as discussed in 18 December 2024 TSC meeting.

anderbubble:
Closed #1060

PR #1020, opened 2023-12-15
removed NodeInfo

mslacken:
- removed NodeInfo and using NodeConf instead
- wwctl comands without NodeInfo
- api changes for use without NodeInfo
- fixed tests, mostly will use testenv now

---

- Closes #916

PR #1306, opened 2024-07-15
Split overlays into discrete functionality and test

### LINKED ISSUE #987: split default overlays into discrete functionality and test
### Summary

Split the `generic` and `wwinit` overlays into multiple discrete overlays based on their functionality. For example, `wwinit` might be split into `NetworkManager`, `ifcfg`, `udev`, etc.

These overlays should then be explicitly included by the `default` profile (not implicitly included by a built-in setting).

### Rationale

The monolithic overlays make it difficult to opt-out of discrete parts of built-in Warewulf functionality.

### Description

_No response_

### Additional information

_No response_

### General information

anderbubble:

Split `wwinit` and `generic` into multiple overlays based on their individual functionality.

- Closes #987


anderbubble:
@griznog and @mslacken this is definitely not done, but I would be very interested in your thoughts on this so far. In particularly, what do you think about where I split the overlays?

griznog:
Following the manifesto after looking at your breakout of stuff, something jumps out at me for which I am still struggling to find an answer to: how to handle services that interact with other services? `udev` for instance might need to pick up rules for any type of device, each of which might be managed by a different overlay. So should the `udev` overlay collect all these or should each individual overlay drop in it's own `udev` rules? 

It's more pronounced with something like `firewalld` where I need to have things opened for `sshd`, `caddy`, `ondemand`, `nfs-server`, ... Do I collect all those in a `firewalld` overlay or have each drop its rules into `/etc/firewalld/...` and then somehow let the `firewalld` overlay know what needs to be loaded? And if I split out to all the unique overlays, which seems intuitively the correct approach to me, what about support for `iptables-scripts` and `shorewall` and ...? 

Multiply that example by all the similar or worse ones, like setting up upstream authentication which touches PAM, `sssd`, namespaces, ssh, slurm, Samba, joining a domain, ...

It's also be nice if overlays were, to the extent possible, distro-agnostic. That will be really hard as it's super-non-trivial just to figure out what OS a node image is using from the node image itself, so you start that problem out by having to set a tag somewhere and introducing a place for human error.

griznog:
Also FWIW, I'd consider splitting these out lower priority than getting support for arbitrary YAML in Tags, as splitting becomes much more interesting once complex, nested data structures can be set in the node configs.

anderbubble:
@griznog I expect you won't find it perfect; but I adjusted the new overlay names to try to focus on what is being configured, rather than just a partial indication of using a given service to configure something. The best example, I think, is `udev.netname` and `systemd.netname`: both of these attempt to configure the system to use the desired network interface names; but one attempts to use udev and the other systemd to accomplish this.

The most broken overlay, in this regard, is currently `fstab`, as it mixes both NFS configuration and Ignition configuration. I have created #1375 to convert this overlay to use systemd mount units, so that they could be split into separate `nfs` and `ignition` overlays.

Let me know if you have any concerns.

anderbubble:
@mslacken I think this is ready. Please check it out and let me know what you think. I'd like to get this merged so that we can rebase #1020 onto it.

mslacken (approved):

mslacken (changes requested):
The PR doesn't apply any more after the merge of #1380 as there also the tests for the overlays were changed. Can you rebase?

anderbubble:
Will do!

anderbubble:
@mslacken rebase is done. Let me know if you have any further concerns.

mslacken (approved):

mslacken:
Moving the templates to own overlays is a huge benefit and we should get this merged soon.
Unfortunately the testing is not in the way I would like it.
I propose to have the tests for the overlay rendering in the same place as the overlay itself. This fits much better in the go testing framework, see #1384 as. POC
I know this will need some rework for this PR, but I think its worth the price.

anderbubble:
I appreciate the feedback, @mslacken. I like the idea in theory, as I understand it, and I will look at it to see what it would take to incorporate.

anderbubble:
@mslacken I incorporated your example from #1384. The added per-overlay flexibility made it easier to get even more of the overlays tested.

Let me know what you think.

anderbubble:
I just realized that I neglected to update warewulf.spec to omit the `internal` directories; we don't need/want to deploy the tests with the binary RPMs.

anderbubble:
> I just realized that I neglected to update warewulf.spec to omit the `internal` directories; we don't need/want to deploy the tests with the binary RPMs.

Done. Looks like I actually needed to change `Makefile`, not `warewulf.spec`.

anderbubble:
@mslacken I rebased on main. Ready to merge?

PR #1308, opened 2024-07-21
wwclient overlay replace changes only

griznog:
Remove the -u (--unconditional) from cpio in wwclient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Bump github.com/spf13/cobra from 1.7.0 to 1.8.0 #1166
 - Bump github.com/fatih/color from 1.15.0 to 1.17.0 #1224
 - Bump github.com/coreos/ignition/v2 from 2.15.0 to 2.19.0 #1239
+- Only replace or update files and symlinks that are newer in the overlay than on the node. #1307
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


anderbubble:
Adjusted changelog entry.

anderbubble:
@griznog I like this change; but on reflection I think it needs to be kept for v4.6.0. It makes sense to not unpack the same file every minute, and if that's all the change did then I wouldn't object to it being in v4.5.x; but this also changes the behavior of overlay application when a file has been changed out-of-band on the node. Right now, if I update a config file on the node, wwclient will put it back as it should be; but, with this change, wwclient will no longer update the node, unless the overlay changes again.

So I'll move it to the v4.6.0 milestone for now; but let me know if you object.

anderbubble:
@griznog on further consideration, I think we actually shouldn't do this, as it is likely to have unintended side-effects. cpio records the modification time of the file on disk; and rpm installs files with their original mtime as well. This would erroneously cause any file that comes from the OS that is newer than the overlay source itself (not newer than the overlay build-time, or newer than wwclient's original provision time) to override what's coming from the overlay.

I _think_ that this wouldn't impact overlay _templates_, as their mtime appears to be when the overlay was rendered during overlay build (it's a new file, after all); but any static file, and directories, would have, imo, surprising behavior.

I agree that this particular concern with overlays is real and should be addressed; but I'm afraid this particular solution would cause its own issues.

Let me know if you think I've misjudged, though.

PR #1320, opened 2024-08-01
Alternative URL /container-by-name/NAME for containers

posch:
By default, warewulf is using a node's MAC address to locate the image. The URL generated for ipxe or grub look like this:

  http://.../provision/{{.Hwaddr}}?...&stage=container&compress=gz

where Hwaddr contains the node's default interface MAC address.

Because containers are not specific to a node, an URL without MAC address could be used. For example:

  http://.../container-by-name/{{.ContainerName}}.img.gz

to refer to the compressed image.

This form can easily be used with standard HTTP proxy caches (e.g. nginx), since many nodes will download the same URL.

This idea to scale warewulf provisioning has been discussed briefly in the warewulf slack help channel (July 24th, @griznog, @mslacken).

We're currently using one proxy with warewulf and iPXE boot, and modified dracut.ipxe to contain:

```
set dracut_wwinit ... wwinit.container="http://$PROXY/images/{{.ContainerName}}.img.gz" ...
```

Without this change, the proxy (nginx) needs a configuration like this that contains some $FAKEMAC MAC address which can be a real node or a dummy node. This, however, is error prone, and you may need to duplicate the config for the uncompressed version.

```
        location /images/$NAMEOFIMAGE.img.gz {
                proxy_pass http://$WAREWULF:9873/provision/$FAKEMAC?assetkey=&stage=container&compress=gz;
                proxy_cache_valid 200 302 1h;
                proxy_cache mycache;
                expires max;
        }
```

With this change, the proxy config would be much simpler.

 - none

- [X] Read the [documentation for contributing to Warewulf](https://warewulf.org/docs/main/contributing/contributing.html)
- [X] Based this PR against the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,6 +161,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add man pages and command reference to userdocs. #1488
 - Document building images from scratch with Apptainer. #1485
 - Added warewulfd:/overlay-file/{overlay}/{path...}?render={id}
+- Added an alternative image URL /images/{{.ContainerName}} that does not contain a MAC addresses
 
 ### Changed
 
--- a/internal/pkg/config/buildconfig.go.in
+++ b/internal/pkg/config/buildconfig.go.in
@@ -46,6 +46,7 @@ type WarewulfConf struct {
 	AutobuildOverlaysP *bool `yaml:"autobuild overlays,omitempty" default:"true"`
 	EnableHostOverlayP *bool `yaml:"host overlay,omitempty" default:"true"`
 	GrubBootP          *bool `yaml:"grubboot,omitempty" default:"false"`
+	CacheControl       string `yaml:"cachecontrol" default:"max-age=60"`
 }
 
 func (conf WarewulfConf) Secure() bool {
--- a/internal/pkg/warewulfd/images.go
+++ b/internal/pkg/warewulfd/images.go
@@ -1,0 +1,41 @@
+package warewulfd
+
+import (
+	"net/http"
+	"path"
+	"strings"
+
+	"github.com/warewulf/warewulf/internal/pkg/image"
+	"github.com/warewulf/warewulf/internal/pkg/util"
+	"github.com/warewulf/warewulf/internal/pkg/wwlog"
+)
+
+func ImagesSend(w http.ResponseWriter, req *http.Request) {
+	wwlog.Debug("Requested URL: %s", req.URL.String())
+
+	url := strings.Split(req.URL.Path, "?")[0]
+	path_parts := strings.Split(url, "/")
+
+	if len(path_parts) != 3 {
+		w.WriteHeader(http.StatusBadRequest)
+		wwlog.Error("invalid /images/$name URL")
+		return
+	}
+
+	image_name := path_parts[2]
+	wwlog.Debug("images: %s", image_name)
+
+	stage_file := path.Join(image.ImageParentDir(), image_name)
+	wwlog.Serv("stage_file '%s'", stage_file)
+
+	if util.IsFile(stage_file) {
+		err := sendFile(w, req, stage_file, "")
+		if err != nil {
+			wwlog.ErrorExc(err, "")
+			return
+		}
+	} else {
+		w.WriteHeader(http.StatusNotFound)
+		wwlog.Error("images: not found: %s", stage_file)
+	}
+}
--- a/internal/pkg/warewulfd/images.go
+++ b/internal/pkg/warewulfd/images.go
@@ -5,6 +5,7 @@ import (
 	"path"
 	"strings"
 
+	warewulfconf "github.com/warewulf/warewulf/internal/pkg/config"
 	"github.com/warewulf/warewulf/internal/pkg/image"
 	"github.com/warewulf/warewulf/internal/pkg/util"
 	"github.com/warewulf/warewulf/internal/pkg/wwlog"
@@ -12,6 +13,7 @@ import (
 
 func ImagesSend(w http.ResponseWriter, req *http.Request) {
 	wwlog.Debug("Requested URL: %s", req.URL.String())
+	conf := warewulfconf.Get()
 
 	url := strings.Split(req.URL.Path, "?")[0]
 	path_parts := strings.Split(url, "/")
@@ -28,14 +30,19 @@ func ImagesSend(w http.ResponseWriter, req *http.Request) {
 	stage_file := path.Join(image.ImageParentDir(), image_name)
 	wwlog.Serv("stage_file '%s'", stage_file)
 
-	if util.IsFile(stage_file) {
-		err := sendFile(w, req, stage_file, "")
-		if err != nil {
-			wwlog.ErrorExc(err, "")
-			return
-		}
-	} else {
+	if !util.IsFile(stage_file) {
 		w.WriteHeader(http.StatusNotFound)
 		wwlog.Error("images: not found: %s", stage_file)
+		return
+	}
+
+	if conf.Warewulf.CacheControl != "" {
+		w.Header().Set("Cache-Control", conf.Warewulf.CacheControl)
+	}
+
+	err := sendFile(w, req, stage_file, "")
+	if err != nil {
+		wwlog.ErrorExc(err, "")
+		return
 	}
 }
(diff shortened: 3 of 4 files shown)


anderbubble:
Thanks for the PR, @posch! I've added it to the agenda for the next steering committee meeting.

mslacken:
@posch Can you please add an entry to CHANGELOG.md and run a `make fmt` afterwards? This will make the warnings go away. 
Also did you update your grub or ipxe templates? Can you share them as well?

posch:
I expected some discussion, so this is more a draft than a polished change.

> @posch Can you please add an entry to CHANGELOG.md

I've added a line to the 4.6.0 section. Is that correct?

> run a `make fmt` afterwards? This will make the warnings go away. 

Didn't know this. It fixed one place. Thanks.

> Also did you update your grub or ipxe templates? Can you share them as well?

So far, we've been experimenting with a single nginx proxy in one of our clusters, and have modified versions of /etc/warewulf/grub/grub.cfg.ww and /etc/warewulf/ipxe/dracut.ipxe.

In both cases, the modified files have hardcoded proxy addresses.

We started with grub, but due to serious boot problems (about ~2/3 of the nodes failed to complete the grub part of the boot process), we switched to ipxe+dracut, which worked nicely and reliably.

In the grub version, we use {{if eq .ClusterName "..."}} to use the proxy with only one specific cluster. In the future, we  intent to use different (nearby) proxies for each cluster. Also, we intent to use DNS names with DNS round-robin instead of hardcoded IP addresses, so we can have a larger number of proxies. The diff is:

```
--- grub.cfg.ww.orig    2024-07-05 11:06:12.954244695 +0200
+++ grub.cfg.ww 2024-07-30 15:22:41.901801744 +0200
@@ -42,7 +42,11 @@
     initramfs="${uri}&stage=initramfs"
 
     uri="http://{{.Ipaddr}}:{{.Port}}/provision/${net_default_mac}?assetkey=${assetkey}"
+    {{if eq .Cluster "clx"}}
+    container="http://10.10.10.10:9874/images/{{.ContainerName}}.img.gz"
+    {{else}}
     container="${uri}&stage=container&compress=gz"
+    {{end}}
     system="${uri}&stage=system&compress=gz"
     runtime="${uri}&stage=runtime&compress=gz"
```

In the dracut.ipxe version we modified the wwinit.container kernel command line parameter to point to the proxy:

```
--- dracut.ipxe.orig    2024-07-31 11:22:21.381660303 +0200
+++ dracut.ipxe.new     2024-08-01 20:40:44.857435733 +0200
@@ -29,7 +29,7 @@
 initrd --name initramfs ${uri}&stage=initramfs || goto reboot
 
 set dracut_net rd.neednet=1 {{range $devname, $netdev := .NetDevs}}{{if and $netdev.Hwaddr $netdev.Device}} ifname={{$netdev.Device}}:{{$netdev.Hwaddr}} {{end}}{{end}}
-set dracut_wwinit root=wwinit wwinit.container=${uri}&stage=container&compress=gz wwinit.system=${uri}&stage=system&compress=gz wwinit.runtime=${uri}&stage=runtime&compress=gz {{if ne .KernelOverride ""}}wwinit.kmods=${uri}&stage=kmods&compress=gz{{end}} init=/init
+set dracut_wwinit root=wwinit wwinit.container="http://10.10.10.10:9874/images/{{.ContainerName}}.img.gz" wwinit.system=${uri}&stage=system&compress=gz wwinit.runtime=${uri}&stage=runtime&compress=gz {{if ne .KernelOverride ""}}wwinit.kmods=${uri}&stage=kmods&compress=gz{{end}} init=/init
 
 echo Booting initramfs
 #echo Network KernelArgs: ${dracut_net}
```

anderbubble:
Discussed during 7 August 2024 TSC meeting.

- Intent is to put 1-N caching proxies in front of this
- griznog: historically we haven’t considered containers private, so this should be fine
- anderbubble: if we do this, we should have the per-node URLs 302 (temporary) redirect to the consistent URL
- Recommendation to explicitly document that container images should not contain intra-cluster private data

posch:
I have a few more ideas I'd like to add here.

1. To have some control over caching, warewulfd should add a Cache-Control response header (e.g. max-age=60). This should be a config file option.
2. In addition to the containers, one may also want to cache kernel and initramfs. This conflicts with per-node kernel overrides, so this shouldn't be the default. I'd change the URL scheme to:
- /container-by-name/$container/kernel
- /container-by-name/$container/initramfs
- /container-by-name/$container/image
- /container-by-name/$container/image.gz

anderbubble:
We discussed this in the TSC meeting, and I'm thinking a bit more about this now. I _think_ this is the current state of the discussion, and the follow-on requirements:

- `http://.../container-by-name/{{.ContainerName}}.img[.gz]` is fine, though I'd probably use `http://.../container/{{.ContainerName}}.img[.gz]` myself.
- `http://.../provision/{{.Hwaddr}}?...&stage=container[&compress=gz]` needs to be retained, though it could be replaced with an HTTP redirect. (I'm currently thinking that 303 / See Other might be the best choice; but let me know if you think otherwise.)

I was thinking that the second / current URL would need to continue in perpetuity; but since both iPXE and Grub configurations are generated by the Warewulf server in the first place, and can be templated, it should actually be ok to point directly to the desired container image. We should retain the `/provision/` url for bacwards-compatibility, at least for some deprecation period; but replacing it with a redirect would allow the existing URL to benefit from caching.

Speaking of which: we'll also need a way to specify the proxy address separately from the Warewulf server address. I propose that it should be a node field, so that it could be specified in a profile or individual node. That way different nodes in the cluster could work from different proxies. (It might be better to find a name for it that doesn't explicitly claim that it's a proxy, but just an alternative Warewulf server address. That way we could possibly re-use it for multiple Warewulf servers in the future.)

Let me know what you think. I'll pass this back to you for now, at least.

anderbubble (commented):
We discussed this in the TSC meeting, and I'm thinking a bit more about this now. I think this is the current state of the discussion, and the follow-on requirements:

- http://.../container-by-name/{{.ContainerName}}.img[.gz] is fine, though I'd probably use http://.../container/{{.ContainerName}}.img[.gz] myself.
- http://.../provision/{{.Hwaddr}}?...&stage=container[&compress=gz] needs to be retained, though it could be replaced with an HTTP redirect. (I'm currently thinking that 303 / See Other might be the best choice; but let me know if you think otherwise.)

I was thinking that the second / current URL would need to continue in perpetuity; but since both iPXE and Grub configurations are generated by the Warewulf server in the first place, and can be templated, it should actually be ok to point directly to the desired container image. We should retain the /provision/ url for bacwards-compatibility, at least for some deprecation period; but replacing it with a redirect would allow the existing URL to benefit from caching.

Speaking of which: we'll also need a way to specify the proxy address separately from the Warewulf server address. I propose that it should be a node field, so that it could be specified in a profile or individual node. That way different nodes in the cluster could work from different proxies. (It might be better to find a name for it that doesn't explicitly claim that it's a proxy, but just an alternative Warewulf server address. That way we could possibly re-use it for multiple Warewulf servers in the future.)

Let me know what you think. I'll pass this back to you for now, at least.

posch:
>     * http://.../container-by-name/{{.ContainerName}}.img[.gz] is fine, though I'd probably use http://.../container/{{.ContainerName}}.img[.gz] myself.

Unfortunately, /container/ is already taken, see internal/pkg/warewulfd/warewulfd.go:
```
        wwHandler.HandleFunc("/provision/", ProvisionSend)
        wwHandler.HandleFunc("/ipxe/", ProvisionSend)
        wwHandler.HandleFunc("/efiboot/", ProvisionSend)
        wwHandler.HandleFunc("/kernel/", ProvisionSend)
        wwHandler.HandleFunc("/kmods/", ProvisionSend)
        wwHandler.HandleFunc("/container/", ProvisionSend)
        wwHandler.HandleFunc("/overlay-system/", ProvisionSend)
        wwHandler.HandleFunc("/overlay-runtime/", ProvisionSend)
        wwHandler.HandleFunc("/status", StatusSend)
```

I haven't checked if it's actually used. I would be fine with redefining /container/.

griznog:
Since we're moving away from the `container` nomenclature anyway, `/images` or `/node-images` fits with that.

anderbubble:

posch:
@anderbubble, @griznog I've rebased my patches. The URL changed to /images/ (was: container-by-name). And I've added the Cache-Control option.

Sorry for being so lazy. Seems I missed the optimal time to merge this. I'd still like to see this or something similar in warewulf. We're using this approach successfully in production with >800 nodes. We're not using the most recent warewulf, though, so this patch isn't well tested.

mslacken:
Thanks for your continuous work on that @posch.
Just a dumb question from my side. What does CacheControl do exactly?
As far as I see it only updates the header and what other effects does it have?

anderbubble:
I know this has been hanging out for a while, but I like the idea of getting this in for v5. We could have `/images/<name>` be the canonical location for each image, and then have the node URLs redirect to that.

Moving to that milestone.

PR #1365, opened 2024-08-20
add cow option for binds

mslacken:
The option cow can now be set for files which are then
mounted during exec, but copied into the image and removed
if not modified.

Example:
```
wwctl container shell leap15.6 --bind /root/test:/root/test:cow
```

### DIFF
--- a/internal/app/wwctl/container/exec/main.go
+++ b/internal/app/wwctl/container/exec/main.go
@@ -72,28 +72,18 @@ func runContainedCmd(cmd *cobra.Command, containerName string, args []string) (e
 	// so do this here.
 	// At first read out conf, the parse commandline, as copy files has the
 	// same synatx as mount points
-	mountPts := conf.MountsContainer
-	mountPts = append(container.InitMountPnts(binds), mountPts...)
-	filesToCpy := getCopyFiles(containerName, mountPts)
-	for i, cpyFile := range filesToCpy {
-		if err = util.CopyFile(cpyFile.Src, path.Join(container.RootFsDir(containerName), cpyFile.FileName)); err != nil {
-			return fmt.Errorf("couldn't copy files into container: %w", err)
+	mountPts := append(container.InitMountPnts(binds), conf.MountsContainer...)
+	filesToCpy := getCopyFiles(mountPts)
+	for _, cpyFile := range filesToCpy {
+		if err := (cpyFile).copyToContainer(containerName); err != nil {
+			wwlog.Warn("couldn't copy file: %s", err)
 		}
-		// we can ignore error as the file was copied
-		stat, _ := os.Stat(path.Join(container.RootFsDir(containerName), cpyFile.FileName))
-		filesToCpy[i].ModTime = stat.ModTime()
 	}
 	wwlog.Verbose("Running contained command: %s", childArgs)
 	retVal := childCommandFunc(cmd, childArgs)
 	for _, cpyFile := range filesToCpy {
-		if modStat, err := os.Stat(path.Join(container.RootFsDir(containerName), cpyFile.FileName)); err != nil {
-			wwlog.Info("copied file was removed: %s", err)
-		} else {
-			if modStat.ModTime() == cpyFile.ModTime {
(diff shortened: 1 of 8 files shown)


anderbubble:
Rebased to main and altered changelog entry.

anderbubble:
I've made some adjustments to help me follow the code and verify its behavior. I'll submit those as additional commits here in a minute; but with those in place, it appears to work reasonably as expected:

```
[root@rocky ~]# echo "Hello, world." >testfile

[root@rocky ~]# wwctl container exec rocky-9.3 --bind /root/testfile:/root/testfile:cow -- -- /bin/bash -c 'cat /root/testfile'
[...]
Hello, world.
[...]

[root@rocky ~]# wwctl container exec rocky-9.3 -- -- /bin/bash -c 'cat /root/testfile'
[...]
cat: /root/testfile: No such file or directory
ERROR  : Failed executing container command: exit status 1

[root@rocky ~]# wwctl container exec rocky-9.3 --bind /root/testfile:/root/testfile:cow -- -- /bin/bash -c 'echo "Goodbye" >>/root/testfile'
[...]
WARN   : couldn't remove file: file modified since copy: /root/testfile
[...]
WARN   : couldn't copy file: file already exists in container: /root/testfile
[...]
WARN   : couldn't remove file: file not previously copied: /root/testfile
[...]

[root@rocky ~]# wwctl container exec rocky-9.3 -- -- /bin/bash -c 'cat /root/testfile'
[....]
Hello, world.
Goodbye
[...]
```

Note that, since `container exec` runs two different chrooted commands (one for the command itself, another for `container_exit.sh`) the warning messages and copy behavior get confusing if you aren't really paying attention. This is exacerbated by the output from exec having grown a bit verbose, so expect a PR from me to silence that soon.

I do think that "cow" should just be called "copy"; so I'll intend to include that in my PR updates; but I'll keep them as new commits so we can back them out if you disagree. (It copies the file regardless; so I don't think it's correct to call it "copy-on-write.")

anderbubble:
I noticed that this doesn't work for directories.

```
[root@rocky ~]# wwctl container exec rocky-9.3 --bind /mnt:/mnt/outer:copy -- -- /bin/bash -c 'echo "Goodbye" >>/root/testfile'
ERROR  : File copy from /mnt to /var/lib/warewulf/chroots/rocky-9.3/rootfs/mnt/outer failed.
 write /var/lib/warewulf/chroots/rocky-9.3/rootfs/mnt/outer: copy_file_range: is a directory
WARN   : couldn't copy file: couldn't copy files into container: write /var/lib/warewulf/chroots/rocky-9.3/rootfs/mnt/outer: copy_file_range: is a directory
```

So that's room for potential future improvement; but I don't think it needs to be a blocker here.

anderbubble:
@mslacken I've pushed my recommended updates. Let me know what you think.

anderbubble (approved):

internal/app/wwctl/container/exec/main.go
```diff
@@ -154,3 +175,56 @@ func SetBinds(myBinds []string) {
...
+func (this *copyFile) removeFromContainer(containerName string) error {
+	containerDest := path.Join(container.RootFsDir(containerName), this.fileName)
+	if this.modTime.IsZero() {
+		return fmt.Errorf("not previously copied: %s", this.fileName)
+	} else if destStat, err := os.Stat(containerDest); err != nil {
+		return fmt.Errorf("no longer present: %s", err)
+	} else if destStat.ModTime() == this.modTime {
+		return os.Remove(containerDest)
+	} else {
+		return fmt.Errorf("modified since copy: %s", this.fileName)
```
mslacken, line 213: This isn't an error. So we should inform the user about that the file was modified, but keeping the modified file in the node image is the intended purpose.

anderbubble: > This isn't an error.

I think I agree with you; this is just my Python brain over-identifying golang errors with exceptions, which they aren't really.

I'll move the more expected parts of these out of Remove.

anderbubble:
@mslacken updated based on your comments. Let me know what you think.

PR #1367, opened 2024-08-21
Fix old links that pointed to warewulf-web

griznog:
- **Replace warewulf-web with warewulf in contributing links.**
- **Update CHANGELOG for PR.**


mslacken (approved):
Thx

PR #1369, opened 2024-08-24
docs: fix dracut build command in docs

epruesse:
The current command line works correctly only if one kernel is installed. It picks the oldest if there are multiple, which is rather unlikely to be correct. Using "tail" instead of "head" would work also, but `--regenerate-all` makes it briefer and safer.


anderbubble (approved):
I love it! Thanks for contributing.

PR #1384, opened 2024-09-09
test overlay rendering

mslacken:
Test the rendering of ther overlay dirtectly besides the overlay, not in one huge test


griznog:
Could the tests for overlays be in `overlays/OVERLAYNAME/tests` rather than `rootfs`? I was thinking about this for sumbitting an overlay PR for a new overlay and it seems like it'd be cleaner if the test and other metadata were kept outside of `rootfs`. 

Part of the reason for this suggestion is that I was thinking about how I might test standalone/contributed overlays and thought it might be nice if the overlay testing was something that worked during the build for checking overlays distributed with Warewulf but also would work post-install to check any overlay I happen to be working on.

mslacken:
Sure, this is more POC how it could look like.  
My original ides was, that you just put a file e.g. with `rendered` as suffix instead of `ww` besides the template, but while trying to implement it, I realized that we need e.g more information for rendering, like the `nodes.conf` *AND* `warewulf.conf`, so using the go testing syntax seemed to be the best solution.

griznog:
I was thinking about how to best get the configuration data in there but keep it separate and thought it'd be nice if the tool (`wwctl overlay test OVERLAYNAME` ??) might just always look in `overlays/OVERLAYNAME/tests` for a `warewulf.conf` and `nodes.conf` which would let every overlay define and include it's own test configuration.

griznog:
Actually I'll amend that. Since testing overlays means pulling in other arbitrary files via `{{ Include ... }}`, the configs could be in `overlays/OVERLAYNAME/tests/rootfs` so that the tests have a complete `/` heirarchy under there in which to place configs and test files.

anderbubble:
Now that I understand a bit more what's going on here, I like this _a lot_. I will definitely incorporate this into #1306, and I agree with @griznog about putting it in `overlays/$name/tests`, rather than `rootfs`. Should work just the same, and avoid distributing the tests with the overlay. :)

griznog:
I've been rolling this around in my head while working on some overlays lately and I'm thinking this could be really useful not just during building warewulf, but when working on overlays in general. Something like 

```
wwctl overlay test OVERLAYNAME 
```

And that would look in tests where it would find some kind of test description and all the files it needs in test/rootfs including a warewulf.conf and nodes.conf for the tests. I haven't been looking at the testing stuff, so I'm flying blind here. I just catch myself all the time waiting on all my overlays to rebuild only to discover I didn't close a () or {{ }} or typoed something. `wwctl overlay show -r ...` kind of works for this, but can't really be automated against the live nodes.conf. Some kind of utility like this which makes it easy and fast to work on tests would encourage me and maybe others to submit back more tests to improve the build process as well. 

The context here being my agenda to make overlays a full fledged config managemnent solution and having it be this easy to build/run tests as you develop an overlay on your live cluster plays into that really well.

anderbubble:
Closing now that #1306 is merged.

PR #1394, opened 2024-09-16
Removed dead container code

mslacken:
As from slack:
These files seem very similar. Do they have different purposes? It looks like they had the same contents and then maybe were altered slightly. But they have some of the same comments and a lot of the same code, albeit rearranged. https://github.com/warewulf/warewulf/blob/main/internal/pkg/api/container/container.go
https://github.com/warewulf/warewulf/blob/main/internal/pkg/api/apiconfig/container/container.go
container.go
￼warewulf/warewulf | Added by GitHub
container.go
￼warewulf/warewulf | Added by GitHub

Code is dead, so remove it


anderbubble (changes requested):
@mslacken I think you didn't mean to base this change on the "better template rendering" commit.

anderbubble:
Closing in favor of #1394.

PR #1411, opened 2024-09-21
InfoErr function for wwctl overlay show -r

griznog:
- **Add INFOERR to wwlog**
- **Use InofErr to send info to stderr**
- **Updated CHANGELOG.md.**

It's useful to be able to capture the output of a rendered template for testing
and debugging, for instance when a template produces a script that has no other
dependencies and can be easily tested without needing to build/deploy the
entire overlay or if an overlay is being used to manage host configs outside of
the exiting `host` overlay framework. 

This commit adds and InfoErr level which allows sending INFO level output to
stderr instead of stdout. This function is then used in the `overlay show -r`
command to send the informational output to stderr allowing the rendered
template to easily be captured by simply redirection the output to a file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added command line parameters for credentials of a container registry
 - Add flag `--build` to `wwctl container copy`. #1378
 - Add `wwctl clean` to remove OCI cache and overlays from deleted nodes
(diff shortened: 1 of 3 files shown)


anderbubble:
Rebased on main, squashed, and signed-off.

anderbubble:
@griznog you can already suppress those log messages with `--quiet`. With that in mind, I'm closing this for now; but let me know if you still need this for a reason I haven't considered.

PR #1459, opened 2024-10-21
Make cache directory configureable and add clean command

mslacken:
- **add Cachedir and point OCIBlobCache to it**
- **added `wwctl clean` to remove cache/dead overlays**

@anderbubble The clean command can/should be expanded. Atm it was the easiest way to delete the cache nobody knew about. Stray overlays I just knew that this might be a problem.


mslacken:
@JasonYangShadow @anderbubble The build for go 1.20 is failing as I use the experimental slices package, which is part of go1.21. The blank import works for go 1.2[1-2] but how can we handle this is in go.mod?

JasonYangShadow:
> @JasonYangShadow @anderbubble The build for go 1.20 is failing as I use the experimental slices package, which is part of go1.21. The blank import works for go 1.2[1-2] but how can we handle this is in go.mod?

we have a dependabot created PR 
https://github.com/warewulf/warewulf/pull/1316
requiring the upgrade of the go version to 1.21.

mslacken:
Switched now back to util.InSlice which is part of warewulf, so no need to update to go1.20

anderbubble:
@mslacken I pushed a commit with some recommended refactors. Let me know what you think.

mslacken:
Looks good. You will have to approve and merge it.

anderbubble (approved):

PR #1568, opened 2024-11-26
Introduce Resource concept

mslacken:
For storing cluster wide resource the `Resource` attribute is introduced.
Resources can be managed via the new `wwctl resource` command in which they
can be added, modified and removed.
On node/profile level resources can be added and removed from a list. If a resouce
is added to this it's considered as enabled.
As example the `NFSHOME` is added and looks in `nodes.conf` like
```
resource:
  NFSHOME:
    mountpoint: /home
    moptions: defaults
    eoptions: rw,rootsquash
    epath: /home
    server: controller
```
The templates `fstab.ww` and `exports.ww` will check if the three first
letters of a (enabled) resource are equal 'NFS' and then create the acording
entries.


griznog:
This can already be accomplished in Tags with the added benefit that it's easy to use profiles for common settings with individual nodes being able to override or extend that. There's actually a lot of stuff in `warewulf.conf` that could be done with Tags and exploit the whole `nodes.conf` and overlay ecosystem. 

Sort of related, dropping the idea of host overlays and just letting the master be a defined node opens up the same methods to use for managing all the master node(s) settings. I don't see why `warewulf.conf` couldn't just be a `nodeprofile` with a special reserved name which the `node` configs for the master node use.

mslacken:
@griznog thanks for your thoughts.
I am also not sure if this is the best solution. For the profile support speaks that nothing changes.
The problem here is that profiles are always are meant to be assigned to a node or node group. Also users may be tempted not just to put in a single resource in a profile but also network and disk configuration which could lead to complex setups.
Regarding the global values in `warewulf.conf`. We need to get more out of them, but still I would say things like `systemd name` should stay as these are very site specific.
I would also like to treat the warewulf controller as node within `nodes.conf`, but for that we need another delivery mechanism for the templates, as `wwclient` simply dumps the cpio archive on the disk. Here we would need some more clever mechanism, like evaluating the templates with wwclient on the fly and the writing the files to disk. This would also be  a benefit on the nodes, so that we could then have e.g. the possibility of template executions, something like `wwrun`.

anderbubble:
I would  be very interested to try to use this as a data source for ignition, replacing the configuration currently built-in to nodes+profiles; but I've got some investigating to do to truly make that argument.

anderbubble:
Rebased to main.

anderbubble (changes requested):
I've fixed the failing tests and whatnot. I also renamed a few things.

- node.Node.ResourceList -> Resources
- node.NodesYaml.Resource -> Resources
- node.RemoteRes -> node.Resource

From a design standpoint, I'd want to change a few things:

- I think resources should just be defined on nodes and profiles, rather than as separate entities and then linked in. I at least want to try it to see how I feel about it.
- I think resources should just be an `interface{}` rather than `map[string]string`. Maybe `map[string]interface{}`. That way we could do arbitrary nesting like what @griznog is after. Again, I at least want to try it.
- I do not think that the server's NFS configuration (or any server configuration at all) should use resources. NFS exports should stay in warewulf.conf, and cluster node mounts should go in nodes.conf

edit: we also need to add documentation

anderbubble:
@mslacken I've refactored this with enhancements:

- Make Resources an `interface{}` to support arbitrary yaml
- Remove `wwctl resource` as incompatible with arbitrary yaml
- Revert changes to host overlay templates
- Remove NFS mount options from warewulf.conf
- Replace NFS support / resource prefix with "fstab" resource
- Move resources to profiles
- Migrate warewulf.conf mounts to nodes.conf with `wwctl upgrade`
- Add documentation

I'm pretty happy with it, and I think @griznog will be, too. For example, the fstab overlay can now be configured to mount arbitrary file systems:

```
nodeprofiles:
  default:
    resources:
      fstab:
        - spec: warewulf:/home
          file: /home
          vfstype: nfs
        - spec: warewulf:/opt
          file: /opt
          vfstype: nfs
```

I'll mark this approved from my side; but look it over and let me know what you think.

anderbubble (approved):

mslacken:
@anderbubble  Great work, thanks.
I merging this one, although I see rendering errors in `wwctl node list` like shown below
```
n01   Resources[fstab]                                           --          [map[file:/home freq:%!s(int=0) mntops:defaults passno:%!s(int=0) spec:warewulf:/home vfstype:nfs] map[file:/opt freq:%!s(int=0) mntops:defaults,ro passno:%!s(int=0) spec:warewulf:/opt vfstype:nfs]]
```
but this is most likely another issue as I also get
```
n01   Disks[/dev/vda].Partitions[rootfs].Resize                  --          %!s(bool=false)
```

anderbubble:
@mslacken I addressed the formatting issues you discovered at https://github.com/warewulf/warewulf/pull/1647. Thanks for noting them!

PR #1569, opened 2024-11-26
Remove the node.ConfigFile initialization

mslacken:
The location for `nodes.conf` can be given a command line parameter, 
but this can fail as some functions (node.Persit()) call node.ConfigFile
which is set by init() what was called *before* the command line parameter 
is read. 
So remove node.ConfigFile completely

### DIFF
--- a/internal/pkg/node/constructors.go
+++ b/internal/pkg/node/constructors.go
@@ -14,13 +14,29 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const NodesConfPath = "warewulf/nodes.conf"
+
+/*
+Get the actual path of nodes.conf, with the possibility to prefix
+*/
+func GetNodesConf(prefix ...string) string {
+	var myPath []string
+	if len(prefix) == 0 {
+		controller := warewulfconf.Get()
+		myPath = append(myPath, controller.Paths.Sysconfdir)
+	} else {
+		myPath = append(myPath, prefix...)
+	}
+	myPath = append(myPath, NodesConfPath)
+	return path.Join(myPath...)
+}
+
 /*
 Creates a new nodeDb object from the on-disk configuration
 */
 func New() (NodesYaml, error) {
-	controller := warewulfconf.Get()
-	wwlog.Verbose("Opening node configuration file: %s", path.Join(controller.Paths.Sysconfdir, "warewulf/nodes.conf"))
-	data, err := os.ReadFile(path.Join(controller.Paths.Sysconfdir, "warewulf/nodes.conf"))
+	wwlog.Verbose("Opening node configuration file: %s", GetNodesConf())
+	data, err := os.ReadFile(GetNodesConf())
 	if err != nil {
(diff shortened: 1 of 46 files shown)


mslacken:
@anderbubble The failing dead code tests don't have anything to do with the added code and need an update.

anderbubble:
Rebased to main.

internal/pkg/node/constructors.go
```diff
@@ -14,23 +14,13 @@ import (
...
-	ConfigFile string
-)
-
-func init() {
-	conf := warewulfconf.Get()
-	if ConfigFile == "" {
-		ConfigFile = path.Join(conf.Paths.Sysconfdir, "warewulf/nodes.conf")
-	}
-}
-
```
anderbubble, line 27: Removing ConfigFile from node means that this path ends up getting copied several times throughout the code. I understand the original ordering concern; but I don't think duplicating this code in a bunch of other places is the answer.

anderbubble (changes requested):

internal/pkg/node/constructors.go
mslacken: In principle you are right *but* the `nodes.conf` should only be read via `node.Get()` and written via `node.Persist()` so this code is just copied two times and not several times.

anderbubble: The path to nodes.conf is duplicated at

- https://github.com/warewulf/warewulf/pull/1569/files#diff-18635393846b27ad0a0b42766e25f0ee688658e3806278b477fc312a26e30530R34
- https://github.com/warewulf/warewulf/pull/1569/files#diff-18635393846b27ad0a0b42766e25f0ee688658e3806278b477fc312a26e30530R35
- https://github.com/warewulf/warewulf/pull/1569/files#diff-cbf7980ed402d328dd25a662ff88774f8d6629f9b89d70e5514df7d5033c8249R36
- https://github.com/warewulf/warewulf/pull/1569/files#diff-2c728dc8f5f1c5947be37b4ccddf31e5655624904fa2aa9001abb3abe53c31feR22
- https://github.com/warewulf/warewulf/pull/1569/files#diff-2c728dc8f5f1c5947be37b4ccddf31e5655624904fa2aa9001abb3abe53c31feR23
- https://github.com/warewulf/warewulf/pull/1569/files#diff-e1aef9e67db5d9c8dac8ef8b0005e650dfecd42740f20fff784721f14b3fc08eR114
- https://github.com/warewulf/warewulf/pull/1569/files#diff-e17cee5fa398c4a024a96998083c6b139e45c35d6f892e4896fbb86828c442e5R56

mslacken:
Access nodes.conf now with `getNodesConf` wich has the possibility to prefix for tests.

anderbubble:
Please see #1595 for a possible alternative.

mslacken:
close #1595

mslacken:
Revoking in favor of #1595

PR #1572, opened 2024-11-30
Enable nested profiles

griznog:
- **Move Profiles from Node struct to Profile struct.**
- **Enable/handle nested profiles.**
- **Add test for merged profile tags.**
- **Update CHANGELOG.md**

fixing or enhancing.

- Fixes #

## Reviewer  checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Added Netplan NIC support for Debian/Ubuntu #1463
+- Add handling/merging of nested profiles
 - Added documentation on ensuring `systemctl restart warewulfd` is ran when editing `nodes.conf` or `warewulf.conf`
 - Add the ability to boot nodes with `wwid=[interface]`, which replaces
   `interface` with the interface MAC address
--- a/internal/pkg/node/constructors.go
+++ b/internal/pkg/node/constructors.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path"
 	"sort"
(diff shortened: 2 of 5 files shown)


griznog:
When I `make test` there are a lot of failing tests, some of which look like they might be related to moving `Profiles` from the `Node` struct to the `Profile` struct, but a whole lot more that don't seem to be related. I must shamefully admit that today is the first time I've ran `make test` locally so I don't know how normal the failed tests are or if they are just something I have broken locally on my system. 

@anderbubble let me know if my test is ok, I think I want to add some more checks to it around tags/values in netdev and ipmi, but wanted to start simple.

mslacken:
Using the profile name as sorting order sound arbitrary and I don't like the idea. Still it might be the best solution as
# users can change the order by renaming the profile, like in `init.d`
# its predictable all the time
@anderbubble what's your take?

griznog:
@mslacken I played a little with the idea of building a graph and merging profiles in the order they are specified/included, but eventually settled on the approach of flattening the list of profiles and sorting as a compromise that makes it easy to understand how the merging is done and easier to control the order in the case someone wants to do that. 

In my local WW foo I've found that both overlays and profiles work much better and much more understandably if I stick to the best practice of never setting the same value in multiple profiles and never letting two included overlays touch/modify the same file. The further I go down the profile/overlay rabbit hole the more I lean toward detecting either of those things happening and throwing an error or at least a warning when it does happen as it, at least in my experience, always leads to confusion and regret when it happens.

griznog:
This still needs some work. Tags merge correctly but overlays are not appending correctly, presumably `mergo` is not doing what I think it is doing.

PR #1588, opened 2024-12-10
OpenAPI REST API

anderbubble:

A simpler OpenAPI v3 compatible REST API for Warewulf.

## Fixes

- Fixes: #1801

## Reviewer  checklist


internal/pkg/warewulfd/api.go
```diff
@@ -0,0 +1,106 @@
...
+	api.Get("/api/nodes", apiGetNodes())
+	api.Get("/api/nodes/{id}", apiGetNodeByID())
+	api.Get("/api/profiles", apiGetProfiles())
+	api.Get("/api/profiles/{id}", apiGetProfileByID())
+
+	api.Docs("/api/docs", swgui.New)
+
+	return api
+}
+
```
mslacken, line 37: So this will only output the unmerged nodes. I strongly disagree deliver the nodes like this in the API, as this is just a complicated way to transform `nodes.conf` to yaml. But this is just a draft, so please provide the merged nodes.

mslacken (commented):
Can you please base this draft on #815 as both touch the warewulfd server code?

internal/pkg/warewulfd/api.go
anderbubble: Yep, definitely just a draft. I'm not looking for a review at this time; I'm pushing it so people can see, but I'm expecting heavy iteration while I prototype.

The reason I started with the unmerged nodes is because that's what people would _push_. I currently intend to switch the default view to the merged node and then support a separate `/api/raw-nodes/` or similar; but I'm still playing with it in my head.

anderbubble:
> Can you please base this draft on #815 as both touch the warewulfd server code?

Not until it's merged; but I'll try to prioritize a review. Otherwise, any rebasing is on me.

internal/pkg/warewulfd/api.go
mslacken: The first thing I would like to see in an API would be the possibility to push a status string to `warewulfd` which then is displayed with `wwctl status`.

internal/pkg/overlay/overlay.go
```diff
@@ -505,3 +505,7 @@ func OverlayGetFiles(name string) (files []string, err error) {
 	})
 	return
 }
+
+func Exists(name string) bool {
+	return util.IsDir(OverlaySourceDir(name))
```
JasonYangShadow, line 510: OverlaySourceDir is undefined, guessing not all changes are pushed.

anderbubble: @JasonYangShadow this was actually a problem of rebasing on main and not fully accommodating changes to overlays from upstream. Should be fixed now. Thanks for pointing it out!

anderbubble:
I think I've got this ready now, and it's all rebased on v4.6.0. Check it out, and let me know what you think.

The API is disabled, and not even configured to be able to respond, unless `warewulf.conf:api.enabled: true`

Docs could maybe use a bit more detail, but there's a skeleton in there, at least.

griznog (approved):

jcsiadal (approved):
I do have questions, but don't want to hold up the commit--I can save those for the next meeting. I walked all of the changes and I don't see any issues in code. It looks great. I have not had an opportunity to live-test this yet.
Future update: we should probably create a common post-run function for the node/image/etc calls instead of duplicating the same code everywhere.

mslacken (changes requested):
Looks quite good, but please don't do a functional unnecessary spree rename of `warewuld` to `daemon` just because you can, because
* this adds a lot of noise to this PR
* makes the code more unreadable, as daemon restart could mean  restart `wwclient`, `warewulfd` or any other daemon we use, so even `dhcpd`.

Remove the hardcoded admin hash in `auth.conf`, this isn't secure by default.

I would consider to use the password of `root` via e.g. https://pkg.go.dev/github.com/msteinert/pam as `warewulfd` runs as `root` and can change anything so using the `root` password is right here. Also the `root` password is part of security policies and just another configuration file for password is not.

anderbubble:
i have removed the packed-in `admin` password hash, and replaced it with an empty hash (which won't equal anything) and a comment that reiterates how to generate a hash.

The admin hash is still in userdocs; but it's specifically called out as such as an example, next to instructions for how to generate a hash from a password.

I am less convinced about inverting the warewulfd package refactor; but now that I understand what you're saying, I'll at least give it another try.

anderbubble:
@mslacken I think I have addressed your feedback successfully, and I begrudgingly agree with it.

The licenses check is failing because of an external DNS issue; so I think we should ignore that.

internal/pkg/api/image/image.go
```diff
@@ -163,7 +163,7 @@ func ImageImport(cip *wwapiv1.ImageImportParameter) (imageName string, err error
 	} else if strings.HasPrefix(cip.Source, "docker://") || strings.HasPrefix(cip.Source, "docker-daemon://") ||
 		strings.HasPrefix(cip.Source, "file://") || util.IsFile(cip.Source) {
 		var sCtx *types.SystemContext
-		sCtx, err = getSystemContext(cip.OciNoHttps, cip.OciUsername, cip.OciPassword, cip.Platform)
+		sCtx, err = GetSystemContext(cip.OciNoHttps, cip.OciUsername, cip.OciPassword, cip.Platform)
 		if err != nil {
```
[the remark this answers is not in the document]
anderbubble: It's exported so that it can be used by `internal/pkg/warewulfd/api/image.go:importImage`. I am not using `internal/pkg/api/image/image.go:ImageImport` because it is encumbered by grpc semantics, which is being removed (though that removal is not complete here).

mslacken (changes requested):
I am still not happy with a separate `auth.conf` So why not move that three lines to `warewulf.conf` ?
Can we also level the `wwctl node status` api call to the new API?

MiddelkoopT:
> I am still not happy with a separate `auth.conf` So why not move that three lines to `warewulf.conf` ?

I believe a separate `auth.conf` would allow folks to store `warewulf.conf` in git without storing sensitive information in git (I assume the credentials are sensitive - I have not been following this PR closely).  Having all config information in git was my teams standard process and how we dealt with change management and DR.

anderbubble:
@mslacken as discussed in TSC, I've updated this with a new `warewulf.conf:api:allowed subnets` configuration variable that defaults to localhost. I also added additional documentation highlighting the risks and mitigations associated with API access.

mslacken (approved):
Looks good, just following to points:
1. /api/docs is accessible even if the requesting host in **not** in allowed networks, this is just odd
2. I get following error ` %!w(*errors.errorString=&{crypto/bcrypt: hashedSecret too short to be a bcrypted password}` when creating a password with `mkpasswd --method=bcrypt`

I can merge, if you think this OK.

anderbubble:
> /api/docs is accessible even if the requesting host in not in allowed networks, this is just odd

`/api/docs` doesn't go through the authentication handler. So while this may be surprising, I think it's benign.

> hashedSecret too short to be a bcrypted password

I'll check that out before I merge.

Thanks for the approval!

anderbubble:
> I'll check that out before I merge.

I think you must've mishandled the string; I just generated a new password with `mkpasswd --method=bcrypt` and it was fine. (Either that or it's a rocky vs suse issue; but that seems unlikely.)

I did fix the message formatting, though.

```
[Thu Apr  3 07:11:12 MDT 2025] WARN   : crypto/bcrypt: hashedSecret too short to be a bcrypted password
```

PR #1595, opened 2024-12-13
Configure nodes.conf dynamically from config

anderbubble:
## Description of the Pull Request (PR):

Removes the use of init() to initialize the path to `nodes.conf`, replacing it with a dynamic lookup from config.

- Closes: #1596 
- See also: #1569

## Reviewer  checklist

The reviewer checks the following items before merging the PR.

- [ ] The PR is based on the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))
- [ ] All commits are "Signed off" (e.g., using `git commit --signoff`) in agreement to the [DCO](DCO.txt)
- [ ] The [CHANGELOG](https://github.com/warewulf/warewulf/blob/main/CHANGELOG.md) has been updated, if necessary, and under the correct release heading
- [ ] The [userdocs](https://github.com/warewulf/warewulf/tree/main/userdocs) have been updated, if necessary
- [ ] The submitter is listed in the [contributors file](https://github.com/warewulf/warewulf/blob/main/CONTRIBUTORS.md)
- [ ] The test suite has been updated, if necessary

PR #1598, opened 2024-12-18
Support nested profiles

anderbubble:

This PR implements nested profiles. Nested profiles are applied from left to right and depth first.

- Replaces: #1572

Also removes DEBUG log level configuration from test suite.

## Reviewer  checklist


anderbubble:
@griznog I haven't deployed this and tested it interactively yet; but the refactor of node merging ended up identifying bugs in the test suite; so I'm feeling pretty good about it so far.

I'll fix the formatting errors and add documentation; but let me know what you think.

griznog:
Installed this on my local test setup and it looks like the profile merging works for tags and overlays. 

`wwctl node list [-a | --json]` doesn't show the list of profiles though, only the profile assigned to the node in its unmerged config. That made me realize a problem in `wwctl * list` generally, there's no clear distinction between showing the merged node config result and the raw node config before merging. `wwctl profile list -a` shows the raw profile and `wwctl node list -a` shows the merged node config. It might be nice to have flags for these, `--merged` and `--raw`, with the default being consistent for both nodes and profiles?

While I like the left to right, depth first merging, one of the reasons I avoided doing that and fell back to alphanumerical order was that alphanumerical order is easy to display. If we are going to merge in the more complex tree struture it'd be nice if the `Profiles` field when displaying the merged node config (`-a`) shows those as ordered nested lists which reflect the graph of the merge order. 

If were going to use ordered lists/trees for the profile merging then it'd also be nice if the `wwctl node set --profile*` options allowed append/prepend/insert/delete operations. That might be getting into `profile` subcommand territory, like

`wwctl node profile [show|show-merged|append|prepend|insert|delete] PROFILE NODERANGE`

Having re-read what I just wrote I'm returning full circle to the original idea I wanted which was for profiles to work like class inheritance with nodes being (logically) nothing more than instances of a profile, which itself is a child of some parent profiles, ultimately making warewulf an infrastructure "compiler" with a YAML based grammar for `nodes.conf` ... which then brings me back to "I'm really busy and that's too hard, `sort` FTW" :grin:

mslacken:
@anderbubble 
Please rebase on main and remove commit [7a23c3e](https://github.com/warewulf/warewulf/pull/1598/commits/7a23c3e1159747acb3c4324d6d7ac683879d3ba0)

anderbubble:
> @anderbubble Please rebase on main and remove commit [7a23c3e](https://github.com/warewulf/warewulf/pull/1598/commits/7a23c3e1159747acb3c4324d6d7ac683879d3ba0)

I'll try. I recall there was an issue that motivated me pulling that commit in; but I forget what it was now.

If I have trouble, I'll wait until #1595 is merged, and then rebase on that.

I definitely agree with the request to remove that before merge, though: thanks for noticing it in your review.

anderbubble:
@griznog since you weren't in the TSC meeting today, I at least wanted to respond to your comments here:

> wwctl node list [-a | --json] doesn't show the list of profiles

This is actually explicitly coded right now, because I couldn't come up with a better thing to display. In theory I could display the list of profiles that actually affect the node after negations; but then that might be confusing because people might think the value of those profiles will be crawled and re-applied _again_. To be more like how we display overlays, it'd be good if it showed the full list of included / excluded profiles; but I'm not actually tracking negated profiles after they're applied right now. (If this was definitely what we wanted, I could probably make it work, though.)

It's also unclear to me what should be listed in the "source" column for a merged profile field.

Mostly, just showing the original Profiles list seemed fine and clear for now; but if you have something else you'd want it to display there, _please_ say so in a future TSC meeting so we can talk through it.

> there's no clear distinction between showing the merged node config result and the raw node config before merging

In theory this is highliged by the `-a` output showing the source of each field. For now, the consensus in the TSC meeting today was that wwctl should not output unmerged nodes, and that the unmerged content should be accessed with `wwctl node edit` or by inspecting the yaml directly. But, again, if you feel otherwise, please bring it up in a future TSC meeting.

> While I like the left to right, depth first merging, one of the reasons I avoided doing that and fell back to alphanumerical order was that alphanumerical order is easy to display.

fwiw, the strong consensus in the TSC meeting today was for left-to-right processing rather than lexical sorting. I don't think I'd want to see a tree structure get output from `wwctl node list`; if people want that, they can just look at the yaml, imo.

> it'd also be nice if the wwctl node set --profile* options allowed append/prepend/insert/delete operations

Maybe true for all the list attributes (e.g., for overlays). But that's starting to get really verbose. At that point, maybe `wwctl node edit` is the better solution.

> the original idea I wanted [...] was for profiles to work like class inheritance with nodes being (logically) nothing more than instances of a profile

In terms of an eventual future where maybe we just have one entity (called Profile or Node, doesn't really matter) I think I can see us getting there eventually. But I'm less certain that comparisons to class inheritance are as straightforward as that; class inheritance is fine for single inheritance; but Warewulf already supports multiple profiles from a Node; so extending that metaphor to multiple inheritance is [notoriously hairy.](https://stackoverflow.com/questions/225929/what-is-the-exact-problem-with-multiple-inheritance)

With your support in Slack, and general support from the TSC in the meeting today, I'll plan on trying to get this approved for merge once I resolve @mslacken's [change request.](https://github.com/warewulf/warewulf/pull/1598#issuecomment-2552153632)

anderbubble:
> > @anderbubble Please rebase on main and remove commit [7a23c3e](https://github.com/warewulf/warewulf/pull/1598/commits/7a23c3e1159747acb3c4324d6d7ac683879d3ba0)
> 
> I'll try. I recall there was an issue that motivated me pulling that commit in; but I forget what it was now.
> 
> If I have trouble, I'll wait until #1595 is merged, and then rebase on that.
> 
> I definitely agree with the request to remove that before merge, though: thanks for noticing it in your review.

@mslacken I found the issue.

```
package github.com/warewulf/warewulf/internal/pkg/node
	imports github.com/warewulf/warewulf/internal/pkg/testenv
	imports github.com/warewulf/warewulf/internal/pkg/node: import cycle not allowed in test
```

Since testenv is currently configuring `node.ConfigFile` it can't be used to test node. I'll try rebasing on #1595 and see if that resolves the issue.

anderbubble:
@mslacken I've confirmed that rebasing this on #1595 fixes the testsuite. I'll avoid pushing that rebase for now, so that we don't accidentally merge them in the wrong order. That means the tests will be failing here until #1595 is merged and this is rebased.

anderbubble:
Marking as draft because I need to update `upgrade` for `Node.Profiles`.

internal/app/wwctl/node/list/main_test.go
```diff
@@ -602,6 +602,29 @@ nodes:
...
+		{
+			name:    "single node list with network",
+			args:    []string{"-a"},
+			wantErr: false,
+			stdout: `
+NODE  FIELD                    PROFILE  VALUE
+----  -----                    -------  -----
+n01   Profiles                 --       default
+n01   NetDevs[default].Hwaddr  --       aa:bb:cc:dd:ee:ff
+n01   NetDevs[default].Ipaddr  --       1.1.1.1
```
mslacken, line 616: Not sure if is must exactly look like that, but no commas and additional zeros should be in here.

mslacken (changes requested):

anderbubble:
I wholeheartedly support PR reviews that add failing tests. :) I'll check it out. Thanks.

anderbubble:
@mslacken found and fixed. Thanks again!

griznog (approved):

anderbubble:
Rebased and added support for nested profiles in `wwctl upgrade nodes`.

anderbubble:
@mslacken still need your signoff before this can be merged.

anderbubble:
Rebased to main.

mslacken (approved):

anderbubble:

PR #1602, opened 2024-12-20
added round robin boot servers

benibr:

This is a draft I put together while experimenting with multiple Warewulf servers all serving boot images into the same network.
The idea is to use the DHCP config generated by Warewulf to distribute the HTTP traffic to a random server out of a given pool to spread out the load during a cluster boot without going through a loadbalancer or depending on DNS.
I'm not aware of a solution to do this directly in the template so I added a config parameter in `warewulf.conf`
If the community is somehow interested in this feature, I can continue to work on that, but I would be happy for any input on how to properly integrate it into Warewulf logic.

## Reviewer  checklist

### DIFF
--- a/etc/warewulf.conf
+++ b/etc/warewulf.conf
@@ -1,5 +1,10 @@
 ipaddr: 10.0.0.1
 netmask: 255.255.252.0
+rrbootaddrs:
+  - 10.0.0.165
+  - 10.0.0.166
+  - 10.0.0.167
+  - 10.0.0.168
 warewulf:
   secure: false
   update interval: 60
--- a/internal/pkg/config/root.go
+++ b/internal/pkg/config/root.go
@@ -29,6 +29,7 @@ var cachedConf WarewulfYaml
 type WarewulfYaml struct {
 	Comment         string        `yaml:"comment,omitempty"`
 	Ipaddr          string        `yaml:"ipaddr,omitempty"`
+	Rrbootaddrs     []string      `yaml:"rrbootaddrs,omitempty"`
 	Ipaddr6         string        `yaml:"ipaddr6,omitempty"`
 	Netmask         string        `yaml:"netmask,omitempty"`
 	Network         string        `yaml:"network,omitempty"`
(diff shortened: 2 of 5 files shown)


anderbubble:
I like the intent, here, for sure; but the implementation won't work because the template is only re-evaluated when you run `wwctl configure dhcp`, not for every DHCP request. Is there any way to actually configure ISC DHCP with multiple addresses that _it_ selects from?

fwiw, I think the initial next-server isn't the most sensitive place to do the load-balancing. Presumably warewulfd could return an HTTP redirect for the container image or overlay images. Then the target of the redirect could be read from a node or profile field.

griznog:
In the end you don't want your nodes to go to a random or round robin boot server/warewuld master, you want them to go to an up and least loaded master or in an extremely large environment maybe the nearest one on the local network segment. You can get this basically for free by having multiple warewuld masters use DNSMasq and let each one point the clients at itself. Then the booting nodes will always get the master that replies first, which will probably be the least busy master. In a very large environment a master in each subnet or with DCHP relays, you can easily direct clients to the nearest warewulf master.

griznog:
For the sake of completeness, if round robin tftp servers are really what you want, then DNSMasq to the rescue again, from the DNSMasq man page for `dhcp-boot` 

> Instead of an IP address, the TFTP server address can be given as a domain name which is looked up in /etc/hosts. This name can be associated in /etc/hosts with multiple IP addresses, which are used round-robin. This facility can be used to load balance the tftp load among a set of servers.

I still think that's a bad idea because there is no dynamic mechanism to avoid offline TFTP servers or masters. Mostly adding this as yet another hint that we need to drop ISC dhcpd and make DNSMasq the default/recommended tool for this :grin:

benibr:
Thanks for the feedback. For clearification: I used this patch with the `static` DHCP config templating. This way I generate the round robin assignment once and each node gets a random Warewulf server.

@anderbubble I like the idea with HTTP forwards. Actually that might be even better since that would be handled completely in warewulf rather than going through the whole DHCP config & next-server.

@griznog interesting research, however I don't have very good experience with dnsmasq in large scale setups. Also I think warewulf should not rely on multiple racy DHCP servers in the same LAN if it wants to go in the direction to be more flexible and HA since in a setup like that you have no control of how traffic is actually distributed and it's also more coplicated to debug.

anderbubble:
@benibr I'd love to see how you end up doing this with resources! It could be great to add to the docs.

PR #1604, opened 2024-12-21
Parallel and optimized overlay builds

anderbubble:
## Description of the Pull Request (PR):

Builds overlays for multiple nodes in parallel, and optimizes the overlay build process.

## This fixes or addresses the following GitHub issues:

- Closes #1018

## Reviewer  checklist

The reviewer checks the following items before merging the PR.

- [x] The PR is based on the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))
- [x] All commits are "Signed off" (e.g., using `git commit --signoff`) in agreement to the [DCO](DCO.txt)
- [x] The [CHANGELOG](https://github.com/warewulf/warewulf/blob/main/CHANGELOG.md) has been updated, if necessary, and under the correct release heading
- [x] The [userdocs](https://github.com/warewulf/warewulf/tree/main/userdocs) have been updated, if necessary
- [x] The submitter is listed in the [contributors file](https://github.com/warewulf/warewulf/blob/main/CONTRIBUTORS.md)
- [x] The test suite has been updated, if necessary

PR #1612, opened 2025-01-07
Benchmark test for overlay build

anderbubble:

In building #1604 I developed a benchmark test to help demonstrate the performance improvement. This included adjustments to the `testenv` interface, which made this change quite large on its own; so I'm separating it out into its own PR, here.

I would have preferred to rebase it to _before_ #1604, but that rebase was difficult; so here it is, after. This shouldn't be merged until _after_ #1604, as it also contains the changes from #1604.

## Reviewer  checklist


mslacken:
What is the difference of this to #1604 @anderbubble ?

anderbubble:
@mslacken This is #1604 _and_ the additional benchmark code and adjustments to testenv, as you requested. This way you can review #1604 without the rather broad changes to testenv, and then review this separately, _after merging #1604_.

internal/pkg/overlay/overlay.go
```diff
@@ -102,40 +101,76 @@ func (this Overlay) IsDistributionOverlay() bool {
...
+	close(errChan)
 
+	for err := range errChan {
+		return err
 	}
 	return nil
 }
 
-// TODO: Add an Overlay Delete for both sourcedir and image
+func BuildSpecificOverlays(nodes []node.Node, allNodes []node.Node, overlayNames []string, workerCount int) error {
```
mslacken, line 142: This is the parallel build as in #1604 so the commits are mangled together

mslacken (changes requested):
Will approve #1604 instead

internal/pkg/overlay/overlay.go
anderbubble: They aren't mangled together; this PR just has to come after #1604, because separating it from changes in #1604 was impractical / unnecessary. Now that #1604 is merged, this one can be considered separately.

mslacken (approved):

Issue #1651, opened 2025-01-24
Slack invite link not working any more


bartwo:
I get the following error when trying to join the slack channel
```
This link is no longer active

To join this workspace, you’ll need to ask the person who originally invited you for a new link.
```

mslacken:
PR is on the ay, you can use
https://join.slack.com/t/warewulf/shared_invite/zt-2yer2fd5y-7rC2r1kBVyGHwQf4VJwzHA
for now

PR #1652, opened 2025-01-24
updated slack link

mslacken:
New slack link is https://join.slack.com/t/warewulf/shared_invite/zt-2yer2fd5y-7rC2r1kBVyGHwQf4VJwzHA

- Close: #1651


anderbubble:
This is frustrating; I _know_ I told Slack to make this link never expire when I updated it last.

It's also at https://github.com/warewulf/warewulf.org/blob/main/src/pages/help.js#L52

How about removing it from the README entirely, replacing it with a link to warewulf.org/help, and updating it at warewulf.org? At least then there's only one place to keep updated.

anderbubble (approved):

PR #1696, opened 2025-02-03
Simplify Makefile and spec

anderbubble:

- explicitly indicate what to include in dist
- prevent duplication of files in specfile

## Reviewer  checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,13 +158,14 @@ init:
 	restorecon -r $(WWTFTPDIR)
 
 .PHONY: dist
-dist:
-	rm -rf .dist/ $(WAREWULF)-$(VERSION).tar.gz
+dist: $(config)
+	rm -rf .dist/
 	mkdir -p .dist/$(WAREWULF)-$(VERSION)
-	rsync -a --exclude=".github"  --exclude=".vscode" --exclude "*~" --exclude $(WAREWULF)-*.tar.gz * .dist/$(WAREWULF)-$(VERSION)/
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -158,13 +158,14 @@ init:
 	restorecon -r $(WWTFTPDIR)
 
 .PHONY: dist
-dist:
-	rm -rf .dist/ $(WAREWULF)-$(VERSION).tar.gz
+dist: $(config)
+	rm -rf .dist/
```
mslacken, line 162: whye the config target? Doesn't this mean we push build artifacts to the tar ball?
Shouldn't this be $(vendor)?

mslacken (changes requested):

Makefile
anderbubble: The dependency on `vendor` is defined at https://github.com/warewulf/warewulf/blob/ac2daae421d4b162afa451286c553b911237f5fa/Makefile#L224, because it is omitted during offline builds. We build a dist in the GitHub action and then pass that to mock which does the actual build; but mock can't fetch vendor because it's intentionally offline at the time. So this prevents errors that would be generated by even _trying_ to update vendor.

`dist` depends on `$(config)` to expand `.in` files to their final versions. That way the distributed dist has complete source code; but this can be over-written later by an additional `make config` if necessary (e.g., with different "defaults" specified).

mslacken (approved):

mslacken:
@anderbubble can you please rebase?

anderbubble:
Thanks for the review! I've rebased and fixed a bug exposed by the conflict.

Issue #1594, opened 2024-12-13
network not inferred from ipaddr + netmask


anderbubble:
### Steps to reproduce

See #1589

If ipaddr and netmask are specified in warewulf.conf, the value of network should be inferred; but it appears that it's not, or at least not correctly.

### Error message

dhcpd.conf subnet section not generated correctly

### Information on your system

v4.5.8

OS immaterial.

### General information

- [X] I have run `wwctl version` and reported the contents of `/etc/os-release`
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have captured and reported relevant error messages and logs

Issue #1541, opened 2024-11-09
Not setting network in warewulf.conf uses default, which can be incorrect and cause issues in overlays that it to be set correctly.


MiddelkoopT:
### Steps to reproduce

```
sms_ip=10.5.0.8
internal_network=10.5.0.0
internal_netmask=255.255.0.0

sed -i "s/^ipaddr:.*/ipaddr: ${sms_ip}/" /etc/warewulf/warewulf.conf
sed -i "s/^netmask:.*/netmask: ${internal_netmask}/" /etc/warewulf/warewulf.conf

[root@head ~]# grep network /etc/warewulf/warewulf.conf 
network: 10.0.0.0
[root@head ~]# sed -i "s/^network:.*//" /etc/warewulf/warewulf.conf
[root@head ~]# grep network /etc/warewulf/warewulf.conf 
[root@head ~]# wwctl configure dhcp
Building overlay for head.novalocal: host
Enabling and restarting the DHCP services
Job for dhcpd.service failed because the control process exited with error code.
See "systemctl status dhcpd.service" and "journalctl -xeu dhcpd.service" for details.
ERROR: failed to start: failed to run start cmd: exit status 1
[root@head ~]# grep Netmask /var/lib/warewulf/overlays/host/rootfs/etc/dhcp/dhcpd.conf.ww 
subnet {{$.Network}} netmask {{$.Netmask}} {
[root@head ~]# grep 10.0 /etc/dhcpd.conf
subnet 10.0.0.0 netmask 255.255.0.0 {
[root@head ~]# 
```
subnet should read 10.5.0.0

### Error message

```
[root@head ~]# wwctl configure dhcp
Building overlay for head.novalocal: host
Enabling and restarting the DHCP services
Job for dhcpd.service failed because the control process exited with error code.
See "systemctl status dhcpd.service" and "journalctl -xeu dhcpd.service" for details.
ERROR: failed to start: failed to run start cmd: exit status 1
```

### Information on your system

```
[root@head ~]# wwctl version
wwctl version:	 61ff492e-1
rpc version: apiPrefix:"rc1"  apiVersion:"1"  warewulfVersion:"61ff492e-1"
```

Build from source, last commit before my latest PR's based on the main branch
`0e0b49a13bf9b37dc209bc0c9c820fe54e61d45f`

```
[root@head ~]# grep '^VERSION=' /etc/*release
/etc/os-release:VERSION="9.4 (Blue Onyx)"
[root@head ~]# go version
[...]

PR #1700, opened 2025-02-04
Fix handling of CIDR values in warewulf.conf

anderbubble:

`warewulf.conf` has historically used `.Ipaddr`, `.Network`, and `.Netmask` settings to configure, primarily, the DHCP server. It is also used to configure the `warewulf` hostname in `/etc/hosts` on the server and cluster nodes.

Later development added support for specifying `.Ipaddr` as a CIDR-style value, e.g. 192.168.0.1/24. But this made the behavior of `.Network` and `.Netmask` unpredictable.

This PR simplifies the handling of all three fields. They remain strings for now; but if, during `Parse()`, `.Ipaddr` is found to be a CIDR-formatted value, the indicated network and netmask are written to the `.Network` and `.Netmask` configuration fields, and `.Ipaddr` is updated to be the bare address. (If either `.Network` or `.Netmask` is already specified, the specified value takes precedence.)

Similar updates have also been made to `.Ipaddr6` and `.Ipv6Net`, though the behavior of `.Ipaddr6` already required a CIDR address with no conversion, so this behavior has been retained.

Helper methods have been added--`.IpCIDR()` and `.NetworkCIDR()`--and their values are now presented to templates. (`.NetworkCIDR` was already present in the tstruct, but not populated.)

Finally, this PR removes a requirements that `.Ipaddr6` represent a `/64` or smaller subnet.

- Fixes: #1541
- Fixes: #1594

## Reviewer  checklist


mslacken:
Thanks for taking care of this mess and structuring it.
I still don't understand why you removed the feature, that without any configuration, we just the default network of the host?

anderbubble:
@mslacken I've added server network auto-detection back in. I'm still mixed on it, as the auto-detection wouldn't make it to `wwclient`. (Right now, wwclient doesn't perform it; but, even if it did, it would detect itself as the server, not the actual Warewulf server.)

I think the only way around this would be to persist the configuration back out from its auto-detected state. I don't think we should write warewulf.conf on every invocation; but maybe we could track an internal boolean on WarewulfYaml that is flagged if auto-detection occurs, and then `.Persist()` during `wwctl configure` if that flag is activated?

I'll think about that some more, and maybe try to implement it later. But let me know if you have any thoughts about that.

anderbubble:
> persist the configuration back out from its auto-detected state

I went ahead and did this, basically as described above. I'm not 100% certain I like just doing it when `wwctl configure` is run, even without arguments; but I think it's probably alright.

Let me know what you think.

JasonYangShadow:
@anderbubble needing rebase

JasonYangShadow (approved):
LGTM, feel free to merge or ask for other members' reviews.

anderbubble:
@mslacken I'm going to go ahead and merge this, and plan to tag rc2 in the morning; but let me know if you have any concerns.

PR #1702, opened 2025-02-05
add completion for profile list

mslacken:
close #1696


anderbubble:
@mslacken I moved the completion function to the completions package so that it could be re-used for all the `wwctl` commands that tab-complete profile names.

Issue #1245, opened 2024-06-07
Add working path to chroot prompt or custom prompt


conxuro:
### Summary

Would be nice to have the working directory in the prompt inside the chroot like in the common shell, or the possibility to customize the prompt, or at least to enable/disable the working dir.

For example:
`[rocky-8-base] Warewulf: /etc/warewulf> `
Instead of:
`[rocky-8-hpcnow-base] Warewulf>`

### Description

To add "\w" to the PS1 variable or the option to set it through warewulf.conf variable, either a variable with a full custom prompt or a true/false var for the working dir.
For example:
```
[...]
container:
  working dir: true
  mounts:
  - source: ...
[...]
```

### General information

- [X] I have searched the issues of this repo and believe this is not a duplicate

PR #1732, opened 2025-02-11
Update PS1 during `wwctl image shell` to include working directory

anderbubble:

- Closes: #1245

## Reviewer  checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - syncuser overlay skips duplicate users and groups in passwd and group databases. #829
 - `wwctl image syncuser --write` is true by default. #1736
 - Update syncuser documentation. #1736
+- Update PS1 during `wwctl image shell` to include working directory by default,
+  and to include `PS1` from the environment if present. #1245
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added `UniqueField` template function. #829
 - Added `wwctl image build --syncuser`. #1321
 - Added support for a DNSSEARCH netdev tag in network configuration overlays. #1256
+- Added `WW_HISTFILE` to control shell history location during `wwctl image shell`. #1732
 
 ### Changed
 
--- a/internal/app/wwctl/image/exec/child/main.go
+++ b/internal/app/wwctl/image/exec/child/main.go
@@ -170,6 +170,8 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 		return fmt.Errorf("failed to mount /run: %w", err)
 	}
 
+	os.Setenv("PATH", "/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin")
+
 	var ps1Base string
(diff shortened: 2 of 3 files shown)


mslacken:
Why not additionally check if there is the environment `WW_PS1` and if set use that one?
Or source a bashrc file which resides in `$CONTAINERDIR/config/bashrc` so that users can have bashrc files for every container?

MiddelkoopT:
> Why not additionally check if there is the environment `WW_PS1` and if set use that one?
> Or source a bashrc file which resides in `$CONTAINERDIR/config/bashrc` so that users can have bashrc files for every container?

I would think that would make it more fragile, also not everyone uses /bin/bash.  I also like to see a different shell on a system like this.

MiddelkoopT (approved):
Look good to me.

anderbubble:
I'll at least experiment with using the existing PS1 variable, and see how that feels.

anderbubble:
@mslacken I updated it to use PS1 if PS1 is present, and simplified the prefix a bit. Let me know what you think.

mslacken:
This is not way, young padawan.
`PS1` is always set so we will extend this variable which may break the syntax of teh specia characters.

anderbubble:
> PS1 is always set

PS1 is usually a shell variable, which does not get passed to processes. It typically has to be exported explicitly before it would go to `wwctl`. Still...

> we will extend this variable which may break the syntax of [the special] characters

Can you provide an example of how it would break? Adding a static prefix to the existing PS1 is pretty normal (e.g., anaconda, virtualenv). I don't think we should introduce a new variable when the information we want (what shell prompt the user prefers) is already in a standard variable.

mslacken:
Try it yourself:
```
Last login: Thu Feb 13 15:27:52 2025 from 192.168.100.1
[root@rocky9 ~]# cat /etc/os-release 
NAME="Rocky Linux"
VERSION="9.2 (Blue Onyx)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="9.2"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Rocky Linux 9.2 (Blue Onyx)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:9::baseos"
SUPPORT_END="2032-05-31"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-9"
ROCKY_SUPPORT_PRODUCT_VERSION="9.2"
REDHAT_SUPPORT_PRODUCT_VERSION="9.2"
[root@rocky9 ~]# echo $PS1
[\u@\h \W]\$
[root@rocky9 ~]# export PS1="$PS1\W]\$" # <- This is what your code will *always* do
[root@rocky9 ~]# ~]$
```
`$PS1` is always set. What you are trying to do is lazily combine the 'inner' and 'outer' shell.
What we enhance and export all shell variables with a `WW_` prefix to the chroot shell without this prefix.
This would be transparent and extensible.

anderbubble:
What you have said is not correct. My original PR only sets `\W]\$ ` if PS1 is _not_ set in the environment. It's a default, not an append.

https://github.com/warewulf/warewulf/pull/1732/commits/6f7540e1fbe6e55d1a639902d5c4b6f7630c01d2#diff-812fd9855df0c48acce819b348ac23d7b65734f17401e5ef396320d162dfecf1R173-R176

anderbubble:
@mslacken updated. Let me know what you think.

mslacken (approved):

PR #1736, opened 2025-02-11
syncuser improvements

### LINKED ISSUE #829: Remove duplicate entries in /etc/passwd on node
### Version of Warewulf

What version of Warewulf are you using?

```
```

### Expected behavior
/etc/passwd on node having unique entries

### Actual behavior
If /etc/passwd within a container is synchronized with that on the head node, the generic overlay's passwd.ww doubles it up.

This doesn't seem to hurt anything at the moment, but it's unexpected and might cause some weirdness if the entries somehow diverge.

wwctl container syncuser --write containername
wwctl overlay show --render nodename generic etc/passwd.ww

```
NAME="Rocky Linux"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"

```

Built rpms from source

### Additional notes:
I'm mostly interested in using the syncuser command to synchronize the system users' UID/GIDs between container and head node from boot, with regular users getting updated with the runtime overlay. Having a "syncsysuser" command that only deals with UIDs in the SYS_{U,G}ID_{MIN,MAX} from /etc/login.defs, then somehow have the passwd.ww template only append the regular users' entries would work.

### LINKED ISSUE #1321: make syncuser emphatically optional, opt-in
### Summary

Syncuser should be emphatically optional and opt-in.

- wwctl commands should never prompt for it automatically or imply that it represents an error or warning
- container build, exec, and import should have --syncuser options to opt-in whenever it makes sense

### Rationale

_No response_

### Description

_No response_

### Additional information

_No response_

### General information

### LINKED ISSUE #1322: don't automatically include users from the head node into the overlay
### Summary

Copying users from the head node into the node's overlay, currently done by wwinit, should be considered part of syncuser, and made optional.

https://github.com/warewulf/warewulf/pull/1306 moves this functionality part of syncuser; and with https://github.com/warewulf/warewulf/issues/1321 I suggest that it should be removed from the list of default overlays.

### Rationale

_No response_

### Description

_No response_

### Additional information

_No response_

### General information

anderbubble:

This change is the overall "make syncuser emphatically optional, opt-in." Mostly I see syncuser cause confusion for people who think you have to do it, rather than people who chose to use it. So this intends to not do any syncuser by default, and then document how to use it if you need it.

- Fixes #1321 
- Fixes #829
- Fixes #1322

## Reviewer  checklist

### DIFF
--- a/internal/pkg/overlay/funcmap.go
+++ b/internal/pkg/overlay/funcmap.go
@@ -130,3 +130,34 @@ func importSoftlink(lnk string) string {
 func softlink(target string) string {
 	return fmt.Sprintf("{{ /* softlink \"%s\" */ }}", target)
 }
+
+// UniqueField returns a filtered version of a multi-line input string. input is
+// expected to be a field-separated format with one record per line (terminated
+// by `\n`). Order of lines is preserved, with the first matching line taking
+// precedence.
+//
+// For example, parsing /etc/passwd filter /etc/passwd for unique user names:
+//
+// Lines without the index field (e.g., blank lines) are always included in the
+// output.
+//
+// UniqueField(":", 0, passwdContent)
+func UniqueField(sep string, index int, input string) string {
+	inputLines := strings.Split(input, "\n")
+	var outputLines []string
+	found := make(map[string]bool)
+	for _, line := range inputLines {
+		inputFields := strings.Split(line, sep)
+		if len(inputFields) > index {
+			field := inputFields[index]
+			if field != "" {
+				if found[field] {
+					continue
+				}
+				found[field] = true
+			}
+		}
+		outputLines = append(outputLines, line)
+	}
+	return strings.Join(outputLines, "\n")
+}
--- a/overlays/syncuser/rootfs/etc/passwd.ww
+++ b/overlays/syncuser/rootfs/etc/passwd.ww
@@ -1,4 +1,7 @@
-# Uncomment the following line to enable passwordless root login
-# root::0:0:root:/root:/bin/bash
-{{IncludeFrom $.ImageName "/etc/passwd"}}
-{{Include "/etc/passwd"}}
+{{- $sources := list }}
+{{- if and .Tags.PasswordlessRoot (eq (lower .Tags.PasswordlessRoot) "true") }}
+{{- $sources = append $sources "root::0:0:root:/root:/bin/bash" }}
+{{- end }}
+{{- $sources = append $sources (IncludeFrom $.ImageName "/etc/passwd" | trim) }}
+{{- $sources = append $sources (Include (printf "%s/%s" .Paths.Sysconfdir "passwd") | trim) }}
+{{- join "\n" $sources | UniqueField ":" 0 | trim }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added support for comma-separated hostlist patterns. #1635
 - Added default value for `warewulf.conf:dhcp.template`. #1725
 - Added `UniqueField` template function. #829
+- Added `wwctl image build --syncuser`. #1321
 
 ### Changed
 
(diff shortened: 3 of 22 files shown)


anderbubble:
@mslacken and @MiddelkoopT, this change is the overall "make syncuser emphatically optional, opt-in." Mostly I see this syncuser cause confusion for people who think you _have_ to do it, rather than people who chose to use it. So this intends to not do any syncuser by default, and then document how to use it _if_ you need it.

But let me know what you think.

mslacken:
Basically I approve this idea, but handling this with templates is a good idea. 
Still there can be a lot of side effects and testing this properly could delay the 4.6.0 release.
Can we push that to 4.6.1?

anderbubble:
> but handling this with templates is a good idea.

I don't understand. What can be done with templates already is being done with templates. Some of it, like the chown and chmod of all the files in the image, must be done in the image, and the updates to /etc/passwd and /etc/group need to match.

None of that is new; syncuser has been doing this since before I joined the project. But it's been built assuming you're always going to use it. But you've never had to, so this change is mostly about disabling all the warnings.

The only other change is in filtering duplicate entries.

> Still there can be a lot of side effects

What new side-effects? Like I said: syncuser is not new. This is just about making it easier to disable, not assuming that people will use it, and (in the most different case) filtering duplicates from the file, which is one of the more common complaints we get about Warewulf.

> testing this properly could delay the 4.6.0 release.

That situation _is_ being tested now. If you want more, I'll add more. In fact, syncuser wasn't being tested at all before, because the dependence on the host's /etc directory meant that we couldn't test it. Moving it to sysconfdir means that testenv can direct where it reads host files from, which lets me enable the tests.

> Can we push that to 4.6.1?

My first inclination is that this is a significant enough change that, if it doesn't make v4.6.0, it'd need to go to v4.7.0. But I'd strongly prefer getting this in v4.6.0.

Just tell me what you want to see tested that isn't. But, again, this isn't a new feature; it's modifications to an existing feature that was _completely_ untested before, and is tested now (in this PR).

internal/pkg/overlay/funcmap.go
```diff
@@ -130,3 +130,21 @@ func importSoftlink(lnk string) string {
 func softlink(target string) string {
 	return fmt.Sprintf("{{ /* softlink \"%s\" */ }}", target)
 }
+
```
mslacken, line 133: Following two points. 
* It just doesn't feel right that this is an internal function. It's run in templating so it's exported. Please name it UniqueField
* Document what the function does, like we document all functions. I like small riddles, but documentation is king.

overlays/syncuser/rootfs/etc/passwd.ww
```diff
@@ -1,4 +1,8 @@
 # Uncomment the following line to enable passwordless root login
 # root::0:0:root:/root:/bin/bash
```
mslacken, line 2: We should remove this completely or make it in such a way that this root entry also runs through the UniqueField pipeline. You put so much effort on this, just to end up with two root entries.

mslacken (changes requested):

internal/pkg/overlay/funcmap.go
anderbubble: > It just doesn't feel right that this is an internal function.

I don't disagree; I just followed the example for all the existing functions. But I'm fine changing it.

anderbubble: > Document what the function does, like we document all functions.

Done.

overlays/syncuser/rootfs/etc/passwd.ww
anderbubble: Thanks for calling this out. I refactored how `/etc/passwd` is building its output and moved the passwordless root login option to an optional inclusion managed by specifying a `PasswordlessRoot` tag. I think this is a real improvement, too, since it doesn't require modification of the template _and_ doesn't get forgotten in the template after debugging.

anderbubble: (I also added a test for this case.)

mslacken (approved):
Just merging this one before thinking more about the idea of not just having `.Tags.PasswordlessRoot` but `Tags.AdditionalPasswdLine` may add the possibility to have new warewulf managed users on the systems. Hopefully nobody reades ever this comment @anderbubble

PR #1740, opened 2025-02-14
create commandline options for Makefile

mslacken:
@anderbubble simply checkout this PR and run `make help`
I would like to have this in 4.6.0 (as it doesn't have any functionality) and can 
you document the targets I forgot?

Stolen from the Schedmd people without any shame as syncuser creates so much
pain which we need only for munge


anderbubble:
@mslacken I'll check it out.

mslacken:
You have to approve @anderbubble  I can't self approve

anderbubble:
I love it.

I made some changes:

- add targets
- update some of the descriptions
- restore the meaning of `all` (while still defining `help` as the default target)
- add info about OFFLINE_BUILD
- update warewulf.spec and userdocs to use "build" or "all" explicitly now

But what a cool thing. It looks great.

I think I'm done making modifications, so I'll merge as soon as the tests are finished running.

anderbubble (approved):

PR #1753, opened 2025-02-20
added test for onboot parameter

mslacken:
onboot paramter is completey on works as expected if not set!

- Addresses: #1754


mslacken:
See #1757 for a fix

anderbubble:
Included in #1762

PR #1757, opened 2025-02-20
use {{ if ne $.OnBoot "false" }} in templates

### LINKED ISSUE #1754: OnBoot implementation is broken
### Steps to reproduce

Setting the onboot network parameter to false will still set this flag in the templates as
```
{{- if $netdev.OnBoot }}
```
will always evaluate to true, regardless what ` $netdev.OnBoot` contains!

### Error message

_No response_

Failing in tests in #1753

### General information

mslacken:
close #1754


mslacken:
If `OnBoot` is now not explicitly to false the interface show up on boot.
I still would rather have a manual switch, but this mimic the old behavior as good as possible, as there the constructor set the `OnBoot` on the interfaces to `true`.

anderbubble:
Closed by #1754.

PR #1766, opened 2025-02-24
Add option to set editor for wwctl overlay edit

MiddelkoopT:

This PR adds the `--command` option to `wwctl overlay edit` to override the `$EDITOR` environment variable.  It also allows `$EDITOR` and `--command` to have arguments (think `emacs -nw`).  The arguments are split by space but preserve single and double quotes as a single argument (escaping quotes not supported).  This allows using editor commands such as `sed -i -e` and `tail -a` to modify overlay files from the command line.

## Reviewer  checklist

### DIFF
--- a/internal/app/wwctl/overlay/edit/root.go
+++ b/internal/app/wwctl/overlay/edit/root.go
@@ -22,11 +22,13 @@ var (
 	}
 	CreateDirs bool
 	PermMode   int32
+	Command    string
 )
 
 func init() {
 	baseCmd.PersistentFlags().BoolVarP(&CreateDirs, "parents", "p", false, "Create any necessary parent directories")
 	baseCmd.PersistentFlags().Int32VarP(&PermMode, "mode", "m", 0755, "Permission mode for directory")
+	baseCmd.PersistentFlags().StringVarP(&Command, "command", "c", "", "Command to run editor")
 }
 
 // GetRootCommand returns the root cobra.Command for the application.
--- a/internal/pkg/util/util.go
+++ b/internal/pkg/util/util.go
@@ -284,6 +284,23 @@ func FindFilterFiles(
 	return ofiles, nil
 }
 
+// Split a string by spaces perserving quoted strings (no escaping)
+func ShellSplit(s string) []string {
+	re := regexp.MustCompile(`"([^"]*)"|'([^']*)'|\S+`)
+	matches := re.FindAllStringSubmatch(s, -1)
+	var result []string
+	for _, match := range matches {
+		if match[1] != "" {
+			result = append(result, match[1]) // Double-quoted match
+		} else if match[2] != "" {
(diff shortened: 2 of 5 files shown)


MiddelkoopT:
Rebased against v4.6.0

MiddelkoopT:
Updated CHANGELOG

internal/pkg/util/util.go
```diff
@@ -284,6 +284,23 @@ func FindFilterFiles(
...
+			result = append(result, match[1]) // Double-quoted match
+		} else if match[2] != "" {
+			result = append(result, match[2]) // Single-quoted match
+		} else {
+			result = append(result, match[0]) // Non-quoted match
+		}
+	}
+	return result
+}
+
```
anderbubble, line 303: I'm worried about this function. If we're going to lex `$EDITOR`, I think it should be done with a library like https://pkg.go.dev/github.com/google/shlex.

internal/app/wwctl/overlay/edit/root.go
```diff
@@ -22,11 +22,13 @@ var (
 	}
 	CreateDirs bool
 	PermMode   int32
+	Command    string
 )
 
 func init() {
 	baseCmd.PersistentFlags().BoolVarP(&CreateDirs, "parents", "p", false, "Create any necessary parent directories")
 	baseCmd.PersistentFlags().Int32VarP(&PermMode, "mode", "m", 0755, "Permission mode for directory")
+	baseCmd.PersistentFlags().StringVarP(&Command, "command", "c", "", "Command to run editor")
```
anderbubble, line 31: What is the use case for `--command` vs just using `env EDITOR="emacs -nw" wwctl [...]`? I looked at `git commit`, for example, and it doesn't seem to have such an argument.

anderbubble (changes requested):

internal/app/wwctl/overlay/edit/root.go
MiddelkoopT: The use case is to use `sed` `yq` directly on files.  This way you don't have to know the parent directory for the overlay and wwctl will do the "right thing" when copying from a distribution to a local overlay.

In OpenHPC we use it for things like this (this is new and needed for the 3.3 release)
```
wwctl overlay cat host /etc/hosts.ww | \
  sed -e 's_\({{$node.Id}}{{end}}\)_{{$node.Id}}.localdomain \1_g' | \
  EDITOR=tee wwctl overlay edit host /etc/hosts.ww
```

With this it would look like
```
wwctl overlay edit --command "sed -e 's_\({{$node.Id}}{{end}}\)_{{$node.Id}}.localdomain \1_g'"  host /etc/hosts.ww
```

middelkoopt:
Abandoning this PR and will revisit if needed.  The usecase for OpenHPC has been removed.  Will revisit with a different strategy if needed.

Issue #1777, opened 2025-02-25
empty network device id leads to undesired default NetworkManager behavior


anderbubble:
### Steps to reproduce

NetworkManager isn’t connecting the config to the interface, due to the lack of a device name being specified.

- You have not specified a device
- This should be ok, as the mac address should be enough for NetworkManager to correlate the config with the interface
- However, Warewulf is generating a config with an empty interface-name: `interface-name=`
- This is interfering with NetworkManager’s detection algorithm: since there is an interface-name parameter specified, it never tries to detect the interface based on its mac address
- With no configuration provided, NetworkManager defaults to dhcp

### Error message

NetworkManager uses DHCP rather than the provided configuration.

### Information on your system

v4.6.0rc3 on Rocky Linux 9

### General information

- [x] I have run `wwctl version` and reported the contents of `/etc/os-release`
- [x] I have searched the issues of this repo and believe this is not a duplicate
- [x] I have captured and reported relevant error messages and logs

anderbubble:
I have confirmed that removing `interface-name=` from the keyfile entirely fixes the issue.

PR #1779, opened 2025-02-25
Better handling for missing NetDev.Device and clean-up dracut NetworkManager connections

anderbubble:

- Updates ifcfg and NetworkManager overlays to account for NetDev.Device being blank.
- Removes left-over NetworkManager connections from dracut

- Fixes: #1777
- Fixes: #1780

## Reviewer  checklist


anderbubble:
A report in Slack matched my experience in dev; but this fix is reportedly not fixing their problem as originally reported in #1777.

Still investigating; but I think this is a clear fix regardless.

anderbubble:
Second part should be fixed now.

anderbubble:
Converting to draft because it's not clear to me why this:

```
autoconnect-priority:-100
```

Isn't causing the dracut connection to be ignored in favor of the new connection. With a configuration of my own, it works; but when I port in the reporter's configuration, I see the same problem they report.

I'll need to do some comparisons to better understand the problem.

anderbubble:
On further review, I think the only reason I wasn't noticing this before was because (a) it was being masked by our mounting a new /run over /run in wwinit before; and (b) my local environment had static DHCP configured, so I was getting the correct address over DHCP even when it was ignoring the local config.

I now think the fix in this PR is correct.

mslacken (approved):

PR #1782, opened 2025-02-26
WWWORKER overwrites runtime.NumCPU

mslacken:
runtime.NumCPU varies for different build hosts, so the
environment variable WWWORKER can be set to keep this number
constant as this number ends up in
docs/man/man1/wwctl-overlay-build.1
and so in the packages


mslacken:
@anderbubble marked this for 4.6.0 but we can postpone this for 4.6.1, still then 4.6.0 wouldn't be reprodible

anderbubble:
It's also in `docs/man/man1/wwctl-overlay-import.1`.

I think I'd rather just use a sentinel value of 0 to indicate 1 cpu per worker, and then handle that in the code. I'll submit a suggestion for your review.

anderbubble (changes requested):
Please review https://github.com/warewulf/warewulf/pull/1789 as an alternative that also fixes `wwctl overlay import`.

PR #1787, opened 2025-02-27
make dracut curl invocation retry for all errors

snehring:

the curl invocation should also include --retry-all-errors so non-transient errors are also retried.
This will cover the case where the network is unreachable, possibly because an ip hasn't been assigned yet.

NA

## Reviewer  checklist


snehring:
In my testing --retry does not cover the case where the network is not accessible. It would cover a timeout but not a complete failure to connect to the server. 

We could go back to my original plan and have ping test connectivity to warewulfd before the curl invocation.

snehring:
closing in favor of #1800

PR #1797, opened 2025-03-03
fixup pdf build

mslacken:
- sphinx_reredirects isn't part of distros standard packages and so local pdf build fails
- latex doesn't like special characaters like ≥ as it's sooo old, use >= instead

### DIFF
--- a/userdocs/conf.py
+++ b/userdocs/conf.py
@@ -7,7 +7,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = 'Warewulf User Guide'
-copyright = '2024, Warewulf Project Contributors'
+copyright = '2025, Warewulf Project Contributors'
 author = 'Warewulf Project Contributors'
 release = 'main'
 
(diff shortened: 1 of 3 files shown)


userdocs/conf.py
```diff
@@ -16,7 +16,6 @@
 
 extensions = [
     'sphinx.ext.graphviz',
-    'sphinx_reredirects',
```
anderbubble, line 19: We need redirects to support redirects on the website. If we need to support an environment that doesn't support redirects, we'll need separate `conf.py` or some other dynamic way to handle it.

anderbubble (changes requested):

userdocs/conf.py
mslacken: If the pdf tag is defined for sphinx the redirects will be exlcuded

Makefile
```diff
@@ -89,7 +89,7 @@ reference: wwctl
 	./wwctl --emptyconf genconfig reference userdocs/reference/
 
 latexpdf: reference
-	make -C userdocs latexpdf
+	SPHINXOPTS='-t pdf -D release=$(VERSION)' make -C userdocs latexpdf
```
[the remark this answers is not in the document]
anderbubble: I now understand that `-D` is what is allowing `release=$(VERSION)` to be specified.

anderbubble (approved):

PR #1800, opened 2025-03-03
Add connectivity test/wait to dracut

snehring:

Curl fetches of system images can fail if the interface is not fully up. This adds an up to 60s delay, testing every second, for connectivity to warewulfd to be functional.

## Reviewer  checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Update GitHub actions to build aarch64 artifacts.
+## Unreleased
+
+### Added
+- Add connectivity check to dracut before image download. 
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Update GitHub actions to build aarch64 artifacts.
-## Unreleased
 
 ### Added
+
 - Add connectivity check to dracut before image download. #1800 
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## v4.6.1, unreleased
 
+### Added
+
+- Add connectivity check to dracut before image download. #1800
+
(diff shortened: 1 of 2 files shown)


snehring:
I didn't have it abort if the check fails after 60s, I figure the attempt to download the image will handle that if the networking still isn't up when it's reached.

snehring:
@anderbubble let me know if you want that one-liner refactored to fit the style of the rest of the bash in that file. Or if you have other suggestions.

anderbubble:
Can you see if #1833 addresses your issue, as an alternative to this?

snehring:
#1833 does not solve the issue, it's not actually getting a connrefused it's failing to connect at all because the network is unreachable . This PR is a bandaid, the actual issue is dracut declaring that the ib interface is 'up' and moving forward with the boot process when it shouldn't. I still haven't figured out what's causing that.

MiddelkoopT:
We are seeing issues during our OpenHPC development for 3.3 where the NetworkManger online check is passing even though the interface is not yet up.  Currently investigating. https://github.com/openhpc/ohpc/issues/2091

MiddelkoopT:
I suspect that the root cause of the issue is that the network is not setup correctly yet and sleeping/retry is a bandage as you indicated.  Dracut uses `nm-wait-online-initrd.service` to wait/block for the network to come online, which uses
```
ExecStart=/usr/bin/nm-online -s -q -t 3600
```
to wait for the network.  Assuming using NetworkManager, which is what I got for a stock Rocky 9.5 install.

In order to use this you must have `rd.neednet=1` on the kernel cmdline.

This may not wait for full connectivity.  The `-s` switch will not wait for network connectivity, so removing this switch from the systems unit for Dracut (`usr/lib/dracut/modules.d/35network-manager/nm-wait-online-initrd.service`) may ensure that the network is fully up (untested).

A dump of journalctl and/or the `/run/initramfs/rdsosreport.txt` would help diagnose the root cause of the issue.

You can get to this by adding `rd.break` to the kernel command line.

MiddelkoopT:
Good point - Have you tried systemd-networkd?  I just realized it is in epel... I will need to look at this.

snehring:
Actually it may be fine. Adding an ```override.conf``` for ```nm-wait-online-initrd.service``` does seem to fix (at least in our test cluster) the issue this PR fixes. I don't have a good setup to test exhaustively with IB and ethernet connections in our test cluster, but adding an ethernet connection to the list of configured addresses (which seemed to connect before ib) didn't seem to break the boot. Created #1850 to hopefully address this.

Issue #1817, opened 2025-03-14
panic in warewulfd when no netdev specified


filthyrake:
### Steps to reproduce

Standing up a new WW 4.6-rc2 system, every time I would try to PXE boot a node I would see a panic from warewulf.  

During node configuration I had NOT been setting --netdev for anything.  Once I went back and modified the nodes to have a defined netdev, the problem was resolved and I am able to PXE boot my nodes properly.

Really basically identical behavior to: https://github.com/warewulf/warewulf/issues/1689 but with a seemingly different cause.

### Error message

```
Mar 14 16:12:30 localhost wwctl[988]: 2025/03/14 16:12:30 http: panic serving 192.168.0.7:43852: runtime error: invalid memory address or nil pointer dereference
Mar 14 16:12:30 localhost wwctl[988]: goroutine 23 [running]:
Mar 14 16:12:30 localhost wwctl[988]: net/http.(*conn).serve.func1()
Mar 14 16:12:30 localhost wwctl[988]: #011/usr/lib/golang/src/net/http/server.go:1903 +0xbe
Mar 14 16:12:30 localhost wwctl[988]: panic({0x10f34a0?, 0x1c23060?})
Mar 14 16:12:30 localhost wwctl[988]: #011/usr/lib/golang/src/runtime/panic.go:770 +0x132
Mar 14 16:12:30 localhost wwctl[988]: github.com/warewulf/warewulf/internal/pkg/warewulfd.GetNodeOrSetDiscoverable({_, _})
Mar 14 16:12:30 localhost wwctl[988]: #011/builddir/build/BUILD/warewulf-4.6.0rc2/internal/pkg/warewulfd/nodedb.go:82 +0x4c1
Mar 14 16:12:30 localhost wwctl[988]: github.com/warewulf/warewulf/internal/pkg/warewulfd.ProvisionSend({0x1432b90, 0xc000208a80}, 0xc0001d8120)
Mar 14 16:12:30 localhost wwctl[988]: #011/builddir/build/BUILD/warewulf-4.6.0rc2/internal/pkg/warewulfd/provision.go:77 +0x72b
Mar 14 16:12:30 localhost wwctl[988]: net/http.HandlerFunc.ServeHTTP(0xc000462e60?, {0x1432b90?, 0xc000208a80?}, 0xc00007d008?)
Mar 14 16:12:30 localhost wwctl[988]: #011/usr/lib/golang/src/net/http/server.go:2171 +0x29
[...]

PR #1818, opened 2025-03-17
Allow discover of nodes with network only in profile

mslacken:
- Fixes: #1817

Add panicing test to confirm


mslacken:
close #1817

anderbubble (approved):
Thanks for fixing this, @mslacken!

PR #1820, opened 2025-03-19
fix SUSE specific udev regression

mslacken:
Fix regression that net.ifnames=1 was missing in the kernel command line, which forces SUSE udev to default behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - wwclient uses `WW_IPADDR`, if set, to contact the Warewulf server. #1788
 - Add `wwctl node import --yes` to assume yes to confirmations.
 - Set an IPMI tag ``vlan`` to configure the vlan during ``ipmiwrite``. #1031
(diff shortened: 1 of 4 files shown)


anderbubble:
What is the behavior without this?

I don't see `net.ifnames` having been set before in commit history. (I see it in the docs; but not in code or config files.)

mslacken:
It was set in the defaults.conf as far as I know and without it we have to add it into the code.

anderbubble:
Weird... I see it now, where I didn't before. I wonder what was wrong with my grepping.

anderbubble:
> Weird... I see it now, where I didn't before. I wonder what was wrong with my grepping.

Actually, I'm looking again, and the only places I see reference to it are in userdocs that you committed. (e.g., d1c734732f5f2ff3f7a2e2df22124238de03d1c9)

This makes me wonder if this was a `defaults.conf` setting that was in the suse package, but not in upstream. Can you check?

I really don't want to push back on this; but if it wasn't upstream before, I want to be careful about introducing it now. It's a suse-specific issue, as you've indicated in the title; so maybe it should just be a suse-specific change to the nodes.conf distributed in the suse package.

anderbubble (approved):

PR #1833, opened 2025-03-22
Add --retry-connrefused to dracut curl commands

anderbubble:
## Description of the Pull Request (PR):

Retry curl during dracut in more circumstances.

Potential alternative to #1800.

## Reviewer  checklist

The reviewer checks the following items before merging the PR.

- [ ] The PR is based on the appropriate branch (typically [main](https://github.com/warewulf/warewulf/tree/main/userdocs))
- [ ] All commits are "Signed off" (e.g., using `git commit --signoff`) in agreement to the [DCO](DCO.txt)
- [ ] The [CHANGELOG](https://github.com/warewulf/warewulf/blob/main/CHANGELOG.md) has been updated, if necessary, and under the correct release heading
- [ ] The [userdocs](https://github.com/warewulf/warewulf/tree/main/userdocs) have been updated, if necessary
- [ ] The submitter is listed in the [contributors file](https://github.com/warewulf/warewulf/blob/main/CONTRIBUTORS.md)
- [ ] The test suite has been updated, if necessary

PR #1850, opened 2025-03-31
Make nm-wait-online-initrd.service wait for interfaces to be online during dracut boot

snehring:

While discussing #1800 the underlying cause of dracut not necessarily waiting for configured interfaces to actually be up was found. This PR adds an ```override.conf``` when the ```systemd``` and ```network-manager``` modules are included in dracut for the ```nm-online-wait-initrd.service``` unit that removes the ```-s``` argument to the ```nm-online``` invocation in that unit. ```nm-online -s``` returns once ```NetworkManager``` is up and configuration for defined interfaces is requested but not necessarily when interfaces are fully configured and up. This change should make ```nm-online``` wait until the timeout is reached (300s in this pr) or configured interfaces are actually up.

## Reviewer  checklist

### DIFF
--- a/dracut/modules.d/90wwinit/nm-wait-online-initrd.service.override
+++ b/dracut/modules.d/90wwinit/nm-wait-online-initrd.service.override
@@ -1,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/nm-online -q -t 300
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+### Added
+
+- Added override.conf for nm-wait-online-initrd.service with dracut.
+
 ## v4.6.1, 2025-04-04
 
 ### Added
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ install: build docs ## Install Warewulf from source
 	for f in docs/man/man1/*.1.gz; do install -m 0644 $$f $(DESTDIR)$(MANDIR)/man1/; done
 	for f in docs/man/man5/*.5.gz; do install -m 0644 $$f $(DESTDIR)$(MANDIR)/man5/; done
 	install -pd -m 0755 $(DESTDIR)$(DRACUTMODDIR)/90wwinit
-	install -m 0644 dracut/modules.d/90wwinit/*.sh $(DESTDIR)$(DRACUTMODDIR)/90wwinit
+	install -m 0644 dracut/modules.d/90wwinit/*.sh  dracut/modules.d/90wwinit/*.override $(DESTDIR)$(DRACUTMODDIR)/90wwinit
 
 .PHONY: install-sos
 install-sos:
--- a/dracut/modules.d/90wwinit/module-setup.sh
+++ b/dracut/modules.d/90wwinit/module-setup.sh
@@ -17,4 +17,8 @@ install() {
     inst_multiple cpio curl dmidecode
     inst_hook cmdline 30 "$moddir/parse-wwinit.sh"
     inst_hook pre-mount 30 "$moddir/load-wwinit.sh"
+    if dracut_module_included "network-manager" && dracut_module_included "systemd"
+    then
+        inst_simple "$moddir/nm-wait-online-initrd.service.override" "/etc/systemd/system/nm-wait-online-initrd.service.d/override.conf"
+    fi
 }


snehring:
I believe the license check is transient and unrelated to this change, but I'm not sure I can restart checks.

anderbubble:
> I believe the license check is transient and unrelated to this change

It is; we've been having this problem in other PRs as well.

@MiddelkoopT and @snehring can you both confirm that this actually resolves the problem you're seeing?

anderbubble:
Also, to be clear, this obsoletes #1800 and #1833, right?

snehring:
It solves the problem I created #1800 for, but I do think #1833 is still good to have. Anything we can do to add resiliency to the image download process is good.

snehring:
This change is only relevant for dracut though, once the switchroot happens the config this does is gone. For @MiddelkoopT 's issue a separate PR that adds the the override into the NetworkManager overlay would probably be good.

MiddelkoopT:
The issues I saw was for the non-dracut boot process (OpenHPC) and I have yet to write an upstream PR for it (an build-in overlay or the warewulf-rockylinux container), I have a fix in for OpenHPC.

I do not know if the issue I discovered also appeared in the dracut version.  I have yet to fully test it (trying to do it today, but I'm at a workshop).

MiddelkoopT:
This does not resolve the issue that I'm seeing.  My issue is later in in the boot process and I cannot reproduce any network issues in the dracut phase (dracut works fine). I still saw my original issue without my fix applied to the node image (bringing up the network manager in the node image, not dracut).  For me the patch did not seem to change any of the dracut process (systemd).

snehring:
That makes sense. Have you tried adding in an override for ```NetworkManager-wait-online.service``` in the NetworkManager overlay and making sure that's in the list of overlays for your system overlay?

MiddelkoopT:
Yes, similar solutions work, I'm working on a PR for it now against the Warewulf containers themselves, since it seems to be more related to the image being broken than having a fix applied to every image via a built-in overlay.

snehring:
I don't think it's an issue with a particular image. I make mine from scratch (well from dnf) and that unit has the same issue. I think that's just how NetworkManager-wait-online behaves per redhat and fedora. I think it's just nm behavior we should address in the overlay.

dracut/modules.d/90wwinit/nm-wait-online-initrd.service.override
MiddelkoopT, line 1: The override should clear the current ExecStart (`ExecStart=`), otherwise it will stack them and run both commands.
```ini
[Service]
ExecStart=
ExecStart=/usr/bin/nm-online -q -t 300
```

MiddelkoopT, line 1: Why is the timeout set to 5 min and not the default?

snehring: The default (which seems to be 3600) seemed a bit long to me, decision was arbitrary.

snehring: Let me test. It's possible running both is adding a delay that's addressing the issue I'm hoping to resolve.

MiddelkoopT: I believe the default is 30 seconds - https://networkmanager.dev/docs/api/latest/nm-online.html

snehring: I'm fine with 30, but as distributed from redhat (and seemingly dracut upstream https://github.com/dracutdevs/dracut/blob/5d2bda46f4e75e85445ee4d3bd3f68bf966287b9/modules.d/35network-manager/nm-wait-online-initrd.service#L4)  the argument to -t is 3600.

MiddelkoopT: Interesting - I just double checked the service file and found this in `/usr/lib/systemd/system/NetworkManager-wait-online.service`
```ini
Type=oneshot
ExecStart=/usr/bin/nm-online -s -q
RemainAfterExit=yes

# Set $NM_ONLINE_TIMEOUT variable for timeout in seconds.
# Edit with `systemctl edit NetworkManager-wait-online`.
#
# Note, this timeout should commonly not be reached. If your boot
# gets delayed too long, then the solution is usually not to decrease
# the timeout, but to fix your setup so that the connected state
# gets reached earlier.
Environment=NM_ONLINE_TIMEOUT=60
```
Looks like Rocky and AlmaLinux does 60 seconds.

snehring: Check the dracut network-manager module (probably `/usr/lib/dracut/modules.d/35network-manager/nm-wait-online-initrd.service`). I suspect it will be different.

MiddelkoopT: Your right!
```ini
[Service]
Type=oneshot
ExecStart=/usr/bin/nm-online -s -q -t 3600
RemainAfterExit=yes
```
So man locations!  I agree that 500 is much better than 3600.

snehring: This seems to still work as expected. I've done a few reboot cycles on our test cluster and so far there've been no issues. I'll get this added in shortly.

snehring:
rebased to current main

MiddelkoopT (approved):
I approve.

PR #1852, opened 2025-04-02
Enable Support for IPv6 PXE

dacianstremtan:

This allow PXE using IPv6 IP address for the controller. 

Manual change needs to be done to the 

/etc/warewulf/ipxe/default.pxe:
e.g. set baseuri http://{{.Ipaddr6}}:{{.Port}}/provision/{{.Hwaddr}}

/etc/dhcp/dhcpd.conf: 
....
if exists user-class and option user-class = "iPXE" {
    filename "http://[.Ipaddr6]:9873/ipxe/${mac:hexhyp}";
} else 
...
/etc/dhcp/dhcpd6.conf:
....
if exists dhcp6.user-class and substring (option dhcp6.user-class, 2, 4) = "iPXE" {
       option dhcp6.bootfile-url "http://[.Ipaddr6]:9873/ipxe/${mac:hexhyp}";
}
else if exists dhcp6.client-arch-type and option dhcp6.client-arch-type = 00:07 {
       option dhcp6.bootfile-url "tftp://[.Ipaddr6]/warewulf/snponly.efi" ;
}
....

- Fixes #

## Reviewer  checklist

### DIFF
--- a/internal/pkg/config/root.go
+++ b/internal/pkg/config/root.go
@@ -137,16 +137,21 @@ func (conf *WarewulfYaml) Parse(data []byte, autodetect bool) error {
 			}
 		}
 	}
-
-	if conf.Ipaddr6 != "" {
-		if _, network, err := net.ParseCIDR(conf.Ipaddr6); err == nil {
-			if conf.Ipv6net == "" {
-				conf.Ipv6net = network.IP.String()
-			}
-		} else {
-			return fmt.Errorf("invalid ipv6 address: must use CIDR notation: %s", conf.Ipaddr6)
-		}
-	}
+        if conf.Ipaddr6 != "" {
+                if conf.Ipv6net == "" {
+                        wwlog.Error("Ipv6 network has not been set in warewulf.conf: ipv6net")
+                        return fmt.Errorf("Invalid Ipv6 network")
+                }
+                _, ipv6net, err := net.ParseCIDR(conf.Ipv6net)
+                if err != nil {
+                        wwlog.Error("Invalid Ipv6 address specified, must be CIDR notation: %s", conf.Ipv6net)
+                        return fmt.Errorf("Invalid Ipv6 network")
+                }
+                if msize, _ := ipv6net.Mask.Size(); msize > 64 {
+                        wwlog.Error("ipv6 mask size must be smaller than 64")
+                        return fmt.Errorf("Invalid Ipv6 network size")
+                }
+        }
 
 	return nil
 }
--- a/internal/pkg/config/root.go
+++ b/internal/pkg/config/root.go
@@ -137,21 +137,21 @@ func (conf *WarewulfYaml) Parse(data []byte, autodetect bool) error {
 			}
 		}
 	}
-        if conf.Ipaddr6 != "" {
-                if conf.Ipv6net == "" {
-                        wwlog.Error("Ipv6 network has not been set in warewulf.conf: ipv6net")
-                        return fmt.Errorf("Invalid Ipv6 network")
-                }
-                _, ipv6net, err := net.ParseCIDR(conf.Ipv6net)
-                if err != nil {
-                        wwlog.Error("Invalid Ipv6 address specified, must be CIDR notation: %s", conf.Ipv6net)
-                        return fmt.Errorf("Invalid Ipv6 network")
-                }
-                if msize, _ := ipv6net.Mask.Size(); msize > 64 {
-                        wwlog.Error("ipv6 mask size must be smaller than 64")
-                        return fmt.Errorf("Invalid Ipv6 network size")
-                }
-        }
+	if conf.Ipaddr6 != "" {
+		if conf.Ipv6net == "" {
+			wwlog.Error("Ipv6 network has not been set in warewulf.conf: ipv6net")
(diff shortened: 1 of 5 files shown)


middelkoopt:
Thanks for the contribution.  I'm currently working on testing this PR against main and targeting v4.6.3, which we hope will happen at the end of the month.  I have a few changes so far that I will comment here.  If you like, you are welcome to address them as well as rebase the patch against main.

internal/pkg/config/root.go
```diff
@@ -137,14 +137,19 @@ func (conf *WarewulfYaml) Parse(data []byte, autodetect bool) error {
...
+			return fmt.Errorf("Invalid Ipv6 network")
+		}
+		_, ipv6net, err := net.ParseCIDR(conf.Ipv6net)
+		if err != nil {
+			wwlog.Error("Invalid Ipv6 address specified, must be CIDR notation: %s", conf.Ipv6net)
+			return fmt.Errorf("Invalid Ipv6 network")
+		}
+		if msize, _ := ipv6net.Mask.Size(); msize > 64 {
+			wwlog.Error("ipv6 mask size must be smaller than 64")
+			return fmt.Errorf("Invalid Ipv6 network size")
```
middelkoopt, line 152: Why is Ipv6net being set explicitly?  Ipv6net is being set from Ipaddr6 and seems redundant to have two things in the configuration.  The current IPv4 code works this way now.  

As for mask size, for most cases having a prefix greater than 64 is a bad idea but it could be used in a static case, and this prevents this.  At least make it a warning if not remove.

As part of merge/testing effort I'm removing this block.

middelkoopt:
> e.g. set baseuri http://{{.Ipaddr6}}:{{.Port}}/provision/{{.Hwaddr}}

This is not a valid URI w/o the `[]` around the raw IPv6 address, why is this the case here?

dacianstremtan:
> > e.g. set baseuri http://{{.Ipaddr6}}:{{.Port}}/provision/{{.Hwaddr}}
> 
> This is not a valid URI w/o the `[]` around the raw IPv6 address, why is this the case here?

Because it works. I wanted to use minimum changes to the file. Since it worked from the start I did not try a different approach.

Screenshot_client1-rocky8_2025-07-21_10:09:53
Screenshot_client1-rocky8_2025-07-21_10:10:01

dacianstremtan:
With square brackets also works. I think we should use square brackets when for consistent configurations

Dacian
Screenshot_client1-rocky8_2025-07-21_10:18:36

internal/pkg/config/root.go
middelkoopt, line 1: These changes have been dropped in #1958 in favor of more recent commits.

middelkoopt: `conf.Ipaddr6` now stored with prefix as per more recent commit.

middelkoopt (changes requested):
I rebased and reworked this PR in #1958.  In order for it to be merged with #1958 you will have to add a signed-off for each commit, please do not rebase as I've already rebased it.  Alternatively I could just squash your commits and add a Co-authored-by - but will need to check if this is sufficient Signed-off-by.

Please note, that you should use `{{.Authority}}` template variable for the address/port combo with `[]` inserted for IPv6 addresses.  `.Ipaddr6` is still set as your original patch.  The `deafult.ipxe` has been upated for this.

Please test to see if this works for you.  

Please let me know if this approach works for you.

It would be great if you could send us your overlay files to help build out full IPv6 support.

internal/pkg/config/root.go
middelkoopt: This section has been removed as part of the rework patch #1958.

dacianstremtan:
If I recall the reason to set the network explicitly, was that the IPv6 was
being used with the \64 afterward when reaching to the warewulf server.

I looked at the changes you made and I will try them on my dev system and
let you know of my findings.

Dacian

În lun., 28 iul. 2025 la 16:40, Timothy Middelkoop ***@***.***>
a scris:

dacianstremtan:
Hi,
"Please let me know if this approach works for you."
Yes. I downloaded and compiled your the tm-ipv6-pxe branch and it works for
me.

"It would be great if you could send us your overlay files to help build
out full IPv6 support."
I don't have specific overlay files. Maybe you can expand on this request.

Dacian


În lun., 28 iul. 2025 la 16:40, Timothy Middelkoop ***@***.***>
a scris:

middelkoopt:
Good to hear that it works.

I mean the dhcpd6.conf and other configuration files.  It is possible to generate files on the host from a template (host overlay files) that get generated with`wwctl configure`.  Did you use this or just create/edit the files yourself? 

Also along similar lines, do you use /etc/hosts for node IP addresses?  If so, how is this generated/maintained.

Could you update your commit's to have a signed-of-by line?

middelkoopt:
I squashed commits and added a "co-authored-by" attribution.

anderbubble:
Closing in favor of #1958.

dacianstremtan:
Thank you.
I appreciate your help in moving this forward.
I will share my dhcpv6.conf and dhcpd.conf files today
Dacian

În vin., 1 aug. 2025 la 14:42, Timothy Middelkoop ***@***.***>
a scris:

PR #1872, opened 2025-04-28
add node overlay buildtime

JasonYangShadow:

This PR adds node overlays (system/runtime) build time. 

- Fixes #

## Reviewer  checklist

### DIFF
--- a/internal/pkg/node/datastructure.go
+++ b/internal/pkg/node/datastructure.go
@@ -24,9 +24,11 @@ type Node struct {
 	id    string
 	valid bool // Is set true, if called by the constructor
 	// exported values
-	Discoverable wwtype.WWbool     `yaml:"discoverable,omitempty" json:"discoverable,omitempty" lopt:"discoverable" sopt:"e" comment:"Make discoverable in given network (true/false)"`
-	AssetKey     string            `yaml:"asset key,omitempty"    json:"asset key,omitempty"    lopt:"asset"                 comment:"Set the node's Asset tag (key)"`
-	Profile      `yaml:"-,inline"` // include all values set in the profile, but inline them in yaml output if these are part of Node
+	Discoverable        wwtype.WWbool     `yaml:"discoverable,omitempty" json:"discoverable,omitempty" lopt:"discoverable" sopt:"e" comment:"Make discoverable in given network (true/false)"`
+	AssetKey            string            `yaml:"asset key,omitempty"    json:"asset key,omitempty"    lopt:"asset"                 comment:"Set the node's Asset tag (key)"`
+	Profile             `yaml:"-,inline"` // include all values set in the profile, but inline them in yaml output if these are part of Node
+	RuntimeOverlayMTime string            `yaml:"runtime overlay mtime,omitempty" json:"runtime overlay mtime,omitempty"`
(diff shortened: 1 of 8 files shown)


MatthewHink (approved):

anderbubble:
Marking draft until the Warewulf PR is merged.

internal/pkg/node/datastructure.go
```diff
@@ -24,9 +24,11 @@ type Node struct {
...
-	AssetKey     string            `yaml:"asset key,omitempty"    json:"asset key,omitempty"    lopt:"asset"                 comment:"Set the node's Asset tag (key)"`
-	Profile      `yaml:"-,inline"` // include all values set in the profile, but inline them in yaml output if these are part of Node
+	Discoverable        wwtype.WWbool     `yaml:"discoverable,omitempty" json:"discoverable,omitempty" lopt:"discoverable" sopt:"e" comment:"Make discoverable in given network (true/false)"`
+	AssetKey            string            `yaml:"asset key,omitempty"    json:"asset key,omitempty"    lopt:"asset"                 comment:"Set the node's Asset tag (key)"`
+	Profile             `yaml:"-,inline"` // include all values set in the profile, but inline them in yaml output if these are part of Node
+	RuntimeOverlayMTime string            `yaml:"runtime overlay mtime,omitempty" json:"runtime overlay mtime,omitempty"`
+	SystemOverlayMTime  string            `yaml:"system overlay mtime,omitempty" json:"system overlay mtime,omitempty"`
```
anderbubble, line 31: Rather than adding `<Runtime|System>OverlayMTime` to the node, I suggest we should add `GET /api/nodes/{id}/overlays`. That should return a new json object, represented here as yaml because it's easier:

```
system overlay:
  overlays:
  - overlay1
  - ...
  - overlayN
  build:
    mtime: <value>

runtime overlay:
  overlays:
  - overlay1
  - ...
  - overlayN
  build:
    mtime: <value>
```

anderbubble (changes requested):

internal/pkg/node/datastructure.go
JasonYangShadow: we either need to pull all nodes' runtime/system overlay info initially(this would result in adding another api) or pull a node's runtime/system overlay info (mainly mtime) when ListPopover is triggered (users might need to wait for a short time to see the overlays info and build time).

anderbubble: At the very least, we should _definitely_ not be tracking `<Runtime|System>OverlayMTime` in the `Node` struct: this should always be derived directly from the file system by looking at the .img and/or .img.gz for the node's overlays.

It's ok for showing the build time to request an additional API hit. I suggest we fetch it when generating the popover for the overlay list.

Screenshot 2025-06-04 at 3 57 58 PM

anderbubble:
@JasonYangShadow I made some small changes and added tests; but I'm doing that over at https://github.com/warewulf/warewulf/pull/1945 in stead. I'm thinking that it will be better to have one big PR for all the API additions we're contributing to support the next release of warewulf-cockpit, rather than try to get them merged one at a time. That way we can make more changes until we have the web interface the way we like it, and then get the necessary changes approved all at once.

Issue #1875, opened 2025-04-30
Add Ignition RAID support


iankgt40:
### Summary

Add RAID support into Warewulf's Ignition integration

### Rationale

Can we add RAID support into Warewulf's Ignition integration? It would be useful to be able to tap into Ignition's ability to build RAID arrays of various levels at boot time so it is reproducible every boot.

### Description

Ignition has RAID support built in, with the ability to create arrays using standard schemes such as RAID0, RAID1, RAID10, RAID5, etc. This would be better than constructing the RAID after node boot using local tools since you would have to either rebuild every time, check for configs and import every time. Also, while post boot scripts can also be used, since we already have Ignition hooks in place as an overlay, it makes sense to extend the disk management capabilities to encompass the complete set of options.

### General information

- [x] I have searched the issues of this repo and believe this is not a duplicate

Issue #1821, opened 2025-03-19
Persistent installation


mslacken:
### Summary

* If a disk is given dump image to disk and provide rsync mechanism to update it on reboot

### General information

- [x] I have searched the issues of this repo and believe this is not a duplicate

PR #1849, opened 2025-03-31
Persistent without rsync

mslacken:
Dump the image simply on a rootfs  if as `root=persistent` is set. As `/` the partition with the name and mount point `/rootfs` is used.
@anderbubble Can you merge this one now. No rsync code is there now and I'm **realy** tired to rebase this all the time. It's quite simple code and doesn't change zillions of files like #1588
Also unlike #1588 it's not just one big stashed commit, which makes it easy to track when the changes were made.
close #1557
close #1821

PR #1894, opened 2025-05-10
support provisioning the image to disk

anderbubble:

Add support for provisioning the node image to disk and provisioning file systems during dracut.

Alternative to
- https://github.com/warewulf/warewulf/pull/1557
- https://github.com/warewulf/warewulf/pull/1849

Compared to the previous PRs:
- ignition is optional in the dracut image
- use an existing file system if it already exists
- specify the target device using `--root` (rather than a "persistent" constant)

## Issues

- Closes: #1821
- Closes: #1875

## Reviewer  checklist


anderbubble:
@mslacken I made some changes. Let me know what you think.

Even if you think it's good, please don't merge until I've had a chance to present to Greg.

internal/pkg/overlay/overlay.go
```diff
@@ -496,12 +496,7 @@ func RenderTemplateFile(fileName string, data TemplateStruct) (
 		"softlink":     softlink,
 		"readlink":     filepath.EvalSymlinks,
 		"IgnitionJson": func() string {
-			str := createIgnitionJson(data.ThisNode)
-			if str != "" {
-				return str
-			}
-			writeFile = false
-			return ""
+			return createIgnitionJson(data.ThisNode)
```
mslacken, line 499: Looks much simpler this way. I think I added the checks as security measurement, but forgot why

overlays/ignition/rootfs/warewulf/ignition.json.ww
```diff
@@ -1 +1 @@
-{{ IgnitionJson }}
+{{ coalesce IgnitionJson (.Resources.ignition | toJson) }}
```
mslacken, line 1: That's clever.

userdocs/nodes/disks.rst
```diff
@@ -79,6 +79,33 @@ existing file system is wiped.
      --diskname /dev/vda --partname scratch \
      --fsname scratch --fsformat btrfs --fspath /scratch
 
+Resource definition
+===================
+
+Ignition may alternatively be configured using an "ignition" resource.
+
```
mslacken, line 86: Great idea, fixes the raid issue.

mslacken (approved):
If I understand the code correctly, the root partition is simply determined by the `root` kernel parameter?
You approach is much cleaner and better than with what I came up.

mslacken:
How are the partitions visible on the running node?
Especially the rootfs?

anderbubble:
I think I've fixed the ability to configure ignition with resources; but I haven't tested it in the lab yet: just in the suite.

mslacken:
Did Greg have any comments?

anderbubble:
> How are the partitions visible on the running node? Especially the rootfs?

Got extra partitions working with a `.Resources` configuration now. Looks like this:

Screenshot 2025-05-23 at 11 21 24 AM

From this config:

```
  resources:
    ignition:
      storage:
        disks:
          - device: /dev/sda
            partitions:
              - label: rootfs
                sizeMiB: 2048
              - label: scratch
                sizeMiB: 512
              - label: swap
                sizeMiB: 1024
            wipeTable: true
        filesystems:
          - device: /dev/disk/by-partlabel/rootfs
            format: ext4
            path: /
            wipeFilesystem: false
          - device: /dev/disk/by-partlabel/scratch
            format: ext4
            path: /scratch
            wipeFilesystem: true
          - device: /dev/disk/by-partlabel/swap
            format: swap
            path: swap
```

dracut/modules.d/90wwinit/load-wwinit.sh
```diff
@@ -22,6 +18,29 @@ do
...
+        /usr/bin/ignition --root "${NEWROOT}" --config-cache=/tmp/wwinit/warewulf/ignition.json --platform=metal --stage=disks || die "warewulf: failed to partition/format disk"
+    else
+        info "warewulf: /usr/bin/ignition not found. Assuming ${wwinit_root_device} already prepared."
+    fi
+fi
+
+info "warewulf: Mounting ${wwinit_root_device} at ${NEWROOT}"
+(
+    if [ "${wwinit_root_device}" = "tmpfs" ]; then
+        mount -t tmpfs -o mpol=interleave ${wwinit_tmpfs_size_option} "${wwinit_root_device}" "${NEWROOT}"
```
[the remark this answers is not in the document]
anderbubble: What happens if it's not supported? Does it break, or just ignore it? (It's silently ignored, for example, with ramfs.)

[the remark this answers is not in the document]
MiddelkoopT: Catastrophic. Fails to mount and fails to boot.

mslacken:
If understand this correctly, we need either a configuration for sfdisk or one for ignition?
Having both configuration would result in trouble?

anderbubble:
> If understand this correctly, we need either a configuration for sfdisk or one for ignition? Having both configuration would result in trouble?

In theory you could partition different disks with different systems; but the point is for sfdisk+mkfs(+mkswap)+fstab to be an alternative to ignition.

userdocs/nodes/disks.rst
```diff
@@ -79,6 +79,33 @@ existing file system is wiped.
...
+                   shouldExist: true
+                   wipePartitionEntry: true
+               wipeTable: true
+           filesystems:
+             - device: /dev/disk/by-partlabel/scratch
+               format: btrfs
+               path: /scratch
+               wipeFilesystem: false
+
+If any configuration is provided for a node with explicit arguments to ``wwctl
```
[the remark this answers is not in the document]
anderbubble: Updated with greater specificity.

anderbubble:
> Did Greg have any comments?

After conversation, the main directive is to support doing it without ignition; so I've included support for sfdisk, mkfs, and mkswap (with fstab).

dracut/modules.d/90wwinit/load-wwinit.sh
```diff
@@ -22,6 +18,29 @@ do
...
             "${wwinit_uri}" \
         | gzip -d \
-        | cpio -im --directory="${NEWROOT}"
+        | cpio -ium --directory="${NEWROOT}"
     ) || die "Unable to load stage: ${stage}"
+}
+
+if [ "${wwinit_root_device}" != "tmpfs" ]; then
+    mkdir /tmp/wwinit
+    NEWROOT=/tmp/wwinit get_stage "system"
```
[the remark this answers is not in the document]
anderbubble: Done.

dracut/modules.d/90wwinit/load-wwinit.sh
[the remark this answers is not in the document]
anderbubble: Please open this as a separate issue.

dracut/modules.d/90wwinit/parse-wwinit.sh
```diff
@@ -1,27 +1,42 @@
...
+case "${root}" in
+wwinit|wwinit:*)
+    info "warewulf: root=${root}"
+
+    export wwinit_uri="$(getarg wwinit.uri)"
+    if [ -n "${wwinit_uri}" ]; then
+        info "warewulf: Found root=${root} and a Warewulf server uri. Will boot from Warewulf."
+        rootok=1
+    else
+        die "warewulf: Found root=${root} but no Warewulf server uri. Cannot boot from Warewulf."
```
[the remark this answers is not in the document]
anderbubble: Updated.

anderbubble:
@mslacken and @MiddelkoopT this ended up getting more sprawling than I realized, but I tested it on Rocky 8 and 9, and it's been pretty solid. But please let me know if you have any additional feedback.

dracut/modules.d/90wwinit/load-wwinit.sh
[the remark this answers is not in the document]
anderbubble: Sorry: I see https://github.com/warewulf/warewulf/issues/1913 now.

dracut/modules.d/90wwinit/load-wwinit.sh
```diff
@@ -1,16 +1,12 @@
 #!/bin/bash
 
-info "Mounting tmpfs at $NEWROOT"
-mount -t tmpfs -o mpol=interleave ${wwinit_tmpfs_size_option} tmpfs "$NEWROOT"
-
-for stage in "image" "system" "runtime"
-do
-    info "Loading stage: ${stage}"
+get_stage() {
```
mslacken, line 3: Sounds picky, but this function should be called `apply_stage` as it does changes and isn't just a getter

mslacken (changes requested):
1. The behavior of` sfdisk` when it detects disk with partitions on it needs to be documented.
2. Having two competing disk approaches isn't good. When including `sfdisk` we need to deprecate ignition and remove it in 4.7.0

userdocs/overlays/overlays.rst
```diff
@@ -310,6 +317,13 @@ ignition
 
 The **ignition** overlay defines partitions and file systems on local disks.
 
+To use ignition during Dracut (so that the root file system may be provisioned
+before the image is loaded) include Ignition in the Dracut image.
+
+.. code-block:: shell
+
+   wwctl image exec rockylinux-9 -- /usr/bin/dracut --force --no-hostonly --add wwinit --add ignition --regenerate-all
+
```
mslacken, line 349: There needs to be a hint, that the dracut files can be added by a bind overlay

userdocs/server/bootloaders.rst
```diff
@@ -391,3 +399,157 @@ The wwinit module provisions to tmpfs. By default, tmpfs is permitted to use up
...
+
+   fdisk:
+     - device: /dev/sda
+       label: gpt
+       partitions:
+         - device: /dev/sda1
+           name: rootfs
+   mkfs:
+     - device: /dev/disk/by-partlabel/sfdisk-rootfs
+       type: ext4
```
mslacken: Why is this different to the sfdisk example?
So do we have three possibilities to setup disks
* ignition
* fdisk
* sfdisk
Or is this wrong?

mslacken:
Screenshot From 2025-06-03 15-15-48

The node still boost, but `/sbin/init` missing doesn't sound good.

userdocs/server/bootloaders.rst
```diff
@@ -391,3 +399,157 @@ The wwinit module provisions to tmpfs. By default, tmpfs is permitted to use up
...
+its resource.
+
+.. code-block:: yaml
+
+   fstab:
+     - file: /
+       spec: /dev/disk/by-partlabel/rootfs
+       vfstype: ext4
+     - file: /scratch
+       mntops: nofail
```
mslacken: Why `file` and not `mountpoint`?

mslacken:
Perhaps we should add a sfdisk profile to the default installation.
Also this is all brittle, as the validity of the values in the resource aren't checked.
What I would like is to have something like
```
wwctl profile add --partition /dev/vda1 --mount / --fs ext4
``` 
and the rest is added.
I know that this can't be done with resources, still I am wondering if it would have been easier to deduct the configurations for sfdisk and friends from the ignition one?

mslacken:
We have now 
* ` /warewulf/wwinit.d/`
and
* `/warewulf/init.d`\

WHY?

middelkoopt:
> I know that this can't be done with resources, still I am wondering if it would have been easier to deduct the configurations for sfdisk and friends from the ignition one?

Not sure I understand what you are trying to say here. 

A related question is why not just go with ignition?

anderbubble:
> why not just go with ignition?

Two reasons:

- ignition is not sufficiently available; e.g., it's not packaged for Rocky Linux 8
- @gmkurtzer asked for an implementation that did not depend on ignition

> WHY? [...] /warewulf/wwinit.d/ [and] /warewulf/init.d

This is documented at https://github.com/warewulf/warewulf/pull/1894/files#diff-3fbedf5fb6cde4b87f7464e1ba51ac37d7d9095f8aec31887f1e04d6239e9d51R195-R203. Effectively, the two directories support scripts being run before the image is written and after the image is written.

> this is all brittle, as the validity of the values in the resource aren't checked

imo, this is the nature of using resources rather than things built into Warewulf: validity checking is left to the overlay. But that brings me to the second point:

> I am wondering if it would have been easier to deduct the configurations for sfdisk and friends from the ignition one

I am on-board with it actually making sense to support partitions and file systems as a first-class configuration in Warewulf, rather than delegating it all out to resources; but it is specifically because the current configuration for that is _too_ based on what ignition expects that I haven't. I think it would be good to take another look at that to see how reasonable it would be, and perhaps (plan to?) change the partition/fs configuration in a future release to make it more general-purpose for both ignition and these tools.

> do we have three possibilities to setup disks

sfdisk is not like ignition: it only does partitioning. So you need separate tools (mkfs and mkswap) to actually format them.

> There needs to be a hint, that the dracut files can be added by a bind overlay

I don't understand what this means.

> The behavior of sfdisk when it detects disk with partitions on it needs to be documented.

Agreed. Good catch.

> Having two competing disk approaches isn't good. When including sfdisk we need to deprecate ignition and remove it in 4.7.0

I disagree: this is the same as supporting ifcfg, NetworkManager, wicked, netplan and debian.network_interfaces. That comparison further supports keeping a centralized partition + fs configuration in the profile/node struct; but I think that configuration needs to be re-worked, if only slightly, to be simplified and generalized.

userdocs/server/bootloaders.rst
anderbubble: First, because this is what it was already called in the previous release.

Then, because "file" is what fstab actually calls the entry. The fstab overlay+resource follows the nomenclature of https://www.man7.org/linux/man-pages/man5/fstab.5.html

anderbubble:
> The node still boost, but `/sbin/init` missing doesn't sound good.

That is very strange. I don't know how that could happen.

https://github.com/warewulf/warewulf/pull/1894/files#diff-c8b65b73e6546b28addf8f96b9d2159cb52ea8219e886ff666b25fa240a80e0dR33

In particular, the "invoking" log should come with a newline, so the output format doesn't even make sense to me. It makes me think that something else is emitting that error.

mslacken:
> 
> 
> > WHY? [...] /warewulf/wwinit.d/ [and] /warewulf/init.d
> 
> This is documented at https://github.com/warewulf/warewulf/pull/1894/files#diff-3fbedf5fb6cde4b87f7464e1ba51ac37d7d9095f8aec31887f1e04d6239e9d51R195-R203. Effectively, the two directories support scripts being run before the image is written and after the image is written.
> 

I completely disagree here.  Everything which runs after the disk image is written can and **must** be a systemd unit. We **must not** reinvent init V.

> > this is all brittle, as the validity of the values in the resource aren't checked
> 
> imo, this is the nature of using resources rather than things built into Warewulf: validity checking is left to the overlay. But that brings me to the second point:
> 
> > I am wondering if it would have been easier to deduct the configurations for sfdisk and friends from the ignition one
> 
> I am on-board with it actually making sense to support partitions and file systems as a first-class configuration in Warewulf, rather than delegating it all out to resources; but it is specifically because the current configuration for that is _too_ based on what ignition expects that I haven't. I think it would be good to take another look at that to see how reasonable it would be, and perhaps (plan to?) change the partition/fs configuration in a future release to make it more general-purpose for both ignition and these tools.

For the time beeing, could we deduct the sfdik configuration from the ignition one? Having two sources of truth leads just to trouble.

> 
> > There needs to be a hint, that the dracut files can be added by a bind overlay
> 
> I don't understand what this means.
> 

I mean something like
```
wwctl image execl leap15.6  --bind /usr/lib/dracut/modules.d/90wwinit/ -- ....
```


> I disagree: this is the same as supporting ifcfg, NetworkManager, wicked, netplan and debian.network_interfaces. That comparison further supports keeping a centralized partition + fs configuration in the profile/node struct; but I think that configuration needs to be re-worked, if only slightly, to be simplified and generalized.

 You are wrong. I am not complaining that we have sfdisk and ignition. The problem here is two sources of truth.

anderbubble:
Following TSC discussion, the three remaining adjustments planned are:

- updates sfdisk, mkfs, and mkswap to support the native fields in the node/profile datastructure
- generate mount units automatically from mkfs and mkswap
- update docs to include a full-stack example of a stateful boot

anderbubble:
@mslacken and @middelkoopt I've got documentation to do still; but I've got the required changes implemented.

This configuration works in my environment:

```
nodes:
  wwnode1:
    profiles:
      - default
    image name: rockylinux-9
    system overlay:
      - sfdisk
      - mkfs
      - mkswap
      - systemd.mount
      - systemd.swap
    root: /dev/disk/by-partlabel/rootfs
    tags:
      IPXEMenuEntry: dracut
    disks:
      /dev/sda:
        wipe_table: true
        partitions:
          rootfs:
            number: "1"
            size_mib: "2048"
            should_exist: true
          scratch:
            number: "2"
            size_mib: "512"
            should_exist: true
          swap:
            number: "3"
            size_mib: "1024"
            should_exist: true
    filesystems:
      /dev/disk/by-partlabel/rootfs:
        format: ext4
        path: /
        wipe_filesystem: true
      /dev/disk/by-partlabel/scratch:
        format: ext4
        path: /scratch
        wipe_filesystem: true
      /dev/disk/by-partlabel/swap:
        format: swap
        path: swap
```

New overlays:

- systemd.mount mounts file systems from the `filesystems` fields
- systemd.swap activates swap spaces from the `filesystems` fields

I have _not_ removed the automatic mounts from the ignition overlay.

overlays/ignition/rootfs/etc/systemd/system/ww4-disks.target.ww
```diff
@@ -1,24 +1,26 @@
...
-{{- $prefix = substr 1 -1 $prefix }}
+{{- if .FileSystems }}
+{{- range $fsdevice, $fs := .FileSystems }}
+{{- if eq $fs.Format "swap" }}
+Wants={{ $fsdevice | SystemdEscapePath }}.swap
+{{- else if ne $fs.Path "" }}
+Wants={{ $fs.Path | SystemdEscapePath }}.mount
+{{- end }}
+{{- end }}
+{{- else if .Resources.ignition.storage.filesystems }}
```
mslacken, line 18: Are we still having this in ressources?

mslacken (approved):
There is a change in symlink handling, can you please drop a line about this in the README?
Was there a reason to write the own systemd escape, as there is https://github.com/coreos/go-systemd/blob/d5623bf85e8e73ae6352f78ee6b55a287619dd4e/unit/escape.go#L102 ?

anderbubble:
> There is a change in symlink handling, can you please drop a line about this in the README?

I'll review and make sure it's noted. I don't currently remember such a change, though; so please point it out to me if you have a chance.

> Was there a reason to write the own systemd escape

I wasn't aware of the implementation in coreos. I'll try to use that.

internal/pkg/node/constructors.go
```diff
@@ -194,3 +194,16 @@ func (config *NodesYaml) FindDiscoverableNode() (Node, string, error) {
 
 	return EmptyNode(), "", ErrNoUnconfigured
 }
+
+func (node *Node) SetIds(id string) {
```
mslacken: This is an internal function and shouldn't exported

PR #1919, opened 2025-06-08
add nfs-ganesha support

rafalop:
This PR intends to add NFS Ganesha  support to Warewulf. NFS ganesha is a feature rich nfs server implemented in userspace and is available to install as system packages for all common linux distros. This PR implements support for NFS ganesha by closely following what is in place for the kernel nfs server, and contains these features:

- updated config structure to include network prefix length
  * used for ganesha configuration
  * useful for non redhat-based OSs network config, and other configs/templates where CIDR notation is used
- adds ganesha export struct
- adds a basic ganesha configuration template in host overlay
- documentation to get started with nfs-ganesha configuration

- Fixes #

## Reviewer  checklist


anderbubble:
@rafalop an interesting suggestion. I have two responses, one technical / detailed and one more architectural / philosophical:

- I don't think this should have ganesha-specific alternatives for existing configuration entities. Configuration should be expressed as generically as possible, and then that configuration should be implemented as necessary in whatever underlying implementation is being used (e.g., the in-kernel NFS server or ganesha).
- But that leads me to my second thought: what is the value here? Warewulf can use a Ganesha server without having to wrap the ability to configure it, especially after v4.6, now that client mounts are handled in `nodes.conf`. The `wwctl configure nfs` support is increasingly a minor convenience, and configuring a ganesha instances is probably better handled directly than wrapped by Warewulf.

I'm happy to hear points to the  contrary; but it might be better to discuss that in an upcoming TSC meeting. Would you be willing to attend to discuss your proposal?

anderbubble (changes requested):
@rafalop an interesting suggestion. I have two responses, one technical / detailed and one more architectural / philosophical:

- I don't think this should have ganesha-specific alternatives for existing configuration entities. Configuration should be expressed as generically as possible, and then that configuration should be implemented as necessary in whatever underlying implementation is being used (e.g., the in-kernel NFS server or ganesha).
- But that leads me to my second thought: what is the value here? Warewulf can use a Ganesha server without having to wrap the ability to configure it, especially after v4.6, now that client mounts are handled in nodes.conf. The wwctl configure nfs support is increasingly a minor convenience, and configuring a ganesha instances is probably better handled directly than wrapped by Warewulf.

I'm happy to hear points to the contrary; but it might be better to discuss that in an upcoming TSC meeting. Would you be willing to attend to discuss your proposal?

middelkoopt:
I was thinking along similar lines.  @anderbubble does the NFS structures do anything anymore except upgrade?

In some unrelated work that I've just started, I've started to want to be able to run some scripts as a part of the overlay generation or something similar (perhaps nodes - its incomplete work) .  I know I could just wrap it locally in a script, but having things work out of the box with Warewulf would be nice.

anderbubble:
> does the NFS structures do anything anymore except upgrade

The NFS configuration in `/etc/warewulf/warewulf.conf` is currently used by `wwctl configure nfs` to set up `/etc/exports` on the Warewulf server.

middelkoopt:
> The NFS configuration in `/etc/warewulf/warewulf.conf` is currently used by `wwctl configure nfs` to set up `/etc/exports` on the Warewulf server.

I wonder this this could also be accomplished with a host overlay, combined with a generic "service reload" (or script) capability as a part of the overlays.  For NFS for now, might be easily done with the current "system name" functionality in `warewulf.conf`?  Also, I'm guessing that at some point the host overlays should to be split like the client ones.

rafalop:
Thanks for considering this change.

> * I don't think this should have ganesha-specific alternatives for existing configuration entities. Configuration should be expressed as generically as possible, and then that configuration should be implemented as necessary in whatever underlying implementation is being used (e.g., the in-kernel NFS server or ganesha).

This is possible, though it would mean extracting the kernel nfs server export options from the string and converting them into the ganesha equivalent parameter.  Doing this in template is a bit cumbersome but not impossible, for example i could do something of this nature in the template for every desirable option:
```
    # Access Type
    {{- if contains .ExportOptions "ro"}}
    Access_Type = RO;
    {{- else if contains .ExportOptions "rw"}}
    Access_Type = RW;
    {{- else}}
    Access_Type = RW;  # default
    {{- end}}
```
Ganesha-only options like 'pseudo' could just be strings in the comma delimited options (would fail if user switches to kernel server without modifying the string). Actually hadn't thought about it this way until now, but I could make the export options string anything I wanted and play with the sprig regex functions in the template and likely achieve the same result.


> * But that leads me to my second thought: what is the value here? Warewulf can use a Ganesha server without having to wrap the ability to configure it, especially after v4.6, now that client mounts are handled in `nodes.conf`. The `wwctl configure nfs` support is increasingly a minor convenience, and configuring a ganesha instances is probably better handled directly than wrapped by Warewulf.

Really, the value is just the ganesha version of why the kernel nfs configuration wrapper exists - convenience. 

> 
> I'm happy to hear points to the contrary; but it might be better to discuss that in an upcoming TSC meeting. Would you be willing to attend to discuss your proposal?

Yes I'd be willing, although I do not feel strongly about it so I am quite okay to drop it and use workarounds. That said, ganesha aside I am keen to make the network prefix length an available param in templates, either from a direct struct field (as in this commit) or from a template func that can convert the netmask to prefix length. Alternatively if you have a good way to obtain the net prefix in a template from current code please share!

anderbubble:
> I am keen to make the network prefix length an available param in templates, either from a direct struct field (as in this commit) or from a template func that can convert the netmask to prefix length.

I'd probably want to do it similarly to how we're already doing the CIDR versions.

https://github.com/warewulf/warewulf/blob/a99b4eead2533a2a1d6748b699a8644db7094228/internal/pkg/config/root.go#L156-L168

https://github.com/warewulf/warewulf/blob/a99b4eead2533a2a1d6748b699a8644db7094228/internal/pkg/overlay/datastructure.go#L71

Can you open a new issue just for that?

PR #1938, opened 2025-07-09
Fix wwctl upgrade nodes to handle kernel argument lists

anderbubble:

`wwctl upgrade nodes` could convert legacy kernel argument formats to the new list format, but then wasn't properly parsing the new list format for repeat runs.

## Reviewer  checklist

### DIFF
--- a/internal/pkg/upgrade/node.go
+++ b/internal/pkg/upgrade/node.go
@@ -588,8 +588,15 @@ type KernelConf struct {
 func (legacy *KernelConf) Upgrade(imageName string) (upgraded *node.KernelConf) {
 	upgraded = new(node.KernelConf)
 	switch args := legacy.Args.(type) {
-	case []string:
-		upgraded.Args = args
+	case []interface{}:
+		for _, arg := range args {
+			switch arg.(type) {
+			case map[string]interface{}, []interface{}, map[interface{}]interface{}:
+				wwlog.Warn("unable to parse Kernel.Args: non-scalar value %v", arg)
+			default:
+				upgraded.Args = append(upgraded.Args, fmt.Sprintf("%v", arg))
+			}
+		}
 	case string:
 		if args != "" {
 			upgraded.Args = strings.Fields(args)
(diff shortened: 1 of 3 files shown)


internal/pkg/upgrade/node.go
```diff
@@ -588,8 +588,14 @@ type KernelConf struct {
...
 	upgraded = new(node.KernelConf)
 	switch args := legacy.Args.(type) {
-	case []string:
-		upgraded.Args = args
+	case []interface{}:
+		for _, arg := range args {
+			if argStr, ok := arg.(string); ok {
+				upgraded.Args = append(upgraded.Args, argStr)
+			} else {
+				wwlog.Warn("unable to parse Kernel.Args: %v", argStr)
```
middelkoopt: When a kernel arg (as a list) is set to a number, say 1, it will not display the kernel arg value and drop it from the arg list (basically silently drop it) (it's not a string)
```
vscode ➜ /workspaces/warewulf (tm-dev) $ sudo /usr/local/bin/wwctl upgrade nodes --add-defaults --replace-overlays
WARN   : unable to parse Kernel.Args: 
```

```
--- nodes.conf-old      2025-07-13 00:39:38.623662693 +0000
+++ nodes.conf  2025-07-13 00:39:38.623662693 +0000
@@ -25,7 +25,6 @@
         - quiet
         - crashkernel=no
         - net.ifnames=1
-        - 1
```

middelkoopt (changes requested):

internal/pkg/upgrade/node.go
anderbubble: Nice catch! Thanks!

anderbubble:
@middelkoopt to be fair, an integer value wouldn't've been in the file to begin with, because wwctl will only put strings there. (KernelArgs is a list of strings.) But I've adjusted the upgrade logic to be more robust, regardless.

I'm going to clear your "changes required" having added a test for this case and fixed it so that @JasonYangShadow can review and potentially approve; but that's only because I don't want to assume you'll be available to review tonight, and I'm trying to get v4.6.3 out tonight or tomorrow. Feel free to review and even request additional changes if you _do_ have time and the inclination.

PR #1939, opened 2025-07-09
API for node power control

MatthewHink:

Contains the API for node power control.

- Fixes #

## Reviewer  checklist

### DIFF
--- a/internal/pkg/warewulfd/api/power.go
+++ b/internal/pkg/warewulfd/api/power.go
@@ -1,0 +1,200 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/swaggest/usecase"
+	"github.com/swaggest/usecase/status"
+	"github.com/warewulf/warewulf/internal/pkg/bmc"
+	"github.com/warewulf/warewulf/internal/pkg/node"
+	"github.com/warewulf/warewulf/internal/pkg/wwlog"
+)
+
+// PowerResult is the result structure for single node power status operations. get/set power status
+type PowerResult struct {
+	NodeId string `yaml:"node_id" json:"node_id" description:"ID of the node for which power status is being reported"`
+	BmcIp  string `yaml:"bmc_ip" json:"bmc_ip" description:"BMC IP address of the node"`
+	Result string `yaml:"result" json:"result" description:"Power status of the node, e.g., 'on', 'off', 'unknown'"`
+	Err    error  `yaml:"error,omitempty" json:"error,omitempty" description:"Error message if any error occurred while getting power status"`
+}
+
+// getPower returns an interactor to get the power status of a node by its ID.
+// At this time we ony support a single node. There is no expansion of the hostlist args or batching yet.
+func getPower() usecase.Interactor {
+	type getPowerInput struct {
+		ID string `path:"id" required:"true" description:"ID of node to get power status for"`
+	}
+
+	u := usecase.NewInteractor(func(ctx context.Context, input getPowerInput, output *[]PowerResult) error {
+		wwlog.Debug("api.getPower(), ID: %s", input.ID)
+		powerResults := []PowerResult{}
+		powerResults = append(powerResults, PowerResult{ // start with one expansion is not yet supported
+			NodeId: input.ID,
+		})
+
+		if registry, err := node.New(); err != nil {
+			wwlog.Debug("api.getPower() getting node registry: %v", err)
+			powerResults[0].Err = err // Set the error in the first PowerResult
+			*output = powerResults    // Set *output on exit. important
+			return err
+		} else {
+			if node_, err := registry.GetNode(input.ID); err != nil {
+				wwlog.Debug("api.getPower() node not found: %v", err)
+				powerResults[0].Err = err
+				*output = powerResults // Set *output on exit. important
+				return status.Wrap(fmt.Errorf("node not found: %v (%v)", input.ID, err), status.NotFound)
+			} else {
+				if node_.Ipmi == nil || node_.Ipmi.Ipaddr.IsUnspecified() || node_.Ipmi.UserName == "" || node_.Ipmi.Password == "" {
+					wwlog.Debug("api.getPower() node ipmi not configured")
+					err = status.Wrap(fmt.Errorf("node ipmi not configured"), status.InvalidArgument)
+					powerResults[0].Err = err // Set the error in the first PowerResult
+					*output = powerResults    // Set *output on exit. important
+					return err
+				}
+
+				// starting with powerResults[0] as the accumulator
+				powerResults[0].BmcIp = node_.Ipmi.Ipaddr.String() // Set the BMC IP address in the PowerResult
+
+				// Setup ipmitool call.
+				ipmiCmd := bmc.TemplateStruct{
+					IpmiConf: *node_.Ipmi,
+				}
+
+				// Make ipmitool call.
+				result, err := ipmiCmd.PowerStatus()
+				if err == nil {
+					if result == "Chassis Power is on" {
+						powerResults[0].Result = "on"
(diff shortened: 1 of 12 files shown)


internal/pkg/warewulfd/api/power.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/swaggest/usecase/status"
+	"github.com/warewulf/warewulf/internal/pkg/bmc"
+	"github.com/warewulf/warewulf/internal/pkg/node"
+	"github.com/warewulf/warewulf/internal/pkg/wwlog"
+)
+
+// PowerResult is the result structure for single node power status operations. get/set power status
+type PowerResult struct {
+	NodeId string `yaml:"node_id" json:"node_id" description:"ID of the node for which power status is being reported"`
+	BmcIp  string `yaml:"bmc_ip" json:"bmc_ip" description:"BMC IP address of the node"`
```
anderbubble, line 19: `node_id` and `bmc_ip` aren't used in the cockpit PR. They should be removed here.

https://github.com/ctrliq/warewulf-cockpit/pull/16

internal/pkg/warewulfd/api/power.go
```diff
@@ -0,0 +1,200 @@
...
+	Result string `yaml:"result" json:"result" description:"Power status of the node, e.g., 'on', 'off', 'unknown'"`
+	Err    error  `yaml:"error,omitempty" json:"error,omitempty" description:"Error message if any error occurred while getting power status"`
```
anderbubble, line 21: Rather than returning an Error in the result, just actually return an HTTP error when there is an error.

internal/pkg/warewulfd/api/power.go
```diff
@@ -0,0 +1,200 @@
...
+// getPower returns an interactor to get the power status of a node by its ID.
+// At this time we ony support a single node. There is no expansion of the hostlist args or batching yet.
+func getPower() usecase.Interactor {
+	type getPowerInput struct {
+		ID string `path:"id" required:"true" description:"ID of node to get power status for"`
+	}
+
+	u := usecase.NewInteractor(func(ctx context.Context, input getPowerInput, output *[]PowerResult) error {
+		wwlog.Debug("api.getPower(), ID: %s", input.ID)
+		powerResults := []PowerResult{}
```
anderbubble, line 33: `getPower()` only ever returns one result for one node, so don't return a list.

internal/pkg/warewulfd/api/power.go
```diff
@@ -0,0 +1,200 @@
...
+					IpmiConf: *node_.Ipmi,
+				}
+
+				// Make ipmitool call.
+				result, err := ipmiCmd.PowerStatus()
+				if err == nil {
+					if result == "Chassis Power is on" {
+						powerResults[0].Result = "on"
+					} else if result == "Chassis Power is off" {
+						powerResults[0].Result = "off"
```
anderbubble, line 72: These result strings, "Chassis Power is on" and "Chassis Power is off" are, iirc, implementation-specific, and just happen to be what ipmitool returns. That said, I don't know that we _have_ good programmatic differentiation between whether it's "on" or "off" currently.

I suggest, for now, using a regex to see if the result string contains "on". Something like

```
regexp.MustCompile(`(?i)\bon\b`)
```

Then maybe PowerResult could contain the original sring in `Result`, and have a boolean `IsOn` for the parsed value.

internal/pkg/warewulfd/api/power.go
```diff
@@ -0,0 +1,200 @@
...
+		powerResults = append(powerResults, PowerResult{
+			NodeId: input.ID,
+		})
+
+		input.State = strings.ToLower(input.State)
+		if !slices.Contains(validPowerStates, input.State) {
+			err := status.Wrap(fmt.Errorf("invalid power state %s valid states are %v", input.State, validPowerStates), status.InvalidArgument)
+			powerResults[0].Err = err
+			*output = powerResults // Set *output on exit. important
+		}
```
anderbubble, line 130: Since you already process `input.State` with a switch statement, I don't think you need to validate it here.

internal/pkg/warewulfd/api/power.go
```diff
@@ -0,0 +1,200 @@
...
+			*output = powerResults // Set *output on exit. important
+			return err
+		} else {
+			if node_, err := registry.GetNode(input.ID); err != nil {
+				wwlog.Debug("api.setPower() node not found: %v", err)
+				powerResults[0].Err = err
+				*output = powerResults // Set *output on exit. important
+				return status.Wrap(fmt.Errorf("node not found: %v (%v)", input.ID, err), status.NotFound)
+			} else {
+				if node_.Ipmi == nil || node_.Ipmi.Ipaddr.IsUnspecified() || node_.Ipmi.UserName == "" || node_.Ipmi.Password == "" {
```
anderbubble, line 144: I don't think you should require UserName or Password to be set. For example, our sample nobmc template doesn't use them.

https://github.com/warewulf/warewulf/blob/main/lib/warewulf/bmc/nobmc.tmpl

internal/pkg/warewulfd/api/power.go
```diff
@@ -0,0 +1,200 @@
...
+					wwlog.Debug("api.getPower() node ipmi not configured")
+					err = status.Wrap(fmt.Errorf("node ipmi not configured"), status.InvalidArgument)
+					powerResults[0].Err = err
+					*output = powerResults // Set *output on exit. important
+					return err
+				}
+
+				powerResults[0].BmcIp = node_.Ipmi.Ipaddr.String() // Set the BMC IP address in the result now that we have it.
```
anderbubble, line 152: You're only returning one `powerResult`, so don't return a list.

internal/pkg/warewulfd/api/power_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+		body, err := io.ReadAll(resp.Body)
+		assert.NoError(t, err)
+		assert.NoError(t, resp.Body.Close())
+
+		assert.JSONEq(t, `{"kernel": {"version": "v1.0.0", "args": ["kernel-args"]}}`, string(body))
+	})
+
+	// add a new profile node000profile profile.
+	t.Run("add test testing1234OverOver profile", func(t *testing.T) {
```
anderbubble, line 55: Clean up this test name, making the profile name simpler (see other tests for examples), and remove the comment (which isn't necessary since you have the test name).

internal/pkg/warewulfd/api/power_test.go
```diff
@@ -0,0 +1,190 @@
...
+		req, err = http.NewRequest(http.MethodGet, srv.URL+"/api/power/ipminode", nil)
+		assert.NoError(t, err)
+
+		resp, err = http.DefaultTransport.RoundTrip(req)
+		assert.NoError(t, err)
+
+		_, err = io.ReadAll(resp.Body)
+		assert.NoError(t, err)
+		assert.NoError(t, resp.Body.Close())
+	})
```
anderbubble, line 97: Since you're not here to test adding nodes and profiles, just write the nodes and profiles you want to test into `nodes.conf` directly. See other tests for examples. (e.g., `internal/app/wwctl/node/list/main_test.go`)

internal/pkg/warewulfd/api/power_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/warewulf/warewulf/internal/pkg/testenv"
+	"github.com/warewulf/warewulf/internal/pkg/warewulfd"
+)
+
+func TestPowerAPI(t *testing.T) {
```
anderbubble, line 16: Try to add some successful tests. You'll likely need to inject a test-only bmc provider template.

anderbubble (changes requested):

PR #1953, opened 2025-07-18
siwtched to dnsmasq as default dhcp and tftp service

mslacken:
## Switch to dnsmasq
Configuration template is already there, also updated docs


anderbubble (changes requested):
On further reflection I think we need to discuss this again in TSC. CIQ already has customers complaining about rapid changes in Warewulf that aren't tied to a major release, and I think switching to dnsmasq by default in this kind of a release is a further violation of user expectations.

What if we started building EL10 / SLE16 packages, and specifically made dnsmasq the default specifically for those builds?

anderbubble:
I'm going to close this for now, since #1974 does this for EL10 builds.

PR #1955, opened 2025-07-22
test for disk vars set to false

mslacken:
The option `"--diskwipe=false` and other boolean disk option aren't working.


mslacken:
Replaced the bools in disk related structures to `WWBool` as bools get lost when set to false on the command line.
It has the notnice[*] side effect that
```
disks:
      /dev/vda:
        wipe_table: "false"
```
is equal to
```
disks:
      /dev/vda: {}
```


close #1956

[*] the german word I want to use here is "unschön", which could be translated in to "ugly", but as "ugly" can be avoided "unschön" means also that this is necessity which can't changed without a big effort

anderbubble:
Rebased to main.

anderbubble:
@mslacken can you review #1994 as an extension of this?

anderbubble:
Marking as draft while #1994 is being reviewed.

PR #1957, opened 2025-07-25
added warewuld configure option

mslacken:
When installing warewulf and you have to run `wwctl configure -a` which will wanr as warewulfd isn't enabled and started with this command.
This PR fixes this.
Pure convinience, which you will only notic if you are doing a lot of testing.


anderbubble:
I like it!

I made a few changes:

- fixed a bug where the new warewulfd command was actually calling the dhcp setup routine
- Removed the default value for the warewulf sytemd name, so that blanking it out can be used to disable the feature
- Added it to `wwctl upgrade`

PR #1958, opened 2025-07-28
Add IPv6 support for iPXE booting

middelkoopt:

Add preliminary/partial support for booting nodes via IPv6 PXE. This PR does not provide full IPv6 provisioning support, however, I believe the only thing missing is DHCP server support.  

Tested with Rocky9 on aarch64 on MacOS with QEMU.  Rocky9 iPXE did not work with IPv6, building iPXE from git fixed it.

Detailed Changes:
* This PR is based off of #1852 (rebased)
* Introduces a new provisioning (iPXE) template variable `Authority` which combines the IP address and port and provides the `[]` if the address is IPv6 according to RFC 3986.
* The value of Authority is set to the same address family as the requesting IP.  This can be used in a template regardless of address family.
* Unit tests added for parsing and Authority.
* default.ipxe has been updated and tested with Authority.
* Now uses netip address parsing for the requesting IP address (go 1.18+)
* Reverts PR #1852 config parser changes in favor of current version.
  * Note: `warewulf.conf` `Ipadd6` requires a subnet prefix and stored as such.
  * Note: If a subnet prefix is set in `warewulf.conf` `Ipaddr`, (IPv4) it is removed and the netmask and broadcast addresses are set from this.
  * The conf.Ipaddr6 is stored as a IPv6 with prefix, #1852 set `warewulf.conf` `Ipaddr6` value to the address without the prefix.
  * The provisioning template (iPXE) `IPaddr` is set to just the IPv6 value, same as #1852

- Rework of #1852

## Reviewer  checklist

The original PR will have to be signed off.

### DIFF
--- a/internal/pkg/warewulfd/provision.go
+++ b/internal/pkg/warewulfd/provision.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/netip"
 	"path"
 	"path/filepath"
 	"strconv"
@@ -32,7 +33,9 @@ type templateVars struct {
 	Ipxe          string
 	Hwaddr        string
 	Ipaddr        string
+	Ipaddr6       string
 	Port          string
+	Authority     string
 	KernelArgs    string
 	KernelVersion string
 	Root          string
@@ -115,12 +118,30 @@ func ProvisionSend(w http.ResponseWriter, req *http.Request) {
 				kernelVersion = kernel_.Version()
 			}
 		}
+		authority := fmt.Sprintf("%s:%d", conf.Ipaddr, conf.Warewulf.Port)
+		ipaddr6 := ""
+		if confIpaddr6, err := netip.ParsePrefix(conf.Ipaddr6); err == nil {
+			ipaddr6 = confIpaddr6.Addr().String()
+		}
+		if rinfoIpaddr, err := netip.ParseAddr(rinfo.ipaddr); err == nil {
+			if rinfoIpaddr.Is6() {
+				if ipaddr6 != "" {
+					authority = fmt.Sprintf("[%s]:%d", ipaddr6, conf.Warewulf.Port)
+				} else {
+					wwlog.Error("No valid IPv6 address configured, but request is IPv6")
+				}
+			}
+		} else {
+			wwlog.Error("Could not parse request IP address: %s", rinfo.ipaddr)
+		}
 		tmpl_data = &templateVars{
 			Id:            remoteNode.Id(),
 			Cluster:       remoteNode.ClusterName,
 			Fqdn:          remoteNode.Id(),
 			Ipaddr:        conf.Ipaddr,
+			Ipaddr6:       ipaddr6,
 			Port:          strconv.Itoa(conf.Warewulf.Port),
+			Authority:     authority,
 			Hostname:      remoteNode.Id(),
 			Hwaddr:        rinfo.hwaddr,
 			ImageName:     remoteNode.ImageName,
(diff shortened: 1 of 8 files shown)


internal/pkg/warewulfd/provision.go
```diff
@@ -115,12 +118,30 @@ func ProvisionSend(w http.ResponseWriter, req *http.Request) {
...
 			}
 		}
+		authority := fmt.Sprintf("%s:%d", conf.Ipaddr, conf.Warewulf.Port)
+		ipaddr6 := ""
+		if confIpaddr6, err := netip.ParsePrefix(conf.Ipaddr6); err == nil {
+			ipaddr6 = confIpaddr6.Addr().String()
+		}
+		if rinfoIpaddr, err := netip.ParseAddr(rinfo.ipaddr); err == nil {
+			if rinfoIpaddr.Is6() {
+				if conf.Ipaddr6 != "" {
```
mslacken: Better use `ipaddr6 != ""` as this would also catch an invalid configure ipv6 address.

internal/pkg/warewulfd/provision.go
```diff
@@ -203,12 +224,30 @@ func ProvisionSend(w http.ResponseWriter, req *http.Request) {
...
 				}
 			}
+			authority := fmt.Sprintf("%s:%d", conf.Ipaddr, conf.Warewulf.Port)
+			ipaddr6 := ""
+			if confIpaddr6, err := netip.ParsePrefix(conf.Ipaddr6); err == nil {
+				ipaddr6 = confIpaddr6.Addr().String()
+			}
+			if rinfoIpaddr, err := netip.ParseAddr(rinfo.ipaddr); err == nil {
+				if rinfoIpaddr.Is6() {
+					if conf.Ipaddr6 != "" {
```
mslacken: Better use ipaddr6 != "" as this would also catch an invalid configure ipv6 address.

mslacken (dismissed):

internal/pkg/warewulfd/provision.go
middelkoopt: Don't quite parse what you mean.  Do you mean put a `ipaddr6 != ""` in Line 126 to short circuit the if statement?

mslacken: No in line 128, as if the current logic we can have an invalid ipv6 address in the configuration, although we checked it's validity in 123. There we make sure that `ipaddr6` is either empty or a valid ipv6 address.

middelkoopt: Good catch.  Fixed along with the later occurrence.

middelkoopt:
`grub.cfg.ww` updated.  This should be a NOP for the rendered `grub.cfg` ipv6 provisioning not tested.

middelkoopt:
rebased against main.

middelkoopt:
passes my virtual cluster tests.

middelkoopt:
Ready for merge.

middelkoopt:
rebased against main.

anderbubble (approved):

anderbubble:
GRUB boot is still inconsistent / unreliable, but that's not new to this. When it works, it works. When it doesn't, it complains about a tcp timeout.

This looks good, though!

dacianstremtan:
I'm adding my dhcpd.conf and dhcpd6.conf files here used for booting. The legacy dhcpd.conf is needed because the ipxe environment does not support DHCP6, only SLAAC. The latter does not provide a bootfile-url option.

The booting process is like this:
The network stack starts IPv6 booting. This stack supports DHPCv6 and downlaods the snponly.efi boot file from tftp via IPv6. The efi environments starts and configures the IPv4 via DHCP and IPv6 via SLAAC and RA. The dhcp service stagess the boot file which points to the node's image file on the IPv6 network. The file is downloaded because of a valid IPv6 configured via SLAAC and RA.

[dhcpd6.conf.txt](https://github.com/user-attachments/files/21580252/dhcpd6.conf.txt)
[dhcpd.conf.txt](https://github.com/user-attachments/files/21580254/dhcpd.conf.txt)

anderbubble:
@middelkoopt I'll leave it to you do determine whether and how these should be incorporated into a follow-up PR.

dacianstremtan:
I don't know how these configs get generated by the code.
Dacian

În lun., 4 aug. 2025 la 11:32, Jonathon Anderson ***@***.***>
a scris:

middelkoopt:
Thanks for the files @dacianstremtan and your contribution.

For the reason, I was looking for more information about the implementation, but I figured out enough from my own environment.  I'll use these as reference as I (slowly) build out full IPv6 support.  However, I'll probably leverage `dnsmasq` given that ISC dhcpd is EOL and not on EL10.

PR #1968, opened 2025-08-04
refactored overlay class

mslacken:
overlay.GetOverlay(name) returns now an error if the overlay doesn't exist.
This is the most canonical way to act if there is no overlay.

Also moved the functional code for chmod, chown and delete to the overlay class and not to wwctl so that theses functions can be called with class.


anderbubble:
@mslacken I added an overlay.Delete() method split off from overlay.DeleteFile(), and fixed up the REST api to properly handle not deleting a distribution overlay. (And I added a test for it.)

Beyond that, I addressed the review from copilot and simplified some of the method names based on my understanding of golang recommendations. But I don't think you'd object to any of it, so I'll go ahead and merge once the tests pass this last time.

Thanks for doing this!

Issue #1975, opened 2025-08-08
slurm job fails to execute srun whilst the runtime overlay for /etc/passwd is being updated


metalllinux:
### Steps to reproduce

1. Set up a Warewulf v4.6.1 cluster with `slurm`.
2. Run a `slurm` job whilst the runtime overlay for `/etc/passwd` is updated.
3. If timed correctly, the `slurm` job will fail.

### Error message

The error that `slurm` provides when the job fails:
```
srun: error: Invalid user for SlurmUser slurm, ignored
srun: fatal: Unable to process configuration file
```

It is believed this is due to a race condition with `cpio` writing to `/etc/passwd` and the `slum` job running at the same time. Thus `slurm` is not able to find the `slurm` user when the job is being ran.

### Information on your system

```
wwctl version:   4.6.1-1
rpc version: apiPrefix:"rc1"  apiVersion:"1"  warewulfVersion:"4.6.1-1"
```

```
NAME="Rocky Linux"
VERSION="8.10 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.10"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.10 (Green Obsidian)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:8:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
SUPPORT_END="2029-05-31"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-8"
ROCKY_SUPPORT_PRODUCT_VERSION="8.10"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.10"
```

Available workarounds:

1. Lower the frequency in which `wwclient` updates via `warewulf.conf`.
2. Remove the `Syncuser` overlay / the overlay you are using and replace that with a system overlay, so that `/etc/passwd` is only updated on boot.

### General information

- [x] I have run `wwctl version` and reported the contents of `/etc/os-release`
- [x] I have searched the issues of this repo and believe this is not a duplicate
- [x] I have captured and reported relevant error messages and logs

Issue #1307, opened 2024-07-18
Runtime overlay placement causes files to be absent


epruesse:
### Steps to reproduce

- Symlinks in Warewulf runtime overlay fixing some legacy paths
- Snakemake pipeline accessing data through those symlinks

This leads to somewhere around 1000 stat calls per second on files within those symlinks

### Error message

The Snakemake pipeline aborts eventually because it cannot find one of the files. My only explanation since it doesn't happen on the production systems is that the runtime overlay for the most tiny amount of time removes a file before recreating it. This creates a race condition my test pipeline eventually hits. 

### Information on your system

Warewulf 4.5.4, Rocky 9.4

### General information

- [x] I have run `wwctl version` and reported the contents of `/etc/os-release`
- [x] I have searched the issues of this repo and believe this is not a duplicate
- [x] I have captured and reported relevant error messages and logs

anderbubble:
Thanks for the report, @epruesse! If you're able to confirm specifically that files do go absent, please do share that procedure here. Right now my understanding of how overlays are unpacked wouldn't explain that behavior; but I've taken a note that we should investigate.

griznog:
This is how `wwclient` unpacks the images, I think:

https://github.com/warewulf/warewulf/blob/b61e3918908bd48ee7bb30af659fe946cc06efac/internal/app/wwclient/root.go#L242

@anderbubble do we really need the `-u` (`--unconditional`) flag there? Without `-u` only changed/newer files would get replaced. 

```
griznog@micropterus:~$ ls -al testing
total 46
drwxr-xr-x   2 griznog griznog   4 Jul 19 12:09 .
drwx------. 84 griznog griznog 128 Jul 19 12:07 ..
lrwxrwxrwx   1 griznog griznog   8 Jul 19 12:09 linktotest -> testfile
-rw-r--r--   1 griznog griznog  15 Jul 19 12:05 testfile
griznog@micropterus:~$ find testing | cpio -ov > testing.img
testing
testing/testfile
testing/linktotest
1 block
griznog@micropterus:~$ cpio -iduv < testing.img 
testing
testing/testfile
testing/linktotest
1 block
griznog@micropterus:~$ cpio -idv < testing.img 
testing
cpio: testing/testfile not created: newer or same age version exists
testing/testfile
cpio: testing/linktotest not created: newer or same age version exists
testing/linktotest
1 block
```

griznog:
[...]

Issue #1226, opened 2024-05-23
wwclient --once


anderbubble:
### Summary

I suggest there should be an option to wwclient to allow it to be run once.

### Rationale

This way wwclient could be run manually.

### General information

- [X] I have searched the issues of this repo and believe this is not a duplicate

PR #1984, opened 2025-08-13
Atomic overlay file application in wwclient

anderbubble:
Re-submitting after the previous merge and subsequent revert.

Update wwclient such that each file in the runtime overlay is applied atomically.

Also adds --once flag to allow wwclient to be run one time.

- Closes: #1307
- Closes: #1975
- Closes: #1226

## Reviewer  checklist

### DIFF
--- a/internal/app/wwclient/root.go
+++ b/internal/app/wwclient/root.go
@@ -1,8 +1,8 @@
 package wwclient
 
 import (
-	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/url"
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
@@ -32,6 +33,7 @@ var (
 		SilenceUsage: true,
 		Args:         cobra.NoArgs,
 	}
+	Once            bool
 	DebugFlag       bool
 	PIDFile         string
 	Webclient       *http.Client
@@ -39,6 +41,7 @@ var (
 )
 
 func init() {
+	rootCmd.PersistentFlags().BoolVar(&Once, "once", false, "Run once and exit")
 	rootCmd.PersistentFlags().BoolVarP(&DebugFlag, "debug", "d", false, "Run with debugging messages enabled.")
 	rootCmd.PersistentFlags().StringVarP(&PIDFile, "pidfile", "p", "/var/run/wwclient.pid", "PIDFile to use")
 	rootCmd.PersistentFlags().StringVar(&WarewulfConfArg, "warewulfconf", "", "Set the warewulf configuration file")
@@ -52,6 +55,12 @@ func GetRootCommand() *cobra.Command {
 }
 
 func CobraRunE(cmd *cobra.Command, args []string) (err error) {
+	if DebugFlag {
+		wwlog.SetLogLevel(wwlog.DEBUG)
+	} else {
+		wwlog.SetLogLevel(wwlog.INFO)
+	}
+
 	conf := warewulfconf.Get()
 	if WarewulfConfArg != "" {
 		err = conf.Read(WarewulfConfArg, false)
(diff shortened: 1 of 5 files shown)


mslacken:
@anderbubble I am just tossing around the idea, that `wwclient` does store the checksum of the downloaded runtime overlay and only  downloads the overlay if the checksum has changed and if the checksum is the same to do nothing at all.
Also `wwclient` could keep a list of the files in the runtime overlay and apply the overlay only if something has changed on the node side.
This would solve the problems and also greatly reduce network traffic.

anderbubble:
@mslacken that would _drastically reduce_ the likelihood of the issue, but there would still be a race condition when it _does_ update.

That would also change behavior: right now, if you change a file on the compute node, the overlay is re-applied and puts it back to the intended configuration. If we only re-apply overlays when they change, that would stop.

mslacken:
Please rebase on #1985 so that when testing `wwclient` behavior we also get checksum handling.
Buy one test, get one feature for free!

anderbubble:
I updated the PR to properly set permissions and selinux context.

anderbubble:
Regarding rsync:

- I do not think we should introduce rsync as a transfer mechanism. Doing so would depend on ssh or an rsyncd, bypassing warewulfd. At the very least, I don't think now is the time to make such a change: that's an architectural issue, and right now I primarily want to fix a bug.
- I considered using rsync as a synchronization mechanism, but I don't want to add a dependency on rsync within the node image. (In fact, I'd prefer it if we could remove the dependency on cpio and gzip; but that's a different conversation.)

anderbubble:
@mslacken @middelkoopt @jcsiadal please review this PR to make each file change made by wwclient atomic.

While this issue has been reported before (#1307), this specific PR is motivated by a customer who has experienced slurm errors due to a race condition where slurm tries to access /etc/passwd while it is being re-written. (#1975) By copying each file to a temporary file first, and then replacing the original file with a move, the view of the file at the target path is now always either the complete original file or the complete new file. The temporary file is written to the same directory as the target file to ensure that it is on the same file system. Otherwise, a non-atomic copy would still be required.

I recognize that this is a relatively large change to wwclient, and that we don't currently have good test coverage for wwclient. I have tested it manually in a test environment both with and without SELinux, and confirmed both file content and metadata. SELinux contexts, ownership, and permissions appear to be handled correctly.

To aid in manual testing and debugging, I have also improved the debugging experience in wwclient. I fixed a bug in pidfile handling, added verbose debug logging to the existing `--debug` flag, and added a `--once` flag to make it possible to run wwclient once. (#1226)

@mslacken I do agree with the intent of #1985, but I think the implementation needs to align to an existing HTTP standard practice, so I prefer to not wait for it to be done before resolving this issue. I am also committed to re-review your pending PRs this week, regardless of the timeline for this being merged.

And @mslacken thank you for pushing back on #1982. I didn't feel good about the way we merged that regardless, and there _were_ severe bugs: it was mishandling regular file permissions, _and_ it didn't do anything to support SELinux contexts.

internal/app/wwclient/root.go
```diff
@@ -246,13 +288,200 @@ func updateSystem(ipaddr string, port int, wwid string, tag string, localUUID uu
...
+			// This ensures the temp file is on the same filesystem as the final
+			// destination, so the rename placement is atomic.
+			destParent := filepath.Dir(destPath)
+
+			// Ensure destination directory exists
+			if err := os.MkdirAll(destParent, 0755); err != nil {
+				return fmt.Errorf("failed to create destination directory %s: %w", destParent, err)
+			}
+
+			tempFile, err := os.CreateTemp(destParent, ".wwclient-tmp-")
```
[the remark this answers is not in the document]
anderbubble: Duplication is accepted for now, as I'd like to retain the ability to have different patterns for files and symlinks later if preferred. That and we're creating symlink files differently now, anywah, which would make it awkward to share.

internal/app/wwclient/root.go
```diff
@@ -64,33 +73,29 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
...
+	target := "/"
 	if os.Args[0] == path.Join(conf.Paths.WWClientdir, "wwclient") {
-		err := os.Chdir("/")
-		if err != nil {
-			return fmt.Errorf("failed to change dir: %w", err)
-		}
 		wwlog.Warn("updating live file system: cancel now if this is in error")
 		time.Sleep(5000 * time.Millisecond)
 	} else {
+		target = "/warewulf/wwclient-test"
```
middelkoopt, line 90: I wonder if this should be an option, not hard coded.  If for some reason wwclient is not put in /warewulf in production this will get triggered.

middelkoopt: If not, I think the warning message is sufficient to debug by someone attempting to do this.

anderbubble: It's at least not new behavior for it to go to `/warewulf/wwclient-test`. I'm open to having that be configurable, but I'd like to handle it as a separate PR if possible.

anderbubble:
@middelkoopt I added a check for if the file has changed. I just look at file size and mtime because I don't want to introduce unnecessary CPU jitter by calculating hashes. Let me know what you think. If it looks good to you, I'd like to move towards merge.

anderbubble:
I just pushed another small change: it now compares if the src and dest mtimes are equal, rather than if the source is newer than dest. This way it retains the previous behavior of overwriting changes that may occur on the cluster node out-of-band from wwclient.

middelkoopt:
> I just pushed another small change: it now compares if the src and dest mtimes are equal, rather than if the source is newer than dest. This way it retains the previous behavior of overwriting changes that may occur on the cluster node out-of-band from wwclient.

The lightweight behavior sounds fine.  The speed of light is not fast enough to create a race condition. QED.

I believe if the permissions change (POSIX) of the file will not be updated if mtime/size stay the same (aka a chmod). (selinux attributes are restored not copied, so that should be ok).

anderbubble:
@middelkoopt I've added a check on file mode as well. Thanks!

PR #1985, opened 2025-08-14
Non atomic

mslacken:
Add checksum check before downloading the overlay

### DIFF
--- a/internal/app/wwclient/root.go
+++ b/internal/app/wwclient/root.go
@@ -1,8 +1,10 @@
 package wwclient
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


internal/pkg/warewulfd/provision.go
```diff
@@ -193,6 +197,45 @@ func ProvisionSend(w http.ResponseWriter, req *http.Request) {
 			wwlog.ErrorExc(err, "")
 			return
 		}
+	} else if rinfo.stage == "runtime_check" {
```
anderbubble, line 200: Rather than introduce a new stage, warewulfd should provide ETag or Last-Modified headers, and then support If-None-Match or If-Modified-Since request headers. If it doesn't already, we should also support fetcheing just headers with a HEAD request. This could be implemented just for the runtime stage, but the same approach would also apply to all stages if we wanted it.

anderbubble (changes requested):

internal/app/wwclient/root.go
```diff
@@ -210,9 +214,59 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
...
+			if err != nil {
+				wwlog.Error("could not read checksum from response: %s", err)
+				resp.Body.Close()
+				return []byte{}
+			}
+			resp.Body.Close()
+			if bytes.Equal(bytes.TrimSpace(remoteSum), currentSum) {
+				wwlog.Info("runtime overlay is current")
+				return currentSum
+			}
```
anderbubble, line 265: There should be a way to tell wwclient to ignore this, so that people can retain the current behavior if they want it. An environment variable would probably be good, though a command-line argument would probably be fine, too.

PR #1992, opened 2025-08-31
Follow the recommended dracut module ordering for out-of-tree modules.

jozzsi:
Move module ordering from 90 to 50.

 For dracut release v108 or later the recommended ordering for out out of tree modules
is 50-59 range. The following is a section from dracut documentation:

> Not using the 50-59 range for out of tree dracut modules will likely
> lead to unintended errors in the initramfs generation process as your
> dracut module will either run too early or too late in the generation process.
> You have been warned.


anderbubble:
I think we shouldn't rush to get this into v4.6.4, but I'm marking int for v4.6.x to get some more review.

jozzsi:
I do not intent to rebase this PR any more. Closing.

For reference - https://github.com/flatcar/bootengine/pull/109#issuecomment-3723591773

middelkoopt:
@jozzsi thanks for this, I've created issue #2084 so we don't loose track of this.

PR #1994, opened 2025-09-03
Fix disk-related booleans using a *bool to capture unset values

anderbubble:

Extends #1955 using `*bool` values rather than wwbool values. This allows undefined to be distinguished from "false" values using a nil pointer.

## Reviewer  checklist


anderbubble:
I don't understand why the `make fmt` check is failing, cause it's not in my environment. But I'll try to figure it out.

anderbubble:
Found it. There was a tiny formatting issue in buildconfig.go.in, which was fixed already in my side in buildconfig.go, but since that file isn't in git, the fix didn't make it into git.

I'll leave the testsuite change in place so that this is easier to diagnose in the future.

anderbubble:
@middelkoopt since @mslacken is still away, would you have time to review this?

.github/workflows/testsuite.yml
```diff
@@ -42,7 +42,9 @@ jobs:
       - name: Vet Warewulf code (golang ${{ matrix.go-version }})
         run: make vet
       - name: Check Warewulf code for formatting (golang ${{ matrix.go-version }})
-        run: test "$(make -s fmt | wc -l)" == 0
+        run: |
+          make -s fmt | tee /tmp/fmt-output
+          test "$(wc -l </tmp/fmt-output)" == 0
       - name: Check for dead Warewulf code (golang ${{ matrix.go-version }})
```
mslacken, line 48: What's wrong with single pipe. Just asking out of curiosity? Also the file isn't removed and mktemp would better here.

internal/pkg/node/datastructure.go
```diff
@@ -102,29 +103,52 @@ partition definition, the label must be uniq so its used as the key in the
...
+	Path            string   `yaml:"path,omitempty"            json:"path,omitempty"            lopt:"fspath"   comment:"the mount point of the file system"`
+	WipeFileSystemP *bool    `yaml:"wipe_filesystem,omitempty" json:"wipe_filesystem,omitempty" lopt:"fswipe"   comment:"wipe file system at boot"`
+	Label           string   `yaml:"label,omitempty"           json:"label,omitempty"                           comment:"the label of the filesystem"`
+	Uuid            string   `yaml:"uuid,omitempty"            json:"uuid,omitempty"                            comment:"the uuid of the filesystem"`
+	Options         []string `yaml:"options,omitempty"         json:"options,omitempty"                         comment:"any additional options to be passed to the format-specific mkfs utility"`
+	MountOptions    string   `yaml:"mount_options,omitempty"   json:"mount_options,omitempty"                   comment:"any special options to be passed to the mount command"`
 }
 
 type Resource interface{}
+
```
mslacken, line 132: Are the  following methods used at all?

mslacken (commented):
Will test this a bit more, but looks good.

.github/workflows/testsuite.yml
anderbubble: I split it out so that the output from `make -s fmt` would show in the GitHub workflow log. Otherwise when the test fails it doesn't tell you why. (This was an issue I was having, so I left it in place so that it would provide better logging in the future.)

As for not cleaning out `/tmp`, in theory you're right, but it's just a GitHub workflow, so the whole thing gets removed at the end anyway.

internal/pkg/node/datastructure.go
anderbubble: These methods are primarily for use by overlay templates. Changing the struct field to a pointer would mean that overlay templates have to deal with null values; so the fields are renamed (to a value with a P suffix, meaning "Pointer") and then overlay templates can continue to use the previous names to call these methods which return simple booleans that will always be either true or false.

.github/workflows/testsuite.yml
anderbubble: From a Copilot review:

> The temporary file /tmp/fmt-output could conflict in parallel builds.

Maybe I was wrong and GitHub might actually run things in parallel here. I'll fix it.

anderbubble: Done.

PR #2002, opened 2025-09-05
Read field names from field tags

anderbubble:

Improves output for boolean-pointer fields.

## Reviewer  checklist

### DIFF
--- a/internal/pkg/node/datastructure.go
+++ b/internal/pkg/node/datastructure.go
@@ -94,7 +94,7 @@ Holds the disks of a node
 */
 type Disk struct {
 	id         string                `yaml:"-"                    json:"-"`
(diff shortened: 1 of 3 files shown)


middelkoopt (approved):
Looks good.

mslacken:
The question here is, hey not use the same name as the name for the command line option?
We have now three different names for a variable:
* the real name in the go code
* it's yaml representation
* the representation in the command line
and now an output name.
I think we should try to reduce the names.

middelkoopt:
I had similar thoughts.

anderbubble:
> [why] not use the same name as the name for the command line option

@mslacken I absolutely agree, and I think that we should consolidate the YAML and the CLI output names (and probably make the json names always be the same, as well); but for now, in this release, I wanted to maintain the behavior people are accustomed to. I think it could be appropriate to consolidate this in v4.7.0, or definitely in v5.0.0.

PR #2013, opened 2025-09-16
Change netplan overlay to use netdev device names instead of the netdev name in bonds

tweiers:

Currently the netplan overlay assumes that the device name and netdevs name of a network device are the same when creating bonds. 
This can cause issues when the netdev name differences from the device name causing the overlay to skip adding the device to the bond. 

For example. 
A node has the following set: 
   NetDevs[default].Device: bond0

   NetDevs[eth0].Device: eno12399np0
   NetDevs[eth0].Tags[master]: bond0

   NetDevs[eth1].Device: eno12409np1
   NetDevs[eth1].Tags[master]: bond0 

Before the if statement would check if the master tag matches the netdev name of the bond. 
In this case that would fail as the device name is bond0 but the netdev name is default. 
This would cause the interfaces to never get added. 

Secondly when check was changed to look at device names instead the interace that would be added would be the netdev name instead of the device name. 
This could cause issues with eth0 and eth1 getting added as interfaces but those interfaces don't exist and instead should be the netdev.device names eno12399np0 and eno12409np1.

## Reviewer  checklist


iankgt40:
This fix needs to go into 01-netcfg.yaml, and I have tested it in the correct file and it does resolve this issue without affecting other functionality.

anderbubble:
@iankgt40 I don't see a `01-netcfg.yaml` in the overlay. Are you saying that it should be introduced in a _new_ file? How do they differ? (I don't know much about netplan.)

iankgt40:
Ah, 01-netcfg.yaml was the default config I was using based on netplan docs and testing/developing ... since I left it there, the "01" made it take precedence over the warewulf.yaml file in the actual lreased code. I tested remove 01-netcfg.yaml, rebuilt the overlays, and rebooted the node with everything built as expected.

So, ignore that part ... the patch tests correctly and performs as expected.Sorry I missed yesterdays meeting ... one of those few emergency Zooms ... I'm being pulled in so many directions, and the beginning of the school year is now. I listened to the recording, and I agree with Tim as far as getting building the deb into the Github pipeline (thogh I am not sure exactly how to do it). I'm not as keen as releasing it as a binary package though ... I think the dependeincies need to be verified and installed ...

Ian

Ian Kaufman
Principal Systems Integration Engineer
UC San Diego, Research IT Services
ikaufman AT ucsd DOT edu
________________________________

anderbubble:
I updated the changelog and added testing for this case to the test suite. Just waiting for the tests to complete now.

PR #2017, opened 2025-09-19
Allow setting mtu and mac addr on bonds

tweiers:

Currently NetworkManager overlay only lets users apply mtu and mac addresses to slave nics. 
This change adds the ability for bonds to have their mac and mtu. 

Reason for the change:
I like setting the mtu at the bond level and have the properties propagate down to the slaves. Will also remove the number of places mtu needs to be set on in a nodes properties. 

Known issues: 
infiniband and ethernet are consider a single type of bond. 
This will cause issues when trying to set the mtu of a infiniband bond. 
Instead of the mtu being set in the [infiniband] section it will instead be set in the [ethernet] section. 

- Fixes #

## Reviewer  checklist


mslacken:
Thanks for fixing this.
Can you please add a test for this to `overlays/NetworkManager/internal/networkmanager_test.go` and an entry to CHANGELOG.md`?

tweiers:
> Can you please add a test for this to `overlays/NetworkManager/internal/networkmanager_test.go`

I can give it a try. I'm not the best go programmer but I'll take a look at the other tests and see what I can do. 

>an entry to CHANGELOG.md`?

I didn't update the changelog.md as that seemed like a release coordinator's thing to do when they decide what patches go into the next release.

mslacken:
Sorry closed this accidentally.
@tweiers The tests in https://github.com/warewulf/warewulf/blob/main/overlays/NetworkManager/internal/networkmanager_test.go are just a struct which takes as input a minimal `nodes.conf` and then the expected template. I will add and also drop a line in CHANGELOG.md.
Thanks again for taking time to improve warewulf

mslacken:
I added a test and failed as I don't know how a bonded infiniband network manager configuration must look like. You can see at #2019  how the tests work, unfortunately may test fails, but the best is if you add a test for the configuration you have.
You can run the tests with `make test`

tweiers:
@mslacken I've added in the new test to the test suite as well as updated the changelog. 
Can you take a look at let me know if anything else is missing?

mslacken:
@middelkoopt Can you please check. Other tests are also failing, but I don't know if this is critical or not. I have no in-depth experience with network bonding.

middelkoopt:
Checking, I'll try to get through it today - I'm behind from being out for two weeks.

tweiers:
I did notice a lot of the NetworkManager tests where failing when I was running the test suite. I assumed most of it was caused by newlines not lining up correctly in the templates.

mslacken:
The problem is not the newlines but for 
```
--- FAIL: Test_networkmanagerOverlay/NetworkManager:ww4-managed.ww_with_bond 
```
where there is 
```
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -17,2 +17,3 @@
            	            	 updelay=0
            	            	+[ethernet]
```
so there is an (empty) `[ethernet]` section for the bond is added. This could be right but feels wrong for me.

tweiers:
apologizes @middelkoopt  and @mslacken  the failings were from me. I forgot to go back and fix up the other tests. 
What managing white space is hard. 

I've added a new line separator from [connection] section and the [ethernet] section as before they were stuck together. 
There shouldn't be any functional difference just looks prettier to my human eyes.

overlays/NetworkManager/internal/networkmanager_test.go
```diff
@@ -337,6 +342,9 @@ mode=balance-rr
 xmit_hash_policy=layer2+3
 updelay=0
 
+
+[ethernet]
```
mslacken: This is the culprit I am hesitating. We have here a stray [ethernet] line for a bond. @anderbubble whats your take?

tweiers: Ah yes we will get extra [ethernet] sections on any ethernet or bond nic as well since the logic defaults to ethernet and writing the [ethernet] section even if its blank. 
This would happen with the old logic before as well if you didn't include a mtu or hwaddr. 
The same would happen to the [infiniband] section but that has a hardcoded entry for `transport-mode=datagram` so it doesn't look as out of place

anderbubble: The problem here, in my mind, is that we don't actually know if the bond is a bond of ethernet devices or infiniband devices. If it's a bond of infiniband devices, then these settings would need to go in the `[infiniband]` section, not in the `[ethernet]` section. But since Warewulf doesn't really know (without _a lot_ more work) it can't know it's putting these settings in the right place.

So the issue isn't so much that there's an empty `[ethernet]` section, but that there's always an `[ethernet]` section, even if it's not an ethernet bond.

This is why I continue to think that this is more appropriately done by specifying these settings on the underlying interfaces, rather than specifying this on the bond. While this does appear to work in specific circumstances, it does not appear to be how NetworkManager expects to work, and it's introducing assumptions that may not be accurate.

tweiers: I'd say this is expected behavior from the existing template as well. 
If you put in a minimally configured nic like 
```
    eth-test:
      device: eno01
      ipaddr: 10.0.0.1
```
Then a empty `[ethernet]` section is  created as well. 
It could be fixed in the if statement checking for if any of the properties that go into the ethernet section are present but I don't know who long we want to make the chain of if/else's currently it would only be 2 so no the worst. 

Just need to find a way that doesn't break old behaviour and without adding any new properties onto the netdev section. 
An example would be 
```
{{- else if and  (if or (eq (default "ethernet" (lower $netdev.Type)) "ethernet") (eq (lower $netdev.Type) "bond")) ( if or ( $netdev.MTU) ($netdev.Hwaddr) ) }}
```
It gets messy quickly.

middelkoopt: > It gets messy quickly.

Remember, readability and maintainability is also important.

middelkoopt: IB Bonding could also be a separate overlay if things git too messy.

tweiers:
> What you want is `cloned-mac-address`, which does appear to work.

OK, I think that sways me to make bonds its own section instead of piling it onto the ethernet type nics.
I'll rework the template so that there are 3 if's in place. 
1. infiniband 
2. bonds 
3. ethernet

This way bonds can use the cloned-mac-address setting instead.

anderbubble:
> This way bonds can use the cloned-mac-address setting instead.

To be clear, `cloned-mac-address` applies to regular interfaces, too, if you wanted to change the effective mac address of an interface.

anderbubble (approved):
@tweiers I think I've got this fixed up to behave as desired, but since I've made some notable changes I'll wait for your or another developer's thoughts before I merge. Let me know what you think.

PR #2043, opened 2025-10-16
Add xattr saving and restoration to images

### LINKED ISSUE #730: file caps lost when booting host
### Version of Warewulf
```
$ wwctl version
```
### Expected behavior

Expected that file caps are preserved from what is produced in a docker image to when the host boots

### Actual behavior

file caps are lost at node boot time

- set a cap on a file in a docker image
- import image into warewulf container
- getcap /usr/local/var/warewulf/chroots/${container_name}/path/to/file (you'll see the caps still there
- boot node using this container with tmpfs
- getcap /path/to/file (you'll see they are gone)

Doesnt matter its to do with the moving ramfs to tmpfs using tar
I'd suspect its due to tar not creating the archive with xattr nor extracting it with xattr

snehring:

This PR adds a way to preserve file xattrs (acls, file capabilities, selinux labels) of files in images and overlays. It does this by searching the root of the image/overlay at creation time and recording the xattrs of all files in an unique to each image/overlay file staged in /warewulf/xattrs of the image/overlay. A new init.d script will read these files and apply the xattrs in them and delete them.  When the runtime overlays are updated by wwclient it will restore any xattrs present in the /warewulf/xattrs directory.

Need to implement the getfattr during image creation and setfattr in runtime updates to native golang implementations.

WIP obviously

- Fixes #730 (potentially)

## Reviewer  checklist

### DIFF
--- a/internal/app/wwclient/root.go
+++ b/internal/app/wwclient/root.go
@@ -23,6 +23,7 @@ import (
 	warewulfconf "github.com/warewulf/warewulf/internal/pkg/config"
 	"github.com/warewulf/warewulf/internal/pkg/pidfile"
 	"github.com/warewulf/warewulf/internal/pkg/version"
+	"github.com/warewulf/warewulf/internal/pkg/util"
 	"github.com/warewulf/warewulf/internal/pkg/wwlog"
 )
 
@@ -329,6 +330,25 @@ func updateSystem(target string, ipaddr string, port int, wwid string, tag strin
 		wwlog.Error("failed running cpio: %s", err)
 		return
 	}
+	// apply saved xattrs
+	xattrsDir := filepath.Join(tempDir, "warewulf", "xattrs")
+	err = filepath.WalkDir(xattrsDir, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.Type().IsRegular() {
+			err = util.SetXattrsFromFile(tempDir, path)
+		}
+		if err != nil {
+			wwlog.Warn("failed to apply xattrs from file %s", path)
+		}
+		os.Remove(path)
+		return err
+	})
+	if err != nil {
+		wwlog.Warn("errors encountered while appying xattr files: %w", err)
+	}
+	wwlog.Debug("xattrs applied to overlay")
 
 	// Atomically move files from temp directory to current working directory
 	err = atomicApplyOverlay(tempDir, target)
@@ -527,6 +547,11 @@ func copyFile(src, dst string, srcInfo os.FileInfo) error {
 	err = os.Chtimes(dst, time.Time{}, srcInfo.ModTime())
 	if err != nil {
 		wwlog.Warn("failed to update timestamps for file %s: %s", dst, err)
+	}
+	wwlog.Debug("Copying xattrs for file: %s", dst)
+	err = util.CopyXattrs(src, dst)
+	if err != nil {
+		wwlog.Warn("failed to copy xattrs for file %s: %s", dst, err)
 	}
 
 	return nil
--- a/internal/app/wwclient/root.go
+++ b/internal/app/wwclient/root.go
@@ -27,6 +27,7 @@ import (
 	warewulfconf "github.com/warewulf/warewulf/internal/pkg/config"
 	"github.com/warewulf/warewulf/internal/pkg/pidfile"
 	"github.com/warewulf/warewulf/internal/pkg/version"
+	"github.com/warewulf/warewulf/internal/pkg/util"
 	"github.com/warewulf/warewulf/internal/pkg/wwlog"
 	"github.com/warewulf/warewulf/internal/pkg/wwurl"
 )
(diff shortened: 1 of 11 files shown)


internal/app/wwclient/root.go
```diff
@@ -321,7 +321,19 @@ func updateSystem(target string, ipaddr string, port int, wwid string, tag strin
 		wwlog.Error("failed running cpio: %s", err)
 		return
 	}
-
+	// apply saved xattrs
+	_, err = os.Stat(filepath.Join(tempDir, "xattrs"))
+	if err != nil {
+		wwlog.Debug("no xattrs file in overlay")
```
anderbubble: wwclient doesn't apply overlays individually, so this could only see one xattrs per node. So you'd need one global xattrs file: they couldn't be mixed. (I might not be expressing this well.)

internal/pkg/util/util.go
```diff
@@ -517,6 +517,12 @@ func BuildFsImage(
 	}
 	wwlog.Debug("Created image directory for %s: %s", name, imagePath)
 
+	err = CreateXattrDump(rootfsPath, filepath.Join(rootfsPath, "xattrs"))
```
anderbubble: Maybe we should store this in /warewulf/xattrs? Or /etc/warewulf/xattrs? (That's where the image_exit.sh file is, for example.)

internal/pkg/util/util.go
```diff
@@ -622,3 +628,34 @@ func EqualYaml(a interface{}, b interface{}) (bool, error) {
...
+	defer outfile.Close()
+	wwlog.Debug("Created empty xattr file for %s at %s", rootfsPath, filePath)
+
+	mask := ".*"
+	if strings.Contains(rootfsPath, "overlay") {
+		// in an overlay, only capture capability and system xattrs
+		mask = "security.capability|system.*"
+	}
+
+	getfattr := exec.Command("getfattr", "-R", "-d", "-m", mask, "-e", "hex", "-P", "-h", ".")
```
anderbubble: I know there's shelling-out in wwctl already; but any chance we could do this with something golang native?

dracut/modules.d/90wwinit/load-wwinit.sh
```diff
@@ -20,6 +20,9 @@ get_stage() {
         | gzip -d \
         | cpio -ium --directory="${NEWROOT}"
     ) || die "Unable to load stage: ${stage}"
+    pushd "${NEWROOT}" 2> /dev/null
+    setfattr -h --restore=xattrs 2> /dev/null
+    popd 2> /dev/null
```
anderbubble: Dracut should call overlays/wwinit/rootfs/warewulf/init.d/80-xattrs, so I don't think this needs to be duplicated here.

anderbubble (changes requested):

dracut/modules.d/90wwinit/load-wwinit.sh
snehring: It's needed (as written currently here) to make sure that the image, system, and runtime overlay xattrs are applied when they're fetched, so they get applied before the next overlay comes in and blows it away. As is the stuff in 80-xattrs will only apply to the runtime overlay. I need to better understand the boot and overlay application process in general.

internal/app/wwclient/root.go
snehring: This is just applying the runtime overlay right? Isn't that just a monolithic single overlay once composed and sent to the client?

internal/pkg/util/util.go
snehring: This is actually probably the solution right here. Make a directory /warewulf/xattrs/ with a unique file created for each overlay, then have 80-xattrs process all files in that directory.

internal/pkg/util/util.go
snehring: I'll look around and see what I can find.

snehring: I am pretty sure this can be accomplished with native golang, there's an xattr pkg at least, it'll just take me a bit to implement.

snehring:
The remaining test failures seem to have been due to xattr files being added. In one case this was because ```/warewulf/xattr``` tree was being created into an image that was expected to be empty, so now if a write for the xattr file fails (due to being empty or other reasons) we delete the xattr file (if it exists), the xattr directory, and the warewulf directory in root (which will only succeed if it too is empty). The api test failures that remained after that were due to the size of the image being increased over the stored value due to (presumably) the addition of the xattrs file and associated structure. 

I'm not sure if the images created during the api tests perist anywhere and can be inspected. I am assuming that the size difference is due to the added files/dirs but I've not checked. 

although as I type that I see those tests failing on  github differently than locally. So I guess I'll reverse that change.

internal/app/wwclient/root.go
mslacken: @snehring you are right, `wwclient` just manages the runtime overlay. Still I think it would be better to have `_, err = os.Stat(filepath.Join(tempDir, "xattrs.d","runtime"))` so store the xattrs separately for the overlays, so that in [overlays/wwinit/rootfs/warewulf/init.d/80-xattrs](https://github.com/warewulf/warewulf/pull/2043/files#diff-aac772994e5d842b5e5500cdc044aad03093a4626a1e719eba60570e989c4490) we might also have xattrs in the images.
@anderbubble can you please add this to agenda for the next meeting.

snehring: That is more or less the current state of things. Each "image" has an xattrs file created under ```/warewulf/xattrs``` of the form ```xattr-<sha256 of image name>```.  which is then applied by the init script. The runtime overlay is then handled  by wwclient.

snehring:
rebased to current master and squashed

I'm hoping to find some time soon to add a couple tests in, mostly just tyring to figure out the best way to do it.

snehring:
Gonna change this from draft to see what the tests show.

PR #2044, opened 2025-10-17
WIP: selinux management in wwctl configure

### LINKED ISSUE #1997: Setup tftpboot directory
### Summary

Setup tftpboot in `/var/lib/warewulf/tftpboot` where `/var/lib` is the state directory `%{_sharedstatedir}` in warewulf.spec.  Based on discussion at TSC meeting.

### Rationale

Previous to EL10 dhcp-server and tftp-server have been used.  As of EL10 dhcp-server is no longer in EL10 due to EOL and we plan on using dnsmasq as default.  Since tftp-server is not installed `/var/lib/tftboot` is no longer being created along with the associated selinux attributes.

### Description

As a part of `wwctl configure tftp`: 
* create the `/var/lib/warewulf/tftboot` directory
* set the appropriate permissions (755/644)
* warn (or ignore) if umask prevents this - or similar, see #1991
* set appropriate selinux attributes, something like
```bash
semanage fcontext -a -t public_content_t "/var/lib/tftpboot(/.*)?"
restorecon -R -v /var/lib/tftpboot
```

Make the default tftp directory `/var/lib/tftpboot`

* Where `/var/lib` is the appropriate macro for the shared state directory.
* Directory location should be setup based on what standards locations are available.  `/tftpboot` might be an option.

### Additional information

Additionally issue #1991 might be related and addressed (explicitly setting permissions at the same time)

### General information

anderbubble:

`wwctl configure tftp` can configure the SELinux context of the TFTP directory.

- Fixes #1997

## Reviewer  checklist

### DIFF
--- a/internal/pkg/configure/tftp.go
+++ b/internal/pkg/configure/tftp.go
@@ -3,7 +3,9 @@ package configure
 import (
 	"os"
 	"path"
+	"strings"
 
+	"github.com/opencontainers/selinux/go-selinux"
 	warewulfconf "github.com/warewulf/warewulf/internal/pkg/config"
 	"github.com/warewulf/warewulf/internal/pkg/util"
 	"github.com/warewulf/warewulf/internal/pkg/warewulfd"
@@ -43,6 +45,22 @@ func TFTP() (err error) {
 			}
 		}
 	}
+
+	// Set SELinux context if configured
+	if selinux.GetEnabled() && getSelinuxContext(controller.TFTP.Selinux) != "" {
+		currentLabel, err := selinux.FileLabel(tftpdir)
+		if err != nil {
+			wwlog.Warn("Failed to get current SELinux context for %s: %s", tftpdir, err)
+		} else if currentLabel != controller.TFTP.Selinux {
+			wwlog.Info("Setting SELinux context for %s to %s", tftpdir, controller.TFTP.Selinux)
+			if err := selinux.Chcon(tftpdir, controller.TFTP.Selinux, true); err != nil {
+				wwlog.Warn("Failed to set SELinux context on %s: %s", tftpdir, err)
+			} else {
+				wwlog.Info("To make the SELinux policy permanent, run: semanage fcontext -a '%s' '%s(/.*)?'", controller.TFTP.Selinux, tftpdir)
+			}
+		}
+	}
+
 	if !controller.TFTP.Enabled() {
 		wwlog.Warn("Warewulf does not auto start TFTP services due to disable by warewulf.conf")
 		return nil
@@ -55,4 +73,12 @@ func TFTP() (err error) {
 	}
 
 	return nil
+}
+
+func getSelinuxContext(value string) string {
+	if strings.ToLower(value) == "default" {
+		return "system_u:object_r:public_content_t:s0"
+	} else {
+		return value
+	}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## v4.6.5, unreleased
 
+### Added
+
+- `wwctl configure tftp` can configure the SELinux context of the TFTP directory. #1997
+
 ### Changed
 
 - Renamed debian.interfaces overlay to ifupdown
 - Change the DHCP server package used on openeuler 24.03 to dnsmasq
+- Renamed debian.interfaces overlay to ifupdown.
+- Use opencontainers/selinux to manage SELinux in wwclient.
 
 ### Fixed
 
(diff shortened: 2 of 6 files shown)


anderbubble:
@middelkoopt @mslacken @snehring this is where the PR stands right now. If you set `selinux: true` (or a few other reasonable sentinel values) then it sets the context to `system_u:object_r:public_content_t:s0`. Otherwise you can provide the explicit context.

But then I started down the path of seeing if we could just build the selinux management into the package, as suggested in the TSC meeting. I have a WIP spec that configures OpenEuler and EL10 to use /var/lib/warewulf/tftpboot, and I started configuring it to manage SELinux contexts there; but when I started researching how to do that I realized that this would make the Warewulf package require selinux, which is probably not desired. And if we just had it call the selinux commands during %post and %postun but ignore errors, then it wouldn't do the right thing if we installed selinux later.

Apparently the recommendation, then, is to have a warewulf-selinux package that does the selinux bits; but that seems over-complicated for what we're actually trying to accomplish here.

So I'm thinking we should just stick with the imperative solution presented here; but what do y'all think?

mslacken:
> @middelkoopt @mslacken @snehring this is where the PR stands right now. If you set `selinux: true` (or a few other reasonable sentinel values) then it sets the context to `system_u:object_r:public_content_t:s0`. Otherwise you can provide the explicit context.
> 
> But then I started down the path of seeing if we could just build the selinux management into the package, as suggested in the TSC meeting. I have a WIP spec that configures OpenEuler and EL10 to use /var/lib/warewulf/tftpboot, and I started configuring it to manage SELinux contexts there; but when I started researching how to do that I realized that this would make the Warewulf package require selinux, which is probably not desired. And if we just had it call the selinux commands during %post and %postun but ignore errors, then it wouldn't do the right thing if we installed selinux later.
> 
> Apparently the recommendation, then, is to have a warewulf-selinux package that does the selinux bits; but that seems over-complicated for what we're actually trying to accomplish here.
> 

As white-haired old linux wicher I would recommend following
* do the selinux rectifying in a script that we ship somewhere under /var/lib/warewulf/scripts/rectifyselinux
* run the script in %post (not sure if this is really necessary)
* the script doesn't fail if selinux isn't present, but just print that selinux isn't present, but that the script can be rerun
* recommend selinux in the warewulf package
* add to the spec file following
```
# rerun selinux setup when it's installed
%triggerin -- selinux
 /var/lib/warewulf/scripts/rectifyselinux
``` 
*  send your coin to the witcher

anderbubble:
> do the selinux rectifying in a script that we ship somewhere under /var/lib/warewulf/scripts/rectifyselinux

Imo, if we do this, we might as well drop `wwctl configure` entirely.

`wwctl configure` _is_ the "script" that we ship somewhere.

(I am not advocating for dropping `wwctl configure`; I just don't want to split related behavior across multiple places.)

internal/pkg/configure/tftp.go
```diff
@@ -43,6 +45,22 @@ func TFTP() (err error) {
 			}
 		}
 	}
+
+	// Set SELinux context if configured
+	if selinux.GetEnabled() && getSelinuxContext(controller.TFTP.Selinux) != "" {
+		currentLabel, err := selinux.FileLabel(tftpdir)
+		if err != nil {
+			wwlog.Warn("Failed to get current SELinux context for %s: %s", tftpdir, err)
+		} else if currentLabel != controller.TFTP.Selinux {
```
snehring, line 54: It might make sense to try to apply the label even if getting the current context failed for whatever reason

snehring:
If we intend to keep the tftpboot directory customizable, then I think this PR is a reasonable fix. Very few other services automagically set selinux labels for you if you opt to change defaults, you're mostly just supposed to understand and handle the problems that can introduce with selinux yourself. Doing it on behalf of the user is very kind!

If we were to elect to make the tftpboot directory fixed, something like /var/lib/warewulf/tftpboot and direct whichever daemon at it, then I think applying proper permanent selinux labels would be appropriate. I don't know how fancy we can get (I'm not an rpm expert) if we can have a section in %post that triggers if selinux is installed that sets the labels or if we can have a warewulf-selinux rpm that is pulled in as a dep when selinux is installed that handles it. That feels most 'proper' to me.

mslacken:
> > do the selinux rectifying in a script that we ship somewhere under /var/lib/warewulf/scripts/rectifyselinux
> 
> Imo, if we do this, we might as well drop `wwctl configure` entirely.
> 
> `wwctl configure` _is_ the "script" that we ship somewhere.
> 
> (I am not advocating for dropping `wwctl configure`; I just don't want to split related behavior across multiple places.)

I just wanted to drop in that instead of  `%post`  we use `%triggerin -- selinux` which will be called when `selinux` is already installed  on the system and if it's not installed on the system, also called at the selinux installation time.

anderbubble:
> %triggerin -- selinux

I will definitely look into this. Thanks!

PR #2072, opened 2025-12-03
Adding TLS

mslacken:
@anderbubble I **strongly** recommend that you add a commit which enables the API only over https

# Added
In `warewulf.conf ` there is now following options
```
warewulf:
  enable https: true
  secure port: 9874
```
which are self explantory

A key can be either imported with `wwctl configure keys --import`
or created with `wwctl configure keys --create`

If https is enabled the runtime overlays can only be delivered via https

# Testing

After enabling you can test this with
```
 curl --cacert /warewulf/keys/warewulf.crt https://warewulf:9874/overlay-runtime/52:54:00:2e:42:48 -- | cpio -t
```

### DIFF
--- a/internal/app/wwclient/root.go
+++ b/internal/app/wwclient/root.go
@@ -117,7 +117,7 @@ func CobraRunE(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	tlsConfig := &tls.Config{}
-	if conf.Warewulf.EnableHttps() {
+	if conf.Warewulf.EnableTLS() {
 		caCert, err := os.ReadFile("/warewulf/keys/warewulf.crt")
 		if err != nil {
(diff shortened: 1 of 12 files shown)


mslacken:
Changed names now to TLS instead of Keys.
Still its "secure port" in the yaml config, but this is consistent with "warewulf port", so I choose consistence naming for this variable

anderbubble:
@mslacken I'm finally working on this now. I'll let you know if I have any questions.

anderbubble:
Rebased to origin/main.

anderbubble:
@mslacken I've finished reviewing this and I think it's ready at https://github.com/warewulf/warewulf/pull/2125; but now it needs your review / approval before merge.

Let me know what you think, or if you have any complains, concerns, or questions.

PR #2086, opened 2026-01-14
Added --partwipe flag for set

mslacken:
The wipe_partition_entry for partitions in node or profiles could only be set via `wwctl node/profile edit`.
Added a commandline option for this.


anderbubble (approved):
Looks good! Just needs a rebase.

PR #2087, opened 2026-01-14
added chrony template

mslacken:
Addec a template with internal doc for chrony


overlays/chrony/rootfs/etc/chrony.conf.ww
```diff
@@ -0,0 +1,31 @@
+{{/* wwdoc: Configures chrony NTP client/server */ -}}
+{{/* chrony_servers: Comma-separated list of NTP servers. Defaults to Warewulf controller IP. */ -}}
+# This file is autogenerated by warewulf
+
+{{- if .Tags.chrony_servers }}
```
anderbubble: Can you make it work for either chrony_servers or ntp_servers?

anderbubble (changes requested):
I absolutely support this; but it raises a question for me, as CIQ currently distributes a chrony overlay outside of Warewulf. How should be prevent such colissions in the future? Perhaps with a `ciq.` prefix to overlay names?

mslacken:
> I absolutely support this; but it raises a question for me, as CIQ currently distributes a chrony overlay outside of Warewulf. How should be prevent such colissions in the future? Perhaps with a `ciq.` prefix to overlay names?

The SUSE solution for this is upstream first.

middelkoopt:
I wonder if Warewulf should have a "vendor" place for overlays.  `/usr/src/warewulf/overlays` for the distribution, `/var/lib/warewulf/vendor/overlays` for other packages overlays (I did not spend much time thinking about the name/path).  Not quite sure how this would work but would put it out there.  Or perhaps a convention that all CIQ/vendor overlays start with a prefix/ciq- to remove the possibility of collision (I'm currently favoring this one).

mslacken:
@anderbubble can we merge that one

anderbubble:
@mslacken sorry for the delay on this one. Should be simple. I'll look it over and get it merged (after potentially a small suggestion).

anderbubble (approved):

anderbubble:
@mslacken I've approved for merge, but I made a change so leaving it for you to actually merge at your discretion.

PR #2028, opened 2025-09-28
Fix mkfs overlay: handle false-positive filesystems and add safe overwrite

kosmolito:
- Updated `already_formatted()` to use `blkid` plus safe read-only mount for accurate filesystem detection.
- Added conditional `-f` flag to `mkfs` only when `overwrite: true` is set.
- Removed `continue` statements from the shell script to prevent runtime errors.
- Updated Go overlay snapshot tests to reflect new template rendering.
- Ensures disks with leftover signatures but no valid filesystem are properly reformatted during node provisioning.

## Description

This PR fixes two critical issues in the `20-mkfs.sh.ww` overlay related to filesystem detection and formatting logic.

### Background

Previously, the script relied on `wipefs -n` to detect existing filesystems. This caused **false positives** when a disk had leftover filesystem signatures but no valid filesystem. As a result, `mkfs` would fail.

Additionally, the script contained `continue` statements outside of loops, which caused runtime errors:

```bash
20-mkfs.sh: line 45: continue: only meaningful in a for, while, or until loop
```

### Changes in This PR

1. **Improved filesystem detection**:
   - `already_formatted` now uses `blkid` to detect existing filesystem types.
   - It attempts a **safe read-only mount** to confirm the filesystem is valid and usable.
   - Treats disks with stale signatures that cannot be mounted as unformatted.

2. **Safe overwrite**:
   - Adds `-f` to the `mkfs` command when the disk is flagged for overwrite, ensuring formatting succeeds even if old signatures exist.

3. **Fix invalid `continue` usage**:
   - Removed `continue` statements outside loops to prevent runtime errors.

### Benefits

- Prevents formatting failures on test and real nodes due to false positive filesystem detection.
- Makes the overlay script compatible with existing Warewulf workflows.
- Eliminates shell errors caused by invalid `continue` statements.

### Testing

- Updated unit tests (`mkfs_test.go`) to reflect new `already_formatted` logic.
- Manual testing on partitions with stale signatures confirms that:
  - Disks are properly formatted when `overwrite: true`.
  - Non-overwritten disks with valid filesystems are skipped safely.

### Conclusion

This PR ensures the mkfs overlay script is robust against false-positive filesystem detection, applies safe forced formatting when required, and removes shell errors caused by invalid continue statements.

## This fixes or addresses the following GitHub issues:

- Fixes #2027

## Reviewer  checklist

The reviewer checks the following items before merging the PR.

[...]

Issue #2027, opened 2025-09-28
mkfs fails on disks with stale filesystem signatures and uses invalid 'continue' statements


kosmolito:
## Description

When running `20-mkfs.sh` on nodes, I encountered two issues:  

### 1. False positives for already formatted disks
- `wipefs` detects an old XFS signature even when the filesystem is gone.  
- `mkfs` fails

### 2. Invalid continue usage
- The script contains continue statements outside loops, results in errors.

## Expected Behavior
- The mkfs command should safely include -f when overwriting is required.
- Script should not use continue outside loops.

## Steps to Reproduce
- Prepare a test node with partitions and filesystems as defined below.
- Add sfdisk and mkfs as system overlays to the node/profile

Run:
```bash
bash 20-mkfs.sh
```

Observe the errors described.

## Proposed Fix
- Add -f/-F to mkfs when overwrite is true.
- Remove invalid continue statements outside loops.

### Environment

```yaml
# nodes.conf
```yaml
nodeprofiles:
  default:
    comment: Base profile automatically included for all nodes
    ipxe template: default
    runtime overlay:
      - hosts
      - ssh.authorized_keys
      - syncuser
    system overlay:
      - wwinit
      - wwclient
      - fstab
      - hostname
      - ssh.host_keys
      - issue
      - resolv
      - udev.netname
      - systemd.netname
      - NetworkManager
      - mkswap
      - systemd.mount
      - sfdisk
      - mkfs
      - systemd.swap
    kernel:
      args:
        - crashkernel=no
        - rd.shell
    ipmi:
      template: ipmitool.tmpl
    init: /sbin/init
    root: initramfs
    network devices:
      default:
        device: eno1
        netmask: 255.255.255.0
        gateway: 10.0.0.254
    resources:
      fstab:
        - file: /home
          mntops: defaults,nofail
          spec: warewulf:/home
          vfstype: nfs
  slurm-common:
    profiles:
      - default
    system overlay:
      - munge
      - slurm.common
    network devices:
      default:
        tags:
          DNS1: 10.0.0.254
    tags:
      IPXEMenuEntry: dracut
  slurm-head:
    profiles:
      - slurm-common
    image name: head
    system overlay:
      - slurm.head
nodes:
  head:
    profiles:
      - slurm-head
    network devices:
      default:
        hwaddr: BC:12:12:12:12:12
        ipaddr: 10.0.0.10
    tags:
      head_disk_mounting_path: /mnt/slurm
[...]

PR #2095, opened 2026-01-25
Fix for CWE-23 in internal/pkg/clean/clean.go

jsco2t:

Change resolves a CWE-23 issue (path traversal) in internal/pkg/clean/clean.go

## Reviewer  checklist

### DIFF
--- a/internal/pkg/clean/clean.go
+++ b/internal/pkg/clean/clean.go
@@ -1,8 +1,10 @@
 package clean
 
 import (
+	"fmt"
 	"os"
-	"path"
+	"path/filepath"
+	"strings"
 
 	warewulfconf "github.com/warewulf/warewulf/internal/pkg/config"
 	"github.com/warewulf/warewulf/internal/pkg/node"
@@ -26,7 +28,11 @@ func CleanOverlays() error {
 		return err
 	}
 	nodes := nodeDB.ListAllNodes()
-	dirList, err := os.ReadDir(warewulfconf.Paths.OverlayProvisiondir())
+
+	// Clean the base directory path FIRST
+	baseDir := filepath.Clean(warewulfconf.Paths.OverlayProvisiondir())
+
+	dirList, err := os.ReadDir(baseDir)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/clean/clean.go
```diff
@@ -26,17 +27,38 @@ func CleanOverlays() error {
...
+
+			// Verify the path is within baseDir
+			rel, err := filepath.Rel(baseDir, cleanTarget)
+			if err != nil {
+				wwlog.Warn("failed to compute relative path for %s: %v", item.Name(), err)
+				continue
+			}
+
+			if strings.HasPrefix(rel, "..") {
+				wwlog.Warn("skipping directory with path traversal attempt: %s", item.Name())
```
anderbubble: @jsco2t I think this message needs to be a bit clearer. (Maybe something like "{item.Name()} is not inside overlay directory {baseDir}".)

I also think that state is probably worth returning an error rather than warning and continuing, since it shouldn't ever happen. But I could be misunderstanding that.

mslacken:
@jsco2t or @anderbubble can you please share what CWE-23 exactly means?

anderbubble:
> @jsco2t or @anderbubble can you please share what CWE-23 exactly means?

https://cwe.mitre.org/data/definitions/23.html

mslacken:
So this code protects us if the user, who is `root`  creates the hilarious node named `node01/../../awsesome_path` and then removes that node?
If yes, why don't we just forbid `..` and `/` when creating a node as if you have node name like this, a lot of other things will also break?

internal/pkg/clean/clean.go
jsco2t: I wasn't clear if it should be an error state or just a warning. I'm fine failing hard if it happens. Will fix.

anderbubble:
> So this code protects us if the user, who is `root` creates the hilarious node named `node01/../../awsesome_path` and then removes that node? If yes, why don't we just forbid `..` and `/` when creating a node as if you have node name like this, a lot of other things will also break?

Both might be the right thing: sanitizing the path _and_ having stricter rules for node names. In this case, though, the goal is to address a security alert generated in an auditing tool that CIQ ran.

mslacken:
> > So this code protects us if the user, who is `root` creates the hilarious node named `node01/../../awsesome_path` and then removes that node? If yes, why don't we just forbid `..` and `/` when creating a node as if you have node name like this, a lot of other things will also break?
> 
> Both might be the right thing: sanitizing the path _and_ having stricter rules for node names. In this case, though, the goal is to address a security alert generated in an auditing tool that CIQ ran.

No. Sanitizing the node names solves the core issue and as we use something like `path.Join(someDir,node.Id())` quite often in the code, so this PR is just some cosmetic change.
Fixing this fringe case with a lot of code for some auditing tools is not sensible as long as e.g. #2027 isn't addressed, as this solves the **real** issue that this protects the asset keys in the network.

anderbubble:
> as long as e.g. #2027 isn't addressed

I suspect you must be talking about something else, since #2027 was fixed last year by #2028.

mslacken:
> > as long as e.g. #2027 isn't addressed
> 
> I suspect you must be talking about something else, since #2027 was fixed last year by #2028.

Right, meant #2072

internal/pkg/clean/clean.go
jsco2t: Fixed

anderbubble:
Rebased to main and squashed. (Also fixed up the changelog.)

PR #2113, opened 2026-02-18
added formated output for node list

mslacken:
Allow the the possiblity for freely formated output via
```
wwctl node list --format "{{ .Id }} profiles: {{range .Profiles}}{{ . }} ips: {{end}}{{range .NetDevs}}{{ .Ipaddr}}{{end}}"
```
Could also be adapated for profiles.
Needed also some modifications for methos as `func(n *Node) foo()` string won't work in templates, but `func(n Node) foo()` works like a charm.


mslacken:
@anderbubble expanding this to `wwctl profile list` and use template formatting instead `--net`, `--impi` so that these options just set the format string might be useful for someone to get used to the warerwulf code.

anderbubble:
I like the idea, for sure! I'll try to look at it after the TLS work.

PR #2125, opened 2026-03-10
TLS for warewulfd and REST API, and warewulfd refactor

anderbubble:

Supplement to #2072 which adds TLS support to the REST API, fixes bugs, and refactors warewulfd overall.

- Closes #2072

## Reviewer  checklist

### DIFF
--- a/internal/pkg/warewulfd/parser.go
+++ b/internal/pkg/warewulfd/parser.go
@@ -14,10 +14,9 @@ import (
 
 // requestContext holds the validated results of the parsed request
 type requestContext struct {
-	conf        *warewulfconf.WarewulfYaml
-	rinfo       parsedRequest
-	remoteNode  node.Node
-	statusStage string
+	conf       *warewulfconf.WarewulfYaml
+	rinfo      parsedRequest
+	remoteNode node.Node
 }
 
 // initHandleRequest performs common initial request parsing, security checks,
@@ -45,17 +44,6 @@ func initHandleRequest(w http.ResponseWriter, req *http.Request) (*requestContex
 		}
 	}
 
-	status_stages := map[string]string{
-		"efiboot":   "EFI",
-		"grub":      "GRUB",
-		"ipxe":      "IPXE",
-		"kernel":    "KERNEL",
-		"system":    "SYSTEM_OVERLAY",
-		"runtime":   "RUNTIME_OVERLAY",
-		"initramfs": "INITRAMFS"}
-
-	statusStage := status_stages[rinfo.stage]
-
 	remoteNode, err := GetNodeOrSetDiscoverable(rinfo.hwaddr, conf.Warewulf.AutobuildOverlays())
 	if err != nil && err != node.ErrNoUnconfigured {
 		wwlog.ErrorExc(err, "")
@@ -66,15 +54,14 @@ func initHandleRequest(w http.ResponseWriter, req *http.Request) (*requestContex
 	if remoteNode.AssetKey != "" && remoteNode.AssetKey != rinfo.assetkey {
 		w.WriteHeader(http.StatusUnauthorized)
 		wwlog.Denied("incorrect asset key: node %s: %s", remoteNode.Id(), rinfo.assetkey)
-		updateStatus(remoteNode.Id(), statusStage, "BAD_ASSET", rinfo.ipaddr)
+		updateStatus(remoteNode.Id(), rinfo.stage, "BAD_ASSET", rinfo.ipaddr)
 		return nil, fmt.Errorf("incorrect asset key")
 	}
 
 	return &requestContext{
-		conf:        conf,
-		rinfo:       rinfo,
-		remoteNode:  remoteNode,
-		statusStage: statusStage,
+		conf:       conf,
+		rinfo:      rinfo,
+		remoteNode: remoteNode,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


internal/pkg/warewulfd/parser.go
```diff
@@ -34,14 +103,23 @@ func parseReq(req *http.Request) (parserInfo, error) {
...
-	if stage != "efiboot" {
+	switch stage {
+	case "efiboot":
+		// /efiboot/{file}: no hwaddr in path; identified via ARP
+		if len(path_parts) > 3 {
+			ret.efifile = strings.Join(path_parts[2:], "/")
+		} else {
+			ret.efifile = path_parts[2]
+		}
+	case "grub":
```
mslacken: Same as default, so it could go away

internal/pkg/warewulfd/parser.go
```diff
@@ -62,19 +140,22 @@ func parseReq(req *http.Request) (parserInfo, error) {
 		ret.stage = req.URL.Query()["stage"][0]
 	} else {
 
-		if stage == "ipxe" || stage == "provision" {
+		switch stage {
+		case "ipxe", "provision":
```
mslacken: second time we do switch on stage, could there just be one?

internal/pkg/warewulfd/parser.go
```diff
@@ -85,14 +166,23 @@ func parseReq(req *http.Request) (parserInfo, error) {
 	if len(req.URL.Query()["compress"]) > 0 {
 		ret.compress = req.URL.Query()["compress"][0]
 	}
+	if ret.efifile == "" && len(req.URL.Query()["file"]) > 0 {
+		ret.efifile = req.URL.Query()["file"][0]
+	}
 	if ret.stage == "" {
 		return ret, errors.New("no stage encoded in GET")
 	}
 	if ret.hwaddr == "" {
```
mslacken: Doesn't the first switch stage a hwaddr?

internal/pkg/warewulfd/parser.go
```diff
@@ -1,15 +1,84 @@
...
+		"efiboot":   "EFI",
+		"grub":      "GRUB",
+		"ipxe":      "IPXE",
+		"kernel":    "KERNEL",
+		"system":    "SYSTEM_OVERLAY",
+		"runtime":   "RUNTIME_OVERLAY",
+		"initramfs": "INITRAMFS"}
+
+	statusStage := status_stages[rinfo.stage]
+
```
mslacken: Can we just use rinfo.stage here do this string conversation in `wwctl node status` where this stages are used?
Also a `switch rinfo` would be a better solution here, as e.g. `baar :=map["foo"]` would fail

userdocs/server/routes.rst
```diff
@@ -0,0 +1,325 @@
...
+
+.. note::
+
+   The port ``warewulfd`` listens on is configured with ``warewulf:port`` in
+   ``warewulf.conf`` (default: ``9873``). When TLS is enabled, a second listener
+   is started on the port configured with ``warewulf:tls port`` (default: ``9874``).
+
+URL Patterns
+============
+
```
mslacken, line 23: Could we get this pattern also in `parser.go` or have in the source code a reference to this doc?

userdocs/server/routes.rst
```diff
@@ -0,0 +1,325 @@
...
+of colons and are normalized automatically.
+
+.. note::
+
+   The port ``warewulfd`` listens on is configured with ``warewulf:port`` in
+   ``warewulf.conf`` (default: ``9873``). When TLS is enabled, a second listener
+   is started on the port configured with ``warewulf:tls port`` (default: ``9874``).
+
+URL Patterns
+============
```
mslacken, line 22: Thanks for documenting this!

internal/pkg/warewulfd/parser.go
anderbubble: Capturing additional context from a Slack conversation:

```
The hwaddr is set in the first switch statement of the function, as there is default whic looks like the following

	default:
		hwaddr = path_parts[2]
		hwaddr = strings.ReplaceAll(hwaddr, "-", ":")
		hwaddr = strings.ToLower(hwaddr)

so hwaddr is set, what means that  ret.hwaddr == "" could never happen or am I wrong?
```

internal/pkg/warewulfd/parser.go
anderbubble: The default case is setting `hwaddr`, but this is setting `ret.efifile`. The `default` case also would not be triggered for case `efiboot`, so this wouldn't get set when not `len(path_parts) > 3`. So it has to be specified explicitly here.

internal/pkg/warewulfd/parser.go
anderbubble: I did a full audit of this function and simplified it a bunch. I think it should resolve this concern.

internal/pkg/warewulfd/parser.go
anderbubble: Check out the re-work of parseRequest; I think it's simpler and more straightforward now.

userdocs/server/routes.rst
anderbubble: Done.

internal/pkg/warewulfd/parser.go
anderbubble: Done.

anderbubble:
@mslacken I've addressed all the comments. Thanks for the review! Great feedback.

mslacken (approved):

PR #2136, opened 2026-03-18
TPM based security

mslacken:
# Node based specific TPM based security

If for node the `tpm` parameter is `true` following things are done:
* the checksums of the files sent out from warewulf to the nodes are recorded
* during the startup of `wwclient` a quote is generated on the node and uploaded on the server
* the integrity of the quote is checked
* the node downloads a challenge, but only if the server can reply PCR9 correctly
* the challenge which can be only be decrypted in the TPM of the node, contains a secret which is used to protect the runtime overlay of the node

TDLR:
The node can only download the runtime image if its has booted correclty


mslacken:
@anderbubble The tests are failing as there is this dreaded licenses problem which I don't know hot to fix.

Although this adds a lot complex stuff, it doesn't interfere with the rest of warewulf. This is definitely a tech preview but makes warewulf to an unique secure distribution mechanism.

PR #2142, opened 2026-03-22
Grpc removal

### LINKED ISSUE #2141: Proto file regneration
For the old v1 API we *used* protogen with the proto file `./internal/pkg/api/routes/v1/routes.proto` to create the resulting go files with the `make proto` target.
This target was removed, so it's hard to recreate that files right now.
So we have two options right now
* recreate the proto target
or
* get rid of the old API calls as these are just confusing and not useful right now

anderbubble:

Removes all dependencies on the previous gRPC system and generated protobuf files.

Also fixes bugs identified by copilot:

- Fix `wwctl image build --all` to build all images
- Fix `wwctl node set --all` to set values on all nodes

- Fixes #2141

## Reviewer  checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Refactored `wwctl <node|profile> set` to use cobra `cmd.Flags().Changed()` to
   apply only explicitly-set fields.
+- Deprecated `wwctl node set --force` (never had any effect; will be removed in
+  a future release)
 - Runtime overlay download failure during dracut/wwinit boot is now non-fatal;
   the node continues to boot and `wwclient` retries the download at runtime.
 - `hosts` overlay added to the default system overlay list
--- a/internal/app/wwctl/node/set/root.go
+++ b/internal/app/wwctl/node/set/root.go
@@ -40,6 +40,8 @@ func GetCommand() *cobra.Command {
 	baseCmd.PersistentFlags().BoolVarP(&vars.setNodeAll, "all", "a", false, "Set all nodes")
 	baseCmd.PersistentFlags().BoolVarP(&vars.setYes, "yes", "y", false, "Set 'yes' to all questions asked")
 	baseCmd.PersistentFlags().BoolVarP(&vars.setForce, "force", "f", false, "Force configuration (even on error)")
+	_ = baseCmd.PersistentFlags().MarkDeprecated("force", "this flag never had any effect and will be removed in a future release")
+	_ = baseCmd.PersistentFlags().MarkHidden("force")
 	// register the command line completions
 	if err := baseCmd.RegisterFlagCompletionFunc("image", completions.Images); err != nil {
 		panic(err)


anderbubble:
@mslacken I think this should resolve your concern from #2141, but let me know what you think.

LICENSE_DEPENDENCIES.md
```diff
@@ -383,12 +383,6 @@ The dependencies and their licenses are as follows:
 
 **License URL:** <https://github.com/gorilla/mux/blob/v1.8.1/LICENSE>
```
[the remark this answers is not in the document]
anderbubble: grpc-gateway is pulled in as a transitive dependency, but it's never actually used by wwctl, so it doesn't actually affect our licensing.

internal/pkg/image/delete.go
```diff
@@ -0,0 +1,78 @@
...
+func Delete(name string) error {
+	// validate image names
+	if !ValidSource(name) {
+		return fmt.Errorf("image name is not valid source: %s", name)
+	}
+
+	nodeDB, err := node.New()
+	if err != nil {
+		return fmt.Errorf("could not open nodeDB: %s", err)
+	}
```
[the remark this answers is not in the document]
anderbubble: I'll submit a separate PR fixing this throughout.

internal/app/wwctl/node/add/main.go
```diff
@@ -65,11 +67,44 @@ func CobraRunE(vars *variables) func(cmd *cobra.Command, args []string) error {
...
+		}
+		nodeArgs := hostlist.Expand(args)
+		var ipv4, ipmiaddr net.IP
+		for _, a := range nodeArgs {
+			n, err := nodeDB.AddNode(a)
+			if err != nil {
+				return fmt.Errorf("failed to add node: %w", err)
+			}
+			err = yaml.Unmarshal(buffer, n)
+			if err != nil {
```
mslacken: I know where the buffer comes from, but no, if we change the code, we can rid off the buffer.
This whole function works all lot around to call the api, in order to know what was changed or added. Without the API we simplify stuff here.

anderbubble: Thanks for catching this. I removed this pattern in several other places, but accidentally left it here.

anderbubble:
> Could we postpone this to next release. It's a bit late in the game right now and may change editing/adding behavior in subtile unwanted ways.

We could, but (1) is this not what you asked for? and (2) it might be better to make a change like this, and risk "subtle changes" in v4.7.0 rather than risk the same changes in v4.7.1.

Fwiw, this _does_ pass all the previous test cases, and there are _new_ test cases, too.

internal/app/wwctl/node/delete/main.go
```diff
@@ -2,36 +2,64 @@ package delete
...
-		var nodeList []node.Node
-		// The checks run twice in the prompt case.
-		// Avoiding putting in a blocking prompt in an API.
-		nodeList, err = apiNode.NodeDeleteParameterCheck(&ndp, false)
-		if err != nil {
-			return
+	nodeArgs := hostlist.Expand(args)
+	var nodeList []node.Node
+	for _, r := range nodeArgs {
+		var match bool
```
mslacken, line 29: Nota bene:
We could check the actual node status and avoid so to delete running nodes, but this is hard to determine....

PR #2155, opened 2026-04-09
adding dracut_static ipxe method

snehring:

There are situations where someone might want to boot dracut with static network config in place instead of relying on dhcp. I've had a custom ipxe template in play for a while to work around an issue where the ipxe client on infiniband cards uses the 48bit form of the guid for the mac address, whereas networkmanager uses the full 64 bits. Figure it might be useful to upstream it.

## Reviewer  checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Document configuring the arp cache for large clusters
 - Expand troubleshooting documentation for container runtimes
 - Ignore local coding agent files
+- Add dracut_static ipxe method
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


anderbubble:
@snehring I've taken a note to review. Thanks for contributing!

mslacken:
Is it sensible to have that functionality also in grub or doesn't that work with the mentioned info I and cards?

snehring:
Maybe. Dracut uses networkmanager to configure interfaces, which is where the problem (in my env) exists when it uses the 64bit guid as the mac.

I'm (completely without actual insight into how grub handles netboots) assuming grub doesn't reconfigure interfaces it's handed from ipxe (the ipxe on the ib card). If that's true then it should be fine, at least until the switch root and hand off to the initrd.

I've not experimented with grub boot at all in my env. I'm guessing once the node starts wwinit and then systemd the normal networkmanager overlay config files should configure the node as desired (in my case manual static addreses)

mslacken (approved):

PR #2167, opened 2026-04-30
Add CI integration test provisioning VMs with Warewulf

adrianreber:

This adds a first integration test that exercises the full Warewulf provisioning workflow in CI. The test uses Lima to spin up a Rocky Linux 10 VM on the GitHub Actions runner, builds and installs Warewulf from source inside that VM, then provisions two QEMU virtual machines via PXE boot over a virtual bridge network. The test verifies that both compute nodes boot successfully and become reachable over SSH.

The integration test performs the following steps:
- Set up a bridge network (br0) with two TAP interfaces
- Configure Warewulf with static DHCP and debug logging
- Import a Rocky Linux 10 container image
- Register two compute nodes with static IPs and MACs
- Launch two QEMU VMs that network-boot via Warewulf
- Poll SSH until both nodes are confirmed running

This is based on an idea to better and earlier test OpenHPC in the development workflow. These changes to Warewulf are a first step towards that goal, providing the infrastructure to validate the provisioning pipeline in CI before changes reach downstream OpenHPC integration testing.

- Fixes #

## Reviewer  checklist

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -2,6 +2,7 @@
 name: testsuite
 
 on:
+  workflow_dispatch:
   pull_request:
     branches:
       - main
@@ -53,3 +54,26 @@ jobs:
         run: make all
       - name: Build the Docker image
         run: docker build . --file Dockerfile
+
+  integration-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Warewulf
+        uses: actions/checkout@v4
+      # /opt and /home are bind-mounted from the host because
+      # overlayfs does not support NFS exporting.
+      - name: Start Rocky Linux 10 container
+        run: >-
+          sudo podman run
+          -w /home/warewulf
+          --privileged
+          -v /home:/home
+          -v /opt:/opt
+          -v ${{ github.workspace }}:/home/warewulf
+          -d
+          --systemd=true
+          --hostname sms
+          rockylinux/rockylinux:10-ubi-init
+          /usr/lib/systemd/systemd
+      - name: Run integration test
+        run: sudo podman exec -it -l bash -c "cd /home/warewulf; bash scripts/integration-test-el.sh"
--- a/scripts/integration-test-el.sh
+++ b/scripts/integration-test-el.sh
@@ -1,0 +1,331 @@
+#!/bin/bash
+set -ex
+
+# Check for KVM and dump CPU info
+ls -la /dev/kvm 2>/dev/null || echo "WARNING: /dev/kvm not found — KVM unavailable"
+lscpu
+
+loop_command() {
+	local retry_counter=0
+	local max_retries=5
+
+	while true; do
+		((retry_counter += 1))
+		if [ "${retry_counter}" -gt "${max_retries}" ]; then
+			exit 1
+		fi
+		# shellcheck disable=SC2068
+		$@ && break
+
+		# In case it is a network error let's wait a bit.
+		echo "Retrying attempt ${retry_counter}"
+		sleep "${retry_counter}"
+	done
+}
+
+# Clean dnf cache to avoid stale metadata in container images
+loop_command dnf clean all
+
+# Enable EPEL and CRB repositories
+loop_command dnf -y install epel-release
+loop_command dnf config-manager --set-enabled crb
+
+# Install build and VM provisioning dependencies
+loop_command dnf -y install \
+	cpio \
+	git \
+	make \
+	golang \
+	gpgme-devel \
+	python3-devel \
+	qemu-kvm \
+	ipxe-roms-qemu \
+	edk2-ovmf \
+	dnsmasq \
+	iproute \
+	iptables-nft \
+	initscripts-service \
+	yq \
+	ipxe-bootimgs-x86 \
+	ipxe-bootimgs-aarch64 \
+	tftp-server \
+	nfs-utils \
+	policycoreutils-python-utils
+
+# Build and install Warewulf
+go mod vendor
+make defaults PREFIX=/usr SYSCONFDIR=/etc
+make build
+make install
+cp -f etc/warewulf.conf-el10 /etc/warewulf/warewulf.conf
+systemctl daemon-reload
+
+# Network configuration
+export sms_ip=192.168.100.1
+export internal_netmask=255.255.255.0
+export internal_network=192.168.100.0
+export eth_provision=br0
+
+# Detect gateway and DNS from the host
+export ipv4_gateway
+ipv4_gateway=$(ip route | awk '/default/ {print $3; exit}')
+export dns_servers
+dns_servers=$(awk '/^nameserver/ {print $2; exit}' /etc/resolv.conf)
+
+# Compute node definitions
+export num_computes=2
+c_ip[0]=192.168.100.100
+c_ip[1]=192.168.100.101
+c_mac[0]=52:54:00:00:01:00
+c_mac[1]=52:54:00:00:01:01
+c_name[0]=c0
+c_name[1]=c1
+c_bmc[0]=10.16.1.1
+c_bmc[1]=10.16.1.2
+
+# Set up bridge and TAP interfaces for VMs
+ip link add br0 type bridge
+ip addr add 192.168.100.1/24 dev br0
+ip link set br0 up
+
+for ((i = 0; i < num_computes; i++)); do
+	ip tuntap add tap$i mode tap
+	ip link set tap$i up
+	ip link set tap$i master br0
+done
+
+# ---------------------------------------------------------------------------
+# Install fake ipmitool
+#
+# The integration test calls ipmitool to power-reset compute nodes.
+# We intercept "chassis power reset" and launch a QEMU VM on the
+# corresponding TAP interface instead.
+# ---------------------------------------------------------------------------
+cat >/usr/local/bin/ipmitool <<'IPMIEOF'
+#!/bin/bash
+# Fake ipmitool: intercepts "chassis power reset" to launch QEMU VMs.
+
+# BMC address -> compute index mapping
+declare -A BMC_TO_IDX=(
+	[10.16.1.1]=0
+	[10.16.1.2]=1
+)
+
+# MAC addresses per compute index
+declare -a MACS=(
+	"52:54:00:00:01:00"
+	"52:54:00:00:01:01"
+)
+
+# Compute node names (for log prefixing)
+declare -a NAMES=(
+	"c0"
+	"c1"
+)
+
+# Parse the BMC host from arguments (-H <host>)
+bmc_host=""
+args=("$@")
+for ((j = 0; j < ${#args[@]}; j++)); do
+	if [[ "${args[$j]}" == "-H" ]]; then
+		bmc_host="${args[$((j + 1))]}"
+		break
+	fi
+done
+
+# Check if this is a "chassis power reset" command
+is_power_reset=0
+for ((j = 0; j < ${#args[@]} - 2; j++)); do
+	if [[ "${args[$j]}" == "chassis" && \
+		"${args[$((j + 1))]}" == "power" && \
+		"${args[$((j + 2))]}" == "reset" ]]; then
+		is_power_reset=1
+		break
+	fi
+done
+
+if [[ "${is_power_reset}" -ne 1 ]] || [[ -z "${bmc_host}" ]]; then
+	# Not a power reset or no host specified — try real ipmitool
+	if command -v /usr/sbin/ipmitool > /dev/null 2>&1; then
+		exec /usr/sbin/ipmitool "$@"
+	fi
+	echo "fake ipmitool: ignoring: $*" >&2
+	exit 0
+fi
+
+idx="${BMC_TO_IDX[${bmc_host}]}"
+if [[ -z "${idx}" ]]; then
+	echo "fake ipmitool: unknown BMC host ${bmc_host}" >&2
+	exit 1
+fi
+
+mac="${MACS[${idx}]}"
+name="${NAMES[${idx}]}"
+tap="tap${idx}"
+
+echo "fake ipmitool: launching QEMU VM ${name} on ${tap} (mac=${mac})"
+
+# Find the iPXE ROM
+ROMFILE=""
+for candidate in \
+	/usr/share/ipxe/qemu/pxe-virtio.rom \
+	/usr/share/ipxe/virtio-net.rom \
+	/usr/share/qemu/pxe-virtio.rom; do
+	if [[ -f "${candidate}" ]]; then
+		ROMFILE="${candidate}"
+		break
+	fi
+done
+
+# Architecture-specific QEMU flags
+ARCH_FLAGS=()
+case "$(uname -m)" in
+aarch64)
+	ARCH_FLAGS=(
+		-M virt,gic-version=max
+		-drive "if=pflash,format=raw,readonly=on,file=/usr/share/AAVMF/AAVMF_CODE.fd"
+	)
+	;;
+esac
+
+NETDEV_OPTS="tap,id=net0,ifname=${tap},script=no,downscript=no"
+DEVICE_OPTS="virtio-net-pci,netdev=net0,mac=${mac}"
+if [[ -n "${ROMFILE}" ]]; then
+	DEVICE_OPTS+=",romfile=${ROMFILE}"
+fi
+
+/usr/libexec/qemu-kvm \
+	"${ARCH_FLAGS[@]}" \
+	-pidfile "/tmp/vm-${idx}.pid" \
+	-m 3072 -smp 2 \
+	-accel kvm -cpu host \
+	-boot n \
+	-netdev "${NETDEV_OPTS}" \
+	-device "${DEVICE_OPTS}" \
+	-device virtio-rng-pci,rng=rng0 \
+	-object rng-random,id=rng0,filename=/dev/urandom \
+	-display none -vga none -machine graphics=off \
+	-chardev "stdio,id=char0,mux=on" \
+	-serial chardev:char0 \
+	-device virtio-serial-pci \
+	-device virtconsole,chardev=char0 \
+	-mon chardev=char0 \
+	< /dev/null 2>&1 |
+	sed -u "s/\x1b\[[0-9;]*[a-zA-Z]//g; s/\r//g" |
+	awk '{ printf "[%s %s]: %s\n", "'"${name}"'", strftime("%H:%M:%S"), $0; fflush() }' &
+
+echo "fake ipmitool: QEMU VM ${name} launched (pid=$!)"
+IPMIEOF
+
+chmod +x /usr/local/bin/ipmitool
+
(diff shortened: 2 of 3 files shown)


adrianreber:
See https://github.com/adrianreber/warewulf/actions/runs/25151879382/job/73724146427 for a successful run.

mslacken:
Great work Adrian.
Can you please update the changelog and rename `scripts/intergation-test.sh` to `scripts/intergation-test-el.sh`?
Thanks again.

adrianreber:
> Can you please update the changelog and rename `scripts/intergation-test.sh` to `scripts/intergation-test-el.sh`?

Done.

adrianreber:
I see:

> The action lima-vm/lima-actions/setup@v1 is not allowed in warewulf/warewulf because all actions must be from   a repository owned by warewulf, created by GitHub, verified in the GitHub Marketplace, or match one of the patterns: actions/checkout@v4, actions/download-artifact@v4, actions/setup-go@v5, actions/setup-python@v5, actions/upload-artifact@v4, dangoslen/changelog-enforcer@v3.


Should I replace it with  a copy of `lima-vm/lima-actions/setup@v1`. It shouldn't be to hard if that is desired.

mslacken:
@anderbubble Can you check this action and allow also actions from lima? I  think this is the right solution here.

.github/workflows/testsuite.yml
```diff
@@ -53,3 +53,26 @@ jobs:
...
+      - name: Install Lima
+        uses: lima-vm/lima-actions/setup@v1
+      - name: Cache Lima images
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/lima
+          key: lima-rocky-10-${{ github.sha }}
+          restore-keys: lima-rocky-10-
+      - name: Start Rocky Linux 10 VM
+        run: limactl start --plain --name=default --cpus=4 --memory=12 template://rocky-10
```
middelkoopt: Newer lima uses "template:rocky-10" (no slashes)

adrianreber:
@mslacken To extend this test (in another PR) to also run based on leap what image do you suggest for lima?

```shell
$ limactl start --list-templates | grep opensuse
experimental/opensuse-tumbleweed
opensuse
opensuse-leap
opensuse-leap-15
opensuse-leap-16
```

and which warewulf base image? `warewulf-leap:15`?

.github/workflows/testsuite.yml
```diff
@@ -53,3 +53,26 @@ jobs:
...
+      - name: Copy source into VM
```
middelkoopt: Wonder if it is worth making this and the "Start Rocky Linux 10 VM" scripts so folks can easily run this locally on their own systems.

middelkoopt:
Should the vm packages be updated first?

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,184 @@
...
+go mod vendor
+make defaults PREFIX=/usr SYSCONFDIR=/etc
+make build
+make install
+cp -f etc/warewulf.conf-el10 /etc/warewulf/warewulf.conf
+systemctl daemon-reload
+
+# Set up bridge and TAP interfaces for VMs
+ip link add br0 type bridge
+ip addr add 192.168.100.1/24 dev br0
```
middelkoopt: 192.168.100.1 is often the ip of a cable modem.  (I know it's internal)

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,184 @@
...
+wwctl image import docker://ghcr.io/warewulf/warewulf-rockylinux:10 \
+	rocky-10 --syncuser
+
+# Add compute nodes
+for ((i = 0; i < num_computes; i++)); do
+	wwctl node add --image=rocky-10 --profile=nodes --netname=default \
+		--ipaddr="${c_ip[$i]}" --hwaddr="${c_mac[$i]}" "${c_name[$i]}"
+done
+
+wwctl nodes set -y -A "crashkernel=no,net.ifnames=1,console=ttyS0,115200 earlyprintk=serial,ttyS0,115200" c[0-1]
```
middelkoopt: Perhaps set the kernel args on a profile, not the node directly?
```suggestion
wwctl profile set -y -A "crashkernel=no,net.ifnames=1,console=ttyS0,115200 earlyprintk=serial,ttyS0,115200" default
```

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,184 @@
...
+wwctl image build rocky-10
+wwctl overlay build
+wwctl configure --all
+
+# Launch QEMU VMs in the background
+# Pipe output through sed to strip ANSI escape sequences and prefix each line
+/usr/libexec/qemu-kvm \
+	-m 3048 -smp 2 -enable-kvm -boot n \
+	-netdev tap,id=net0,ifname=tap0,script=no,downscript=no \
+	-device virtio-net-pci,netdev=net0,mac="${c_mac[0]}" \
```
middelkoopt: I have found the hard way that random bad-things happen without the random seed generator.
```suggestion
	-device virtio-net-pci,netdev=net0,mac="${c_mac[0]}" \
	-device virtio-rng-pci \
```

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,184 @@
...
+	-nographic -serial mon:stdio -cpu host 2>&1 |
+	stdbuf -oL sed -u "s/\x1b\[[0-9;]*[a-zA-Z]//g; s/\r//g; s/^/[${c_name[0]}]: /" &
+VM0_PID=$!
+
+/usr/libexec/qemu-kvm \
+	-m 3048 -smp 2 -enable-kvm -boot n \
+	-netdev tap,id=net1,ifname=tap1,script=no,downscript=no \
+	-device virtio-net-pci,netdev=net1,mac="${c_mac[1]}" \
```
middelkoopt:
```suggestion
	-device virtio-net-pci,netdev=net1,mac="${c_mac[1]}" \
	-device virtio-rng-pci \
```

mslacken:
So for the vm host I sugggest: `opensuse-leap-16` and for the warewulf base image I always use `registry.opensuse.org/science/warewulf/leap-16.0/images/disk:latest`

adrianreber:
> Should the vm packages be updated first?

Added.

.github/workflows/testsuite.yml
middelkoopt:
```suggestion
        run: limactl start --yes --plain --name=default --cpus=4 --memory=12 --set '.nestedVirtualization=true' template:rocky-10
```

anderbubble:
Thanks, @adrianreber! My main concern is the time that it takes (looks like about 17m). What would you think of it being something triggered by dispatch rather than on pull request?

adrianreber:
> Thanks, @adrianreber! My main concern is the time that it takes (looks like about 17m). What would you think of it being something triggered by dispatch rather than on pull request?

Yes, it is longer than the others. I am used to long CI runs in the projects I am working on. 17 minutes would be short 🤣. I wouldn't mind but it would be a change. Up to you. I prefer more testing in PRs.

middelkoopt:
> Yes, it is longer than the others. I am used to long CI runs in the projects I am working on. 17 minutes would be short 🤣. I wouldn't mind but it would be a change. Up to you. I prefer more testing in PRs.

I'm working on performance.  Seeing much better numbers with enabling nested virtualization etc.  post results soon.

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,186 @@
...
+wwctl image import docker://ghcr.io/warewulf/warewulf-rockylinux:10 \
+	rocky-10 --syncuser
+
+# Add compute nodes
+for ((i = 0; i < num_computes; i++)); do
+	wwctl node add --image=rocky-10 --profile=nodes --netname=default \
+		--ipaddr="${c_ip[$i]}" --hwaddr="${c_mac[$i]}" "${c_name[$i]}"
+done
+
+wwctl nodes set -y -A "crashkernel=no,net.ifnames=1,console=ttyS0,115200 earlyprintk=serial,ttyS0,115200" c[0-1]
```
middelkoopt: Need to quote console= and earlyprintk= because they contain comma's.  Will fix in performance suggestion

adrianreber:
> > Yes, it is longer than the others. I am used to long CI runs in the projects I am working on. 17 minutes would be short 🤣. I wouldn't mind but it would be a change. Up to you. I prefer more testing in PRs.
> 
> I'm working on performance.  Seeing much better numbers with enabling nested virtualization etc.  post results soon.

Ah, didn't I enable KVM for the nested VMs?

middelkoopt:
> Ah, didn't I enable KVM for the nested VMs?

Not enabled for lima, you need to add  `--set '.nestedVirtualization=true'` - it's in the suggestion.  I'm also working on some tripple nested performance tweaks (lima is itself in a vm) and arm support.

adrianreber:
> > Ah, didn't I enable KVM for the nested VMs?
> 
> Not enabled for lima, you need to add  `--set '.nestedVirtualization=true'` - it's in the suggestion.  I'm also working on some tripple nested performance tweaks (lima is itself in a vm) and arm support.
> 

Thanks. Do the arm action runners support nested virtualization. Last time I checked they didn't.

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,186 @@
...
+wwctl configure --all
+
+# Launch QEMU VMs in the background
+# Pipe output through sed to strip ANSI escape sequences and prefix each line
+/usr/libexec/qemu-kvm \
+	-m 3048 -smp 2 -enable-kvm -boot n \
+	-netdev tap,id=net0,ifname=tap0,script=no,downscript=no \
+	-device virtio-net-pci,netdev=net0,mac="${c_mac[0]}" \
+	-nographic -serial mon:stdio -cpu host 2>&1 |
+	stdbuf -oL sed -u "s/\x1b\[[0-9;]*[a-zA-Z]//g; s/\r//g; s/^/[${c_name[0]}]: /" &
```
middelkoopt:
```suggestion
	sed -u "s/\x1b\[[0-9;]*[a-zA-Z]//g; s/\r//g; s/^/[${c_name[0]}]: /" &
```

Sed -u makes it unbuffered, double unbuffering.

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,186 @@
...
+VM0_PID=$!
+
+/usr/libexec/qemu-kvm \
+	-m 3048 -smp 2 -enable-kvm -boot n \
+	-netdev tap,id=net1,ifname=tap1,script=no,downscript=no \
+	-device virtio-net-pci,netdev=net1,mac="${c_mac[1]}" \
+	-nographic -serial mon:stdio -cpu host 2>&1 |
+	stdbuf -oL sed -u "s/\x1b\[[0-9;]*[a-zA-Z]//g; s/\r//g; s/^/[${c_name[1]}]: /" &
+VM1_PID=$!
```
middelkoopt: This is the sed process, not qemu-kvm

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,186 @@
...
+
+# Build and install Warewulf
+go mod vendor
+make defaults PREFIX=/usr SYSCONFDIR=/etc
+make build
+make install
+cp -f etc/warewulf.conf-el10 /etc/warewulf/warewulf.conf
+systemctl daemon-reload
+
+# Set up bridge and TAP interfaces for VMs
```
middelkoopt: Wrap in a loop

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,186 @@
...
+
+# Rebuild image, overlays, and reconfigure
+wwctl image build rocky-10
+wwctl overlay build
+wwctl configure --all
+
+# Launch QEMU VMs in the background
+# Pipe output through sed to strip ANSI escape sequences and prefix each line
+/usr/libexec/qemu-kvm \
+	-m 3048 -smp 2 -enable-kvm -boot n \
```
middelkoopt:
```suggestion
	-m 3072 -smp 2 -enable-kvm -boot n \
```

scripts/integration-test-el.sh
```diff
@@ -0,0 +1,186 @@
...
+	-netdev tap,id=net0,ifname=tap0,script=no,downscript=no \
+	-device virtio-net-pci,netdev=net0,mac="${c_mac[0]}" \
+	-nographic -serial mon:stdio -cpu host 2>&1 |
+	stdbuf -oL sed -u "s/\x1b\[[0-9;]*[a-zA-Z]//g; s/\r//g; s/^/[${c_name[0]}]: /" &
+VM0_PID=$!
+
+/usr/libexec/qemu-kvm \
```
middelkoopt: Wrap in a loop

middelkoopt:
> Thanks. Do the arm action runners support nested virtualization. Last time I checked they didn't.

No they don't, we would need to run some arm runner.

middelkoopt:
Ok my review is done - see my comments above (that gets most of them).  Overall my run was 20+ min, I think the slowest part was downloading the initramfs, it could be slower due to the use of the UEFI iPXE.  Here is my work:

https://github.com/middelkoopt/warewulf/commit/f69d76df02aa29fede3ccbe4b319ddec1732871d

@adrianreber feel free to take as much or as little of this and integrate it into your PR.

I tested the scripts on a x86_64 vm (Jetstream2), aarch (pi) on Linux and on Mac (m3 with lima) and on my own github account.

Overall I like the approach and approve in principal.  The approach is clean and I had been trying to do something similar, the key here is to run the nodes inside the head node!  This reduces the multiple layers w/o loosing much.

middelkoopt:
Networking probably could use some work if it is going to be used.

```console
[root@lima-default ~]# ssh 192.168.100.101
[root@c1 ~]# cat /etc/resolv.conf 
# This file is autogenerated by warewulf
[root@c1 ~]# ip route
default via 192.168.5.2 dev br0 proto static metric 100 
192.168.5.2 dev br0 proto static scope link metric 100 
192.168.100.0/24 dev br0 proto kernel scope link src 192.168.100.101 metric 100 
[root@c1 ~]# ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host proto kernel_lo 
       valid_lft forever preferred_lft forever
2: br0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:00:01:01 brd ff:ff:ff:ff:ff:ff
    inet 192.168.100.101/24 brd 192.168.100.255 scope global noprefixroute br0
       valid_lft forever preferred_lft forever
```

adrianreber:
@middelkoopt I applied your changes. I added back `stdbuf` to have better line based log output, I also reduced the loglevel of the kernel to getter faster boottimes.  I didn't switch to UEFI for x86_64 because that seems to make booting extra slow. The selection in the ipxe bootmenu does not work and it first boots using imgextract, that fails and then it boots a second time with the initrd.

In my tests nested virtualization is always turned on using lima. Not sure what the difference is.

@anderbubble concerning timing, the first 8 minutes are spent setting up the VM and compiling warewulf and about 8 minutes to start the two VMs with warewulf images.

For some reason the whole thing is much slower in github actions than locally.

Switching from serial console to hvc0 brings the whole run down to 13 minutes.

Please take another look.

adrianreber:
> Networking probably could use some work if it is going to be used.

I ignored for now. I guess if we really want to install packages a proxy could be the easiest way to download things.

adrianreber:
Just for comparison. Booting the VM on my local VM->container->lima->qemu setup takes 40 seconds. In Github Actions it takes 320 seconds. Not sure where this difference comes from. My setup should have the same level of nesting as GitHub Actions plus one container.

middelkoopt:
> > Networking probably could use some work if it is going to be used.
> 
> I ignored for now. I guess if we really want to install packages a proxy could be the easiest way to download things.

Or run NAT on the head node.  This is what I use on my systems.

```bash
## Enable IP forwarding and NAT so compute nodes can reach external network via head node
sms_eth_external=$(ip -j route show default | jq -r '.[0].dev')
echo "net.ipv4.ip_forward = 1" > /etc/sysctl.d/80-ip_forward.conf
sysctl -p /etc/sysctl.d/80-ip_forward.conf
dnf -y install nftables
cat > /etc/sysconfig/nftables.conf <<EOF
table ip nat {
        chain POSTROUTING {
                type nat hook postrouting priority 100; policy accept;
                oifname "${sms_eth_external}" masquerade
        }
}
EOF
systemctl enable --now nftables
```

resolv.conf will need to be fixed and the GW set to 192.168.100.1

Ignoring is a fine option (perhaps a note in the script would be good where you "setup" networking - I would keep that stuff there so it's easier to get working).

middelkoopt:
> Switching from serial console to hvc0 brings the whole run down to 13 minutes.

Glad you independently found that one.  I had implemented it and noted that it was not that much faster, so dropped it.  I only visually tested on boot, not install. I failed to consider that a full install would compound over time.  I've put this one in my back pocket as I've had other situations where the serial console is laggy.

middelkoopt:
> Just for comparison. Booting the VM on my local VM->container->lima->qemu setup takes 40 seconds. In Github Actions it takes 320 seconds. Not sure where this difference comes from. My setup should have the same level of nesting as GitHub Actions plus one container.

I saw similar differences as well.  I wonder if the github network is throttled -- ah I see you cached the image.

Gemini had the following suggestions that might be worth considering.
1. don't over-provision the runner (assuming we are using a runner with less than 12 GB) limactl start ... --cpus=2 --memory=6
2. Don't compress the initramfs (I doubt this one)
3. mount /var/lib/warewulf on a tmpfs

it also said drop the bridge to MTU 1450, but I don't believe it as well.

middelkoopt (approved):
Looks good!

middelkoopt:
I'm good with how it stands. Feel free to improve networking and/or performance (or not from my perspective).

adrianreber:
> also said drop the bridge to MTU 1450, but I don't believe it as well.

Claude said to try 9000, but that also doesn't change anything.

Let's see if @anderbubble wants to integrate it at all. Then we could merge it for now and do follow ups to improve it.

adrianreber:
Wow, just saw your comments now. Haven't seem them before. Strange.

anderbubble:
I need to try running it myself, but I definitely want to integrate this by intent.

I do want to separate it into its own workflow, though. I'm thinking we could make it happen on merge to main and on dispatch. Right now the on_dispatch is in addition to on pull request, so it will still run every time.

adrianreber:
I changed it a bit and it is now faster than the other tests. 5 minutes only. I removed one layer of nested virtualization. I dropped the lima VM and I am now running in a container. My first attempts with containers failed and that is why I went with the lima VM. Using a container with systemd in it solved my problems.

@anderbubble @mslacken Please take another look.

PR #2180, opened 2026-05-07
Add Leap support to integration test and matrix CI

adrianreber:

Detect the distribution via /etc/os-release and use zypper for Leap and dnf for EL. Install distribution-specific packages, copy the correct warewulf.conf, and use the appropriate compute node image. Add the OBS utilities repository for yq on Leap. Find qemu-kvm in both /usr/libexec and /usr/bin. Use uname -n instead of hostname for SSH reachability checks. Set dnsmasq as DHCP systemd service name and remove dsa from SSH key types. Rename the script from integration-test-el.sh to
integration-test.sh and add a matrix build for Rocky Linux 10 and SUSE Linux 16.0.

## Reviewer  checklist

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -57,12 +57,19 @@ jobs:
 
   integration-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - name: Rocky Linux 10
+            container: rockylinux/rockylinux:10-ubi-init
+          - name: SUSE Linux 16.0
+            container: registry.suse.com/bci/bci-init:16.0
     steps:
       - name: Checkout Warewulf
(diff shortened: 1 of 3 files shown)


adrianreber:
@mslacken @middelkoopt Extended the integration test to leap.

mslacken (approved):

mslacken:
@anderbubble  I think it would also possible to create a `qemu.tmpl` file in 
```lib/warewulf/bmc/```
which would then be available with `wwctl node power $CMD`.
This would make warewulf more or less make a mini proxmox...
Problem here is that this `tmpl` files are not rendered by the normal template engine (as far as I remember), if we change that it would be possible to set VM properties with tags...
This is not meant as a suggestion or critic, just something which came to my mind...
Thanks again @adrianreber

PR #2198, opened 2026-05-31
Fix kernel version detection for kernels with a version-like suffix after the dist tag

### LINKED ISSUE #2199: Image with multiple kernels does not default to booting off the highest versioned kernel
### Steps to reproduce

Three kernels are installed in an image:
```
$ ls -l rockylinux-9-chemistry/rootfs/boot/vmlinuz-*
-rwxr-xr-x. 1 root root 15117144 May 13 11:40 rockylinux-9-chemistry/rootfs/boot/vmlinuz-5.14.0-611.55.1.el9_7.0.1.x86_64
-rwxr-xr-x. 1 root root 15117144 May 18 20:47 rockylinux-9-chemistry/rootfs/boot/vmlinuz-5.14.0-611.55.1.el9_7.0.3.x86_64
-rwxr-xr-x. 1 root root 15174472 May 29 13:09 rockylinux-9-chemistry/rootfs/boot/vmlinuz-5.14.0-687.12.1.el9_8.x86_64
```
The newest has version number 5.14.0-687.12.1 so I'd expect that to be the default, but the nodes boot with the older kernel 5.14.0-611.55.1.el9_7.0.3 unless I set the kernel version for the nodes explicitly in nodes.conf.

### Error message

```
$ wwctl image kernels rockylinux-9-chemistry
Image                   Kernel                                                 Version          Default  Nodes
-----                   ------                                                 -------          -------  -----
rockylinux-9-chemistry  /boot/vmlinuz-5.14.0-611.55.1.el9_7.0.1.x86_64         7.0.1            false    0
rockylinux-9-chemistry  /boot/vmlinuz-5.14.0-611.55.1.el9_7.0.3.x86_64         7.0.3            true     0
rockylinux-9-chemistry  /boot/vmlinuz-5.14.0-687.12.1.el9_8.x86_64             5.14.0-687.12.1  false    15
rockylinux-9-chemistry  /lib/modules/5.14.0-611.55.1.el9_7.0.1.x86_64/vmlinuz  7.0.1            false    0
rockylinux-9-chemistry  /lib/modules/5.14.0-611.55.1.el9_7.0.3.x86_64/vmlinuz  7.0.3            false    0
rockylinux-9-chemistry  /lib/modules/5.14.0-687.12.1.el9_8.x86_64/vmlinuz      5.14.0-687.12.1  false    0

```
The version for the 5.14.0-611.55.1.el9_7.0.1 and 5.14.0-611.55.1.el9_7.0.3 kernels isn't what I expected, and the default isn't the 5.14.0-687.12.1.el9_8 kernel.

```$ wwctl version
wwctl version:   4.6.0-1
rpc version: apiPrefix:"rc1" apiVersion:"1" warewulfVersion:"4.6.0-1"
$ cat /etc/os-release
NAME="Rocky Linux"
VERSION="9.8 (Blue Onyx)"
RELEASE_TYPE="stable"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="9.8"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Rocky Linux 9.8 (Blue Onyx)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:9::baseos"
VENDOR_NAME="RESF"
VENDOR_URL="https://resf.org/"
SUPPORT_END="2032-05-31"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-9"
ROCKY_SUPPORT_PRODUCT_VERSION="9.8"
REDHAT_SUPPORT_PRODUCT_VERSION="9.8"
```

### General information

guojx1998:
### Problem

`internal/pkg/util/version.go` `ParseVersion` returns the **last** `\d+\.\d+\.\d+(-[\d\.]+|)` match found in the kernel path. Some kernel RPMs carry a version-like suffix in the release field _after_ the dist tag — for example `5.14.0-687.10.1.el9_8.0.1.x86_64`, where `el9_8.0.1` yields a second, spurious match `8.0.1` to the right of the real release `5.14.0-687.10.1`. Because the loop returns the last match, `wwctl image kernels` and the `.Kernel.Version` overlay variable report `8.0.1`. The same shape affects e.g. `el10_2.0.1` (→ `2.0.1`). Kernels without such a suffix produce a single match and are unaffected, which is why this went unnoticed.

This also breaks default-kernel selection: an image with `el9_7.0.1` / `el9_7.0.3` / `el9_8` kernels parses the first two as `7.0.1` / `7.0.3`, which `go-version` ranks above the genuinely newer `5.14.0-687.12.1` (major `7 > 5`), so `kernel.collection.Default()` boots the older kernel.

### Fix

Iterate the matches forward and return the **first** valid version. The real kernel release is always the leftmost version token in the search-glob paths (`/boot/vmlinuz-<ver>`, `/lib/modules/<ver>/vmlinuz`); any dist-tag fragment is to its right.

### Tests

*   Added `internal/pkg/util/version_test.go` (there was no test for `ParseVersion`) covering plain `el9_4`/`el8_6` paths, paths with a version-like suffix after the dist tag (`el9_8.0.1`, `el10_2.0.1`), a plain `el9_8` path, a `.gz` path, a path with no version, and an `el9_7` respin case using the exact path from #2199. The suffix cases fail before this change and pass after.
*   Existing `internal/pkg/kernel` tests are unaffected (their paths have a single version match).

### Note for reviewers

I changed the selection order rather than the regex. I couldn't determine the original rationale for iterating in reverse; if there was a case that required it, an alternative is to select the longest match — happy to adjust.

Also fixes #2199.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Kernel version detection for kernels whose RPM release field contains a
   version-like suffix after the dist tag (e.g. `5.14.0-687.10.1.el9_8.0.1`).
   These were mis-detected (e.g. `8.0.1` instead of `5.14.0-687.10.1`) because
-  `ParseVersion` returned the last regex match found in the kernel path.
+  `ParseVersion` returned the last regex match found in the kernel path. This
+  also caused the wrong kernel to be selected as the default. #2199
 
 ## v4.7.0, 2026-05-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Remove `dsa` from default `ssh: key types`; sshd silently skips DSA host keys
   on EL9 / OpenSSH 8.7p1+, leaving nodes with no usable host keys. #1185
 
+### Fixed
+
+- Kernel version detection for kernels whose RPM release field contains a
+  version-like suffix after the dist tag (e.g. `5.14.0-687.10.1.el9_8.0.1`).
+  These were mis-detected (e.g. `8.0.1` instead of `5.14.0-687.10.1`) because
+  `ParseVersion` returned the last regex match found in the kernel path.
+
 ## v4.7.0, 2026-05-12
 
 ### Fixed
--- a/internal/pkg/util/version.go
+++ b/internal/pkg/util/version.go
@@ -17,8 +17,8 @@ func init() {
 
 func ParseVersion(versionString string) *version.Version {
 	matches := versionPattern.FindAllString(versionString, -1)
-	for i := len(matches) - 1; i >= 0; i-- {
-		if version_, err := version.NewVersion(strings.TrimSuffix(matches[i], ".")); err == nil {
+	for _, match := range matches {
+		if version_, err := version.NewVersion(strings.TrimSuffix(match, ".")); err == nil {
 			return version_
 		}
 	}
--- a/internal/pkg/util/version_test.go
+++ b/internal/pkg/util/version_test.go
@@ -12,7 +12,9 @@ func TestParseVersion(t *testing.T) {
 		"el10_2 with suffix after dist": {"/lib/modules/6.12.0-211.16.1.el10_2.0.1.x86_64/vmlinuz", "6.12.0-211.16.1"},
 		"el9_8 with suffix after dist":  {"/lib/modules/5.14.0-687.10.1.el9_8.0.1.x86_64/vmlinuz", "5.14.0-687.10.1"},
 		"el9_8 plain dist tag":          {"/lib/modules/5.14.0-687.10.1.el9_8.x86_64/vmlinuz", "5.14.0-687.10.1"},
+		"el9_7 respin suffix (#2199)":   {"/boot/vmlinuz-5.14.0-611.55.1.el9_7.0.3.x86_64", "5.14.0-611.55.1"},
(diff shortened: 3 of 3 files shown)


anderbubble:
@guojx1998 thanks for the PR! I'll plan to review.

anderbubble:
I think this fixes #2199

guojx1998:
As @anderbubble noted, this also resolves #2199 — the same root cause surfacing as wrong default-kernel selection. I verified it end-to-end locally against the real `kernel.collection.Default()` (which sorts via `collection.Less` → `(*Kernel).version()` → `util.ParseVersion`), using the reporter's three kernels:

| path | parsed on `main` (last match) | parsed in this PR (first match) |
| --- | --- | --- |
| `…-5.14.0-611.55.1.el9_7.0.1.x86_64` | `7.0.1` | `5.14.0-611.55.1` |
| `…-5.14.0-611.55.1.el9_7.0.3.x86_64` | `7.0.3` | `5.14.0-611.55.1` |
| `…-5.14.0-687.12.1.el9_8.x86_64` | `5.14.0-687.12.1` | `5.14.0-687.12.1` |

The middle column matches the `Version` column in the issue's `wwctl image kernels` output exactly. `go-version` ranks `7.0.3` above `5.14.0-687.12.1` (major `7 > 5`), so `Default()` returns the `el9_7.0.3` kernel — the reported behavior. Flipping only the `ParseVersion` loop direction makes all three parse to their true release, and `Default()` then returns the `el9_8` kernel as expected. Added a regression test using the issue's exact path.

mslacken (approved):

mslacken:
Thx

PR #2203, opened 2026-06-10
wwclient unitfile update

buzh:

Update unit file to avoid systemd terminating wwclient if startup is slow, f.ex if networking takes a while to be ready.

wwclient has internal retry logic, but the previous unit config didn't take this into account and would just kill the process after 60 seconds regardless, due to TimeoutSec being set.

This PR changes the behaviour to wait forever on startup and allow up to 30s for service stops.

It also adds Restart=on-failure so that if wwclient crashes or fails it won't require manual intervention to come back up should the cause of failure disappear.  Stopping or killing wwclient will _not_ trigger a restart.

In addition, the obsolete pidfile directive is removed and a documentation link is provided

## Reviewer  checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- wwclient.service changed to allow internal retry logic to continue forever,
+  instead of being terminated after 60s if startup is slow (f.ex temporary
+  network failure). Attempt restart if wwclient terminates abnormally.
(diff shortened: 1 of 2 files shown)


anderbubble:
Thanks for the PR! The test suite is failing because of an upstream break in the container build process. That's fixed in a different PR, so I'll rebase this on that once it's in.

anderbubble:
> Wants=network.target — systemd.special(7) documents this as incorrect usage; network.target is passive and pulling it in with Wants= does nothing useful.

anderbubble (approved):

Issue #986, opened 2023-11-21
add the ability to ssh against a profile


anderbubble:
### Summary

`wwctl ssh [profile_name]` should allow running a command against all nodes with a given profile.

### Rationale

Allows treating profiles as groups.

### General information

- [X] I have searched the issues of this repo and believe this is not a duplicate

mslacken:
Closing this should also include the ability to use profiles as selection criteria for `wwctl node list`

anderbubble:
See also #835.

anderbubble:
I'm moving this to 4.6 for the general functionality, but we may/should still generate clustershell and/or pdsh config, and not try to replace either with Warewulf.

anderbubble:
I think we should re-scope this to include use for other commands, not just ssh.

PR #2204, opened 2026-06-12
Adds nodegroups for targeting sets of nodes in wwctl

buzh:

This proposes introducing a nodegroups concept to hold sets of nodes so that wwctl commands can use those instead of explicit ranges.  Nodegroups can be declared in three ways: as a top-level stanza in nodes.conf, explicitly for individual nodes, or in profiles (causing nodes using that profile to inherit the group).

This is somewhat related to #986. While not explicitly including the ability to do `wwctl ssh @someprofile` you could add a group to the profile and achieve the same result. It will also make it easier to generate config files for `clush` and others using a single source of truth regarding group membership.

Passing "@groupname" as a nodelist resolves group members in most wwctl commands, f.ex `wwctl power reset @rack1`.

All nodes are members of the built-in default group `@all`, unless explicitly masked by using ~all. Nodes and profiles can also negate themselves, so you could for example add a profile "butterfingers" and add an "~all" nodegroup, then add the profile to any nodes that should not be touched by wwctl'ing against `@all`.

Further details in [userdocs/nodes/nodegroups.rst](https://github.com/warewulf/warewulf/compare/main...buzh:warewulf:nodegroups?expand=1#diff-a966a22aca6ef8986df292419218ec61f1e117aa78c50eb84a492a02ac28da15)

I realize this has wider consequences and needs a proper review, but I'm putting up this PR as a starting point for that discussion

## Reviewer  checklist


anderbubble:
Thanks for the PR, @buzh.

I'm curious what you think about the value of this vs just using @group syntax to refer to "the nodes that include this profile" and using profiles for this rather than a new groups concept.

If we do add "nodegroups" as described here, a nitpick might be that in the yaml they should just be "groups"; rather than repeating "node" as a prefix. That could just be my aesthetic sensibilities, though.

I'll look at it some more.

buzh:
Thanks for considering it!

Personally I think it's worth adding groups as a concept just to have something that is _only_ that; a set of nodes. You could get a similar result by adding profiles as pseudo-groups, e..g "rack1" or "gpu", but if you're only using the profile for grouping nodes I would argue it clutters the list of profiles _and_ the list of groups (f.ex for tab completing @-names or generating config for oob tools like clush). Plus - it could "facilitate" unintended grouping down the line as profile evolve, especially in cases where there's multiple admins, staff changes and so on. 

In my mind it's better if it's completely obvious and unambiguous what `wwctl command @foo` hits, even if the person running the command isn't fully conscious about what the profile hierarchy looks like.

An option for profile groups could  be to add some other method to populate by calling for @profile_someprofile, %someprofile, p:someprofile or similar to resolve the nodes by profiles rather than nodegroup, to get the full functionality of both approaches.

As for "nodegroups" vs "groups", I originally had it as "groups" in node defs, but figured that one term is preferable to two - and I didn't want to use just "groups" for both since this is only ever a group of nodes and shouldn't be confused with any other type of group. Not a hill I want to die on, but for me this is more intuitive despite the extra bytes.

buzh:
I changed the command to `wwctl node group` instead of adding `nodegroup` to the top level. Feels more in line with the existing structure

anderbubble:
@buzh I'll be including this PR in our TSC agenda this week. Let me know if you want to join to discuss it.

I think we should just do `wwctl group` and just call these groups rather than node groups. For now, nodes should be implicit, since that's the base entity that everything is about anyway; but if we add, say, networks in the future, I don't see any reason we shouldn't allow those to be grouped as well. Then we could just use context to indicate what kind of entity we're dealing with, without having to either rename later or invent parallel entities.

buzh:
Thanks @anderbubble. Not sure if I'll be able to join the TSC meeting because of the TZ, but I'll show if I can. I'm already on the calendar invite.

anderbubble:
Feedback from TSC:

- We recommend just `groups` rather than `nodegroups` (with subcommands likely at `wwctl group[s]`)
- We recommend consolidating down to a single definition style, with the second syntax preferred:

```
   nodeprofiles:
     gpu:
       nodegroups:
         - gpu-nodes
   nodes:
     n01:
       profiles:
         - gpu
     n02:
       nodegroups:
         - admin
         - rack3
```

We think this is less likely to cause confusion between groups and profiles. (We could always add the second syntax later if necessary; but better to start simpler for now.)

buzh:
Allright, I've watched the recording and agree that adding the top level could be problematic down the line. I've revised the PR to remove the top level groups stanza, instead enumerating profiles and nodes. The functionality is unchanged.

I also did some other minor polish, including a "-n/--noheader" option to return a comma separated list instead of a table

This is the new behaviour:
```
# No user-defined groups yet:
[root@wwtest01 ~]# wwctl group ls
GROUP  MEMBERS
-----  -------
all    head01,head02,login01,login02,n02,n03,n04,n05,n06,n07,n08

# Add group via profile
[root@wwtest01 ~]# wwctl profile add headnodes --groups admin
[root@wwtest01 ~]# wwctl node set -y head0[1-2] --profile headnodes
[root@wwtest01 ~]# wwctl group ls
GROUP  MEMBERS
-----  -------
all    head01,head02,login01,login02,n02,n03,n04,n05,n06,n07,n08
admin  head01,head02

# Exclude profile members from @all:
[root@wwtest01 ~]# wwctl profile set -y headnodes --groups admin,~all
[root@wwtest01 ~]# wwctl group ls
GROUP  MEMBERS
-----  -------
all    login01,login02,n02,n03,n04,n05,n06,n07,n08
admin  head01,head02

# Use @group syntax with other wwctl command, f.ex. node set:
[root@wwtest01 ~]# wwctl node set -y @admin --kernelargs "usbcore.blinkenlights=1"
[root@wwtest01 ~]# wwctl node ls -a head01
NODE    FIELD        PROFILE    VALUE
----    -----        -------    -----
head01  Profiles     --         headnodes
head01  Groups       headnodes  admin,~all
head01  Kernel.Args  --         usbcore.blinkenlights=1

# Profile structure:
[root@wwtest01 ~]# wwctl profile ls -y headnodes
headnodes:
    groups:
        - admin
        - ~all

# Node structure:
[root@wwtest01 ~]# wwctl node ls -y head01
head01:
    profiles:
        - headnodes
    groups:
        - admin
    kernel:
        args:
            - usbcore.blinkenlights=1
```

Is this in line with what you had in mind?

mslacken (approved):

PR #2205, opened 2026-06-15
build: default source-install PREFIX to /opt/warewulf

### LINKED ISSUE #2094: change default source-install directories to /opt/warewulf
We should update the default source-install directories to install Warewulf to /opt/warewulf so that it's easier to manage for people who install from source.

shravyavorugallu:
Closes #2094

When Warewulf is built from source with `make install` the current default `PREFIX` is `/usr/local`. On systems where Warewulf was also installed via an OS package (RPM or DEB) this causes the source-built binaries and config files to overwrite or conflict with package-manager-tracked files.

**Change**
When `USE_LSB_PATHS` is not set (i.e. no OS type was detected, which is the case for a plain source build), the default `PREFIX` is now `/opt/warewulf`. All Warewulf files then install under a single self-contained directory, making it easy to inspect and remove a source build without touching system paths.

Distro packagers and users who prefer the traditional layout are unaffected: RPM/DEB builds set `USE_LSB_PATHS` before `PREFIX` is evaluated, so they continue to use `/usr/local` (or their own `PREFIX` override).

Users can still override explicitly:
```
make PREFIX=/usr/local install
```

### DIFF
--- a/Variables.mk
+++ b/Variables.mk
@@ -28,7 +28,15 @@ endif
 
 # System directory paths
 VARLIST += PREFIX BINDIR SYSCONFDIR SRVDIR DATADIR MANDIR DOCDIR LOCALSTATEDIR RELEASE CACHEDIR
-PREFIX ?= /usr/local
+# When building from source on a system without an OS-release-detected install
+# (i.e. not a distro package build), default to /opt/warewulf to avoid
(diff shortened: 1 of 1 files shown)


anderbubble:
Thanks for the PR! We'll review.

mslacken (approved):

PR #2207, opened 2026-06-17
Fix breaking Dockerfile build due to upstream change

anderbubble:

The devel_basis pattern requires patterns-devel-base-devel_basis, which pulls in the real GNU gawk package. But the Tumbleweed base image now ships busybox-gawk (part of the recent move to busybox for base utilities), which already provides the gawk capability. zypper sees a conflict (two packages can't both own /usr/bin/gawk) and won't auto-swap one for the other in non-interactive mode, so it aborts.

## Reviewer  checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM opensuse/tumbleweed:latest AS builder
 
-RUN zypper  -n install --no-recommends git go1.25 unzip &&\
+RUN zypper  -n install --no-recommends gawk git go1.25 unzip &&\
   zypper -n install -t pattern devel_basis
 
 # now build the warewulf


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM opensuse/tumbleweed:latest AS builder
 
 RUN zypper  -n install --no-recommends git go1.25 unzip &&\
-  zypper -n install -t pattern devel_basis
+  zypper -n install --force-resolution -t pattern devel_basis
```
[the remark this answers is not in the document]
anderbubble: I leave this to the reviewer to decide; this is working now, and I avoided the package-removal solution before because while that's the issue _now_, it won't necessarily be the issue again in the future; so for our purposes, this seemed like it was more likely to prevent future breaking.

Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM opensuse/tumbleweed:latest AS builder
 
 RUN zypper  -n install --no-recommends git go1.25 unzip &&\
```
middelkoopt:
```suggestion
RUN zypper  -n install --no-recommends gawk git go1.25 unzip &&\
```

and drop the `--force-resolution`.  `busybox-gawk` is pulled in by the installation of git/go so specifying it explicitly installs it first, no need to uninstall it and install gawk with the `--force-resolution`

Dockerfile
middelkoopt:
```suggestion
  zypper -n install -t pattern devel_basis
```

middelkoopt: Removed the proposed `--force-resolution`

middelkoopt:
Updates tested on my local github fork on the main branch.  Dockerfile also builds clean locally.  (add gwak, remove force resolution)

middelkoopt (changes requested):

anderbubble:
@middelkoopt accepted and incorporated. Ready for re-review and (hopefully) merge.

PR #2230, opened 2026-07-28
Fix ipv6_method tag being ignored when a static IPv6 address is set

ssimpson89:

PR #2217 added an opt-in `ipv6_method` netdev tag so a node could set the NetworkManager `[ipv6]` method directly, mainly for `auto` (SLAAC).

The tag was only checked after the `if $netdev.Ipaddr6` branch, so any node with a static IPv6 address took `method=manual` unconditionally and the tag was silently ignored, exactly the case for a node that has a deterministic IPv6 address but still needs its default route from RA.

This PR reorders the `[ipv6]` block so `ipv6_method` takes precedence. When set, it drives `method=`, and if the node also has an `ipaddr6`, that's still emitted as a static `addresses=`/`gateway=` line alongside it. NetworkManager treats `auto` plus a static address as additive, so this keeps the known address while restoring RA processing for the
default route.

## Reviewer  checklist


middelkoopt:
There should be some unit tests for this.

ssimpson89:
> There should be some unit tests for this.

Test has been added!

anderbubble:
@ssimpson89 the test suite is failing; but it looks like it's because of an outdated license file. Maybe it just needs to be rebased to main? Cause it doesn't look like you changed any dependencies.

If rebasing doesn't fix it, `make licensees` should.

ssimpson89:
@anderbubble Interesting, neither fixed it in my local environment but i did rebase and force a push so lets see if that does anything

anderbubble (approved):

anderbubble:
Thanks, Stephen!

PR #2235, opened 2026-08-03
added machine-id

### LINKED ISSUE #2234: Handling /etc/machine-id
See also: #1948  #1340 #2018  https://github.com/warewulf/warewulf-node-images/pull/86 
Search related: https://github.com/warewulf/warewulf/issues?q=is%3Aissue%20%22machine-id%22

This came up because it turns out all of our nodes have the same machine_id, or the machine ID from the last image build process.

Should:
1. `/etc/warewulf/image_exit.sh` be called during `wwctl image import`? 

   Derived containers adding packages may not always call this themselves, and if images are never `wwctl image shell` entered, the script is never run. 

2. An actual peristent machine ID be generated by Warewulf?

   The "identity" of a warewulf controlled system is ultimately the hostname as configured in `nodes.conf`. It's the key in the yaml map. Ideally, the ID would be managed by WW.

   The machine-id could be generated on node-add, wwctl upgrade and simply kept in nodes.conf. Or it could be generated on the fly, e.g. by hashing some secret (warewulf host's machine ID, some secret in the warewulf.conf ...) with the hostname.

   One benefit to putting the ID under WW control is that e.g. the dracut boot stage can be given the correct machine id so that the logs actually work and are assigned to the right host. Also, it's just indeed one of those "under the radar" things.

mslacken:
added a machine-id which is created when the node is added or upgrade config is run
close #2234

### DIFF
--- a/overlays/wwinit/rootfs/etc/machine-id.ww
+++ b/overlays/wwinit/rootfs/etc/machine-id.ww
@@ -1,0 +1,3 @@
+{{- if .MachineId -}}
+{{ .MachineId }}
+{{- end }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Support for generating, persisting, and querying a unique node-specific `/etc/machine-id` inside the `wwinit` system overlay.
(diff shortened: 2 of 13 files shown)


overlays/wwinit/rootfs/etc/machine-id.ww
anderbubble, line 1: Maybe don't write machine-id if there isn't a `.MachineId` value.

PR #2236, opened 2026-08-05
Propagate wwinit.d script failures instead of discarding them

metalllinux:
## Issue I observed

I observed that a provisioning script that fails in `/warewulf/wwinit.d/` produces no diagnostic anywhere in the boot.

The boot carries on past it, and then either stops at a later step that names the wrong thing, or comes up on a disk that does not match the node's configuration.

I encountered this whilst working through a two-stage provision-to-disk problem in a customer's environment - the hardware was an aarch64 NVIDIA GH200 node on Warewulf 4.7.0 using IPv6-only.

The node was configured with `format: ext4` on partitions that already held xfs, with no wipe requested, which I understand Ignition declines by design.

Nothing on the console, in the dracut log, or in the node's journal said that Ignition had declined the wipe.

The two visible outcomes, depending on what is already on the disk:

- The partition holds a leftover filesystem - `mount` has no `-t`, so it mounts whatever is actually there, the boot completes, and the node runs on a filesystem that contradicts its own configuration.
- The partition holds nothing yet - the boot dies at the mount, and the only thing the user sees is the boot falling over, which points at the mount rather than at the provisioning step that was supposed to create the filesystem:

```
mount: /sysroot: no filesystem type specified.
dracut: FATAL: warewulf: failed to mount /dev/disk/by-partlabel/rootfs at /sysroot
dracut: Refusing to continue
```

To highlight, that second capture is from my own lab reproduction of the same configuration, not from the customer's node.

There is a third outcome on the single-stage path, which I measured later and which has no mount to fail at all.

With `root: tmpfs` the root file system is built in RAM, so a failed `wwinit.d` script is discarded and the node boots through to a login prompt as though the step had succeeded.

The script's own message does reach the console on this path, because `init` is PID 1 and its children's stderr is the console, but nothing treats the failure as fatal and nothing stops the boot.

The root cause below is from reading the tree on `main` at `bf910c8d`. The validation section is measured when nodes are booted.

## Root cause

I believe the root cause is split into three layers. Each is described below:

### 1. `run-wwinit.d` does not test the status, and its pipeline hides all but the last script

`overlays/wwinit/rootfs/warewulf/run-wwinit.d`, the whole of it (18 lines), with the loop at 12-15:

```sh
scriptdir="${PREFIX}/warewulf/wwinit.d"
if [ -d "${scriptdir}" ]; then
    info "warewulf: running scripts in ${scriptdir}..."
    ls -1 "${scriptdir}/" | while read -r name; do
        info "warewulf: ${name}"
        PREFIX=$PREFIX sh "${scriptdir}/${name}"
    done
else
    info "warewulf: ${scriptdir} not found"
fi
```

`$?` is never examined, so the loop runs on to the next script regardless.

Because the loop is also the last element of a pipeline, the only thing that survives at all is the final script.

A failure in any earlier position is discarded outright, and the following scripts still run against whatever state the failed one left behind.

I understand the ordering does matter. The shipped scripts are numbered as a dependency chain.

`10-sfdisk.sh` partitions the disk and `20-mkfs.sh` formats those partitions. As it stands, `10-sfdisk.sh` can fail and `20-mkfs.sh` will still be handed the device.

### 2. Both call sites ignore the return value

`dracut/modules.d/90wwinit/load-wwinit.sh:64-66`:

```sh
if [ -x /tmp/wwinit/warewulf/run-wwinit.d ]; then
        PREFIX=/tmp/wwinit /tmp/wwinit/warewulf/run-wwinit.d
fi
```

`overlays/wwinit/rootfs/init:35-37`:

```sh
if [ -x /warewulf/run-wwinit.d ]; then
        /warewulf/run-wwinit.d
fi
```

Both surrounding scripts have a `die` available and use it freely elsewhere.

`load-wwinit.sh` guards the stage fetch with `|| die "Unable to load stage: system"` on line 62 and the root mount with `|| die` on line 75, so this is the one step in that sequence that is not guarded.

### 3. How the scripts are written

Every one of the four `wwinit.d` scripts in the tree defines the same `die` fallback and calls it when its core operation fails:

| Script | Failure handling |
|---|---|
| `overlays/ignition/rootfs/warewulf/wwinit.d/10-ignition.sh:26` | `die "warewulf: ignition: failed to partition/format disk"` |
| `overlays/sfdisk/rootfs/warewulf/wwinit.d/10-sfdisk.sh.ww:37` | `... \|\| die "warewulf: sfdisk: failed to partition {{ $device.device }}"` |
| `overlays/mkfs/rootfs/warewulf/wwinit.d/20-mkfs.sh.ww:68` | `... \|\| die "warewulf: mkfs: failed to format {{ $fs.device }}"` |
| `overlays/mkswap/rootfs/warewulf/wwinit.d/20-mkswap.sh.ww:60` | `... \|\| die "warewulf: mkswap: failed to format {{ $fs.device }}"` |

Each also carries this same preamble, for example at `10-ignition.sh:11-16`:

```sh
if ! command -v die >/dev/null; then
    die() {
        printf '%s\n' "$*" >&2
        exit 1
    }
fi
```

Because `run-wwinit.d` invokes each script as a separate `sh` process, dracut's own `die` is never in scope, so `command -v die` fails and the local fallback is what runs.

That fallback is a message on stderr and `exit 1`, which is then dropped by layer 1 or ignored by layer 2.

So the intent is unambiguous and stated four times, and none of it takes effect.

What that costs depends on the boot path, and on the single-stage path it is not a later mount failure.

Two-stage reaches a mount that may have nothing to mount.

Single-stage with `root: tmpfs` never mounts anything from the disk, so the failure has nothing to trip over and the node provisions to a login prompt regardless. Both are measured below.

## What I suggest to change

`run-wwinit.d` becomes a `for` loop in the current shell that reports and spreads the first failure:

```diff
--- a/overlays/wwinit/rootfs/warewulf/run-wwinit.d
+++ b/overlays/wwinit/rootfs/warewulf/run-wwinit.d
@@ -6,12 +6,25 @@ if ! command -v info >/dev/null; then
     }
 fi
 
+if ! command -v warn >/dev/null; then
+    warn() {
+        printf '%s\n' "$*" >&2
+    }
+fi
+
 scriptdir="${PREFIX}/warewulf/wwinit.d"
 if [ -d "${scriptdir}" ]; then
     info "warewulf: running scripts in ${scriptdir}..."
-    ls -1 "${scriptdir}/" | while read -r name; do
+    for script in "${scriptdir}"/*; do
+        [ -f "${script}" ] || continue
+        name="${script##*/}"
         info "warewulf: ${name}"
-        PREFIX=$PREFIX sh "${scriptdir}/${name}"
+        PREFIX=$PREFIX sh "${script}"
+        status=$?
+        if [ "${status}" -ne 0 ]; then
+            warn "warewulf: ${name} exited with status ${status}"
+            exit "${status}"
+        fi
     done
 else
     info "warewulf: ${scriptdir} not found"
```

The new `warn` fallback deliberately writes to stderr rather than following the existing `info` fallback onto stdout.

In a two-stage boot, a child process's stderr is journaled and reaches the console, whilst its stdout is not.

Please see the "Validation testing" section below for more evidence.

`load-wwinit.sh` gains the same guard the steps around it already have:

```diff
--- a/dracut/modules.d/90wwinit/load-wwinit.sh
+++ b/dracut/modules.d/90wwinit/load-wwinit.sh
@@ -62,7 +62,8 @@ mkdir /tmp/wwinit
     get_stage "system" || die "Unable to load stage: system"
 )
 if [ -x /tmp/wwinit/warewulf/run-wwinit.d ]; then
-        PREFIX=/tmp/wwinit /tmp/wwinit/warewulf/run-wwinit.d
+        PREFIX=/tmp/wwinit /tmp/wwinit/warewulf/run-wwinit.d \
+            || die "warewulf: a wwinit.d script failed"
 fi
 
 info "warewulf: mounting ${wwinit_root_device} at ${NEWROOT}"
```

As well as `init`, so the single-stage path also stops reporting nothing:

```diff
--- a/overlays/wwinit/rootfs/init
+++ b/overlays/wwinit/rootfs/init
@@ -33,7 +33,7 @@ else
 fi
 
 if [ -x /warewulf/run-wwinit.d ]; then
-        /warewulf/run-wwinit.d
+        /warewulf/run-wwinit.d || die "warewulf: a wwinit.d script failed"
 fi
 
 NEXT_INIT=/warewulf/run-init
```

The two `die`s do not mean the same thing, and this needs a decision - on the two-stage path, `die` is dracut's, so this gives `dracut: FATAL` and, under `rd.shell`, an emergency shell at the point of failure.

On the single-stage path, `init` supplies its own fallback (`init:9-17`):

```sh
if ! command -v die >/dev/null; then
    die() {
        printf '%s\n' "$*" >&2
        echo
        echo "Rebooting in 1 minute..."
        sleep 60
        echo b > /proc/sysrq-trigger || /sbin/reboot -f
    }
fi
```

On a single-stage node whose `wwinit.d` script fails, `|| die` produces a 60 second pause and a hard reboot, over and over.

That is a reboot loop, and I have measured it on a booting node rather than inferred it from the source.

It ran for 8 cycles with 7 completed reboots before I stopped the window, and the capture is in the validation section below.

My reading is that this is the existing convention for a fatal single-stage failure rather than something this change introduces, because `init` already calls the same `die` when `/warewulf/config` is missing - `init:32`, which is equally deterministic.

On that basis I have kept it, so that the two paths agree.

I am happy to change it to `warn` on the single-stage path only if the Warewulf Team would rather not add another reboot loop to that path.

The bulletpoint at `userdocs/overlays/overlays.rst:239-241` gains an additional sentence:

```diff
--- a/userdocs/overlays/overlays.rst
+++ b/userdocs/overlays/overlays.rst
@@ -238,7 +238,8 @@ provisioning actions:
 
 - **/warewulf/wwinit.d/:** executed in the initial root final system before the
   image is loaded into its final location. In a two-stage boot, these scripts
-  are executed in the Dracut initramfs.
+  are executed in the Dracut initramfs. A script which exits non-zero aborts
+  the boot, and the remaining scripts in the directory are not run.
 
 - **/warewulf/init.d/:** executed in the final root file system but before
   calling ``init``.
```

`CHANGELOG.md` gains an entry under `## Unreleased` / `### Fixed`, since this changes a behaviour:

```markdown
- A failing script in `/warewulf/wwinit.d/` now aborts the boot instead of being
  ignored. `run-wwinit.d` ran each script in a pipeline subshell without testing
  its exit status, so only the last script's status survived, and neither the
  two-stage `load-wwinit.sh` nor the single-stage `init` examined it. All four
  shipped `wwinit.d` scripts call `die` when their operation fails, so an
  Ignition, `sfdisk`, `mkfs` or `mkswap` failure was discarded and the boot
  continued as though the step had succeeded, either mounting a filesystem that
  may never have been created or, on a single-stage boot, coming up on a root
  file system built in RAM with the failure not treated as fatal.
```

## Validation testing

`run-wwinit.d` takes `PREFIX` from its environment and nothing else, so it can be run standalone.

I ran the current file from `main` at `bf910c8d` and the proposed file against the same synthetic `wwinit.d` directories, changing nothing else between the two runs.

The test was performed on Rocky Linux 9.8, kernel `5.14.0-687.26.1.el9_8.x86_64`, under three shells: `/bin/sh` bash 5.1.8, `dash` 0.5.11.5, and busybox 1.35 `ash`.

Each run used `env -i PATH=<dir containing only ls and sh>` so that the `info` and `warn` fallbacks in the script are the ones that run, as they are in an initramfs.

That restricted `PATH` is not just for show, `/usr/bin/info` exists on EL9 as the GNU texinfo reader, so with a normal `PATH` the fallback would be skipped and `info "warewulf: ..."` would invoke a pager.

| Case | `wwinit.d` contents | Before | After |
|---|---|---|---|
| Failure in a middle position | `10-ok.sh`, `20-fail.sh` (exit 7), `30-ok.sh` | exit **0**, and `30-ok.sh` ran anyway | exit **7**, stops at `20-fail.sh` |
| Failure in the last position | `10-ok.sh`, `90-fail.sh` (exit 7) | exit 7 | exit 7 |
| All scripts succeed | `10-ok.sh`, `20-ok.sh` | exit 0 | exit 0, output identical |
| Directory present but empty | none | exit 0 | exit 0, output identical |
| Directory absent | n/a | exit 0, `... not found` | exit 0, `... not found`, identical |
| Stray subdirectory beside a script | `10-ok.sh`, `adir/` | exit 126 | exit 0, subdirectory skipped |

Before, on the first case:

```
warewulf: running scripts in <PREFIX>/warewulf/wwinit.d...
warewulf: 10-ok.sh
10 ok
warewulf: 20-fail.sh
20 about to fail
warewulf: 30-ok.sh
30 ok
EXIT=0
```

After, on the same directory:

```
warewulf: running scripts in <PREFIX>/warewulf/wwinit.d...
warewulf: 10-ok.sh
10 ok
warewulf: 20-fail.sh
20 about to fail
warewulf: 20-fail.sh exited with status 7
EXIT=7
```

### On a booting node, two-stage, before and after

I again performed testing on a real Rocky Linux 9.8 node with x86_64 and with a real dracut initramfs, so that `die` is dracut's own rather than any fallback.

Kernel `5.14.0-687.33.1.el9_8`, `warewulf-4.7.0-1.el9` serving the overlays, the `90wwinit` dracut module and all overlays taken from `main` at `bf910c8d`.

The node in the reported case is aarch64, and I did not reproduce on that architecture, because everything this change touches is POSIX shell-related with no architecture-dependent behaviour.

The boot cmdline was read from the live rendered `/ipxe/<hwaddr>` endpoint, and iPXE was deliberately left out of the loop with a direct kernel boot so that only the initramfs and the overlay were under test.

The `wwinit.d` directory held two scripts: `50-fail.sh`, which prints a marker and exits 7, and `60-after.sh`, which prints a marker and exits 0.

The rootfs partition existed with the right `PARTLABEL` but carried no filesystem. Only one thing changed between the two boots.

Before, on `main` at `bf910c8d`:

```
dracut-pre-mount[520]: TESTMARK: 50-fail.sh starting (stderr)
dracut-pre-mount[520]: TESTMARK: command -v die -> NOTFOUND
dracut-pre-mount[522]: TESTMARK: 60-after.sh RAN
dracut-pre-mount[524]: mount: /sysroot: no filesystem type specified.
dracut: FATAL: warewulf: failed to mount /dev/disk/by-partlabel/rootfs at /sysroot
dracut: Refusing to continue
```

`50-fail.sh` exits 7, `60-after.sh` runs anyway, and the boot dies naming the mount.

Nothing anywhere names the script that failed. This is the reported defect.

Afterwards on the same node, with the same script and disk:

```
dracut-pre-mount[516]: TESTMARK: 50-fail.sh starting (stderr)
dracut-pre-mount[516]: TESTMARK: command -v die -> NOTFOUND
dracut-pre-mount[515]: warewulf: 50-fail.sh exited with status 7
///lib/dracut/hooks/pre-mount/30-load-wwinit.sh@66(source): die 'warewulf: a wwinit.d script failed'
dracut: FATAL: warewulf: a wwinit.d script failed
```

`60-after.sh` does not run, the mount is never attempted, and both the script name and the status appear on the console.

Under `rd.shell` the emergency shell opens here rather than at the mount.

Two incidental confirmations from those captures.

`command -v die -> NOTFOUND` is printed from inside a `wwinit.d` script, which settles that dracut's `die` really is out of scope there and the in-file fallback really is what runs.

`50-fail.sh` printed one marker to stdout and one to stderr, of which only the stderr marker appears, which is the same stream asymmetry described above, observed here on a fresh boot.

No regression when everything succeeds - with a `wwinit.d` whose scripts all exit 0 and a formatted rootfs, before and after were run back to back.

Both reached `warewulf: invoking /sbin/init...` with zero `dracut: FATAL` lines, having loaded the `image`, `system` and `runtime` stages.

Reducing both captures to their `warewulf:` and `TESTMARK` lines and diffing gives no differences at all.

### On a single-stage node, before and after

The same four cells were then run on the single-stage path, where `die` is `init`'s own fallback rather than dracut's.

Same host, `root: tmpfs`, `wwinit.d` holding the same `50-fail.sh` and `60-after.sh`, plus the stock `10-ignition.sh` that the `wwinit` overlay ships, which ran first in every boot below and exited 0 with `warewulf: ignition not found`.

The cmdline was again read from the live rendered `/ipxe/<hwaddr>` endpoint, this time the `:boot_single_stage` entry, and iPXE was again left out of the loop with a direct kernel boot.

Kernel banner counts are zero throughout because the node's real `KernelArgs` contain `quiet`, so cycles are counted from `init`'s own unconditional first line at `init:19`.

Before, on `main` at `bf910c8d`:

```
warewulf: 50-fail.sh
TESTMARK: 50-fail.sh starting (this line went to stdout)
TESTMARK: 50-fail.sh starting (this line went to stderr)
TESTMARK: command -v die -> NOTFOUND
warewulf: 60-after.sh
TESTMARK: 60-after.sh RAN, so the failure was discarded (this line went to stderr)
warewulf: configuring root file system...
warewulf: mounting tmpfs at /newroot
warewulf: switching to /newroot and invoking /warewulf/run-init...
```

and onward to `warewulf: invoking /sbin/init...` and a `login:` prompt.

This is worth stating separately from the two-stage case.

On two-stage the discarded failure at least kills the boot later, at the mount, naming the wrong step.

On single-stage with `root: tmpfs` there is no mount to fail, so nothing stops the boot at all and the node comes up on a RAM root file system as though provisioning had succeeded.

The failing script's own stderr is visible on the console here, so this is not an invisible failure so much as a non-fatal one, which is a different problem from the two-stage case and arguably a worse one.

Afterwards on the same node with the same script:

```
warewulf: 50-fail.sh
TESTMARK: 50-fail.sh starting (this line went to stdout)
TESTMARK: 50-fail.sh starting (this line went to stderr)
TESTMARK: command -v die -> NOTFOUND
warewulf: 50-fail.sh exited with status 7
warewulf: a wwinit.d script failed

Rebooting in 1 minute...
```

then `SeaBIOS version 1.16.3-5.el9` and the whole cycle again.

Inside a 480 second window: 8 init cycles, 8 `exited with status 7` lines, 8 `a wwinit.d script failed` lines and 8 `Rebooting in 1 minute...` lines, on a constant 27 line period.

The capture holds 8 `SeaBIOS` banners, of which the first is the initial start, so 7 reboots were observed completing into a following cycle, and the window elapsed during the eighth 60 second pause.

`60-after.sh` does not appear once, `configuring root file system` does not appear once, and the boot never reaches `/sbin/init`.

The reboot loop I described above is measured.

## Reviewer  checklist


mslacken (approved):

mslacken (changes requested):
Integration fails on rocky10. Unfortunately I don't see an error message.

metalllinux:
> Integration fails on rocky10. Unfortunately I don't see an error message.

Thank you kindly for letting me know @mslacken and I will get those changes made.

metalllinux:
I found the issue @mslacken and it was a defect in my change. On Rocky Linux 10, the virtio console is a module, so PID 1 has no usable stdout under `console=hvc0`. That makes the trailing `info` call in `10-ignition.sh` fail and my change then propagates that into a silent 60 second reboot loop.

I have modified the PR and pushed a fix that has the fallback log helpers `return 0`, has each `wwinit.d` script end in an explicit `exit 0`, and `console=ttyS0` added alongside `console=hvc0` in the integration test.

For testing, I verified this on a Rocky Linux 10.2 bare metal, where the test passes with CI's kernel arguments left untouched and still aborts correctly on a script that fails.

anderbubble:
Since this does change the behavior of error handling, I'm moving it to v4.8.0. (Otherwise, user scripts that exit nonzero but don't expect that to halt boot would have an unexpected change in behavior.)

PR #2237, opened 2026-08-06
fix: use NetTagsAdd instead of TagsAdd for --nettagadd in profile set

SergioZ3R0:
# Description
This PR fixes an issue where the `--nettagadd` flag fails to apply changes when used at the profile level via the CLI.

### Steps to Reproduce
Running the following commands currently does not save the tags to the profile:

```bash
wwctl profile set compute --nettagadd DNS1=1.1.1.1
wwctl profile set compute --nettagadd DNS2=1.0.0.1
wwctl profile set compute --nettagadd DNSSEARCH=hpc.customerdomain.com

```

### Expected Behavior
The tags should be applied to the profile properly, consistent with how it currently works when using:

* Interactive mode: `wwctl profile edit`
* Node level: `wwctl node set cpu001 --nettagadd DNS1=1.1.1.1`

### Fix

The refactoring in commit `4cc6d03f2f216a1fe9da932b38de06a5e8f1964d` incorrectly changed `NetTagsAdd` to `TagsAdd` when applying network tags in `profile set`.

```diff
- if len(vars.profileAdd.TagsAdd) > 0 && netDev.Tags == nil {
+ if len(vars.profileAdd.NetTagsAdd) > 0 && netDev.Tags == nil {

- for key, val := range vars.profileAdd.TagsAdd {
+ for key, val := range vars.profileAdd.NetTagsAdd {

```

The `--nettagadd` flags were stored in `NetTagsAdd` but the code read from `TagsAdd` (which is for `--tagadd`), causing `--nettagadd` to silently have no effect on profiles.

The `node set` equivalent code already used `NetTagsAdd` correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- `wwctl profile set --nettagadd` no longer silently ignores the given tags
+  when the network device already exists.
 - The two-stage dracut boot now completes on IPv6-only nodes. The `:dracut` entry
   rendered `ip=<device>:dhcp` for every device, which NetworkManager's initrd
   generator turns into a mandatory DHCPv4 lease, so the connection never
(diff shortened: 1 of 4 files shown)


mslacken (changes requested):
@SergioZ3R0 thanks for catching this and providing the fix. I have added a test, but I would like that you add yourself as contributor.
If you don't want this give me a hint and I will merge it.

SergioZ3R0:
@mslacken Thank you! I've added myself to the contributors file.

mslacken (approved):

mslacken:
Muchas gracias.

PR #2241, opened 2026-08-10
Fix IPMI address handling in power commands to skip nodes without an IP Address

ssimpson89:

`wwctl power` acted on the BMC of the Warewulf server itself when run against a node with no IPMI address. A node with no BMC correctly has no `ipmi: ipaddr:` and therefor is unaffected.

### Changes

- **Guard** (all six `power/*/main.go`): use the nil-safe idiom already in `wwctl node console` / `sensors`, `node.Ipmi == nil || node.Ipmi.Ipaddr == nil || node.Ipmi.Ipaddr.IsUnspecified()`. The node is skipped, and `returnErr` is now set so a run that skips nodes exits non-zero. Nodes with an address are processed as before.
- **Backstop** (`internal/pkg/bmc/bmc.go`): refuse any rendered `ipmitool` command with neither `-H` nor `-I`, so the safety property holds even for a site-edited template.

Reproduced end to end on a VM with a working local BMC (QEMU `ipmi-bmc-sim`): before, `wwctl power reset` against an address-less node reset the head node itself; after, it is refused and the host is untouched.

The code and tests were written with AI but validated manually

## Reviewer  checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- `wwctl power` (`cycle`, `off`, `on`, `reset`, `soft`, `status`) no longer runs
+  `ipmitool` without `-H` for a node that has no `ipmi: ipaddr:`, which fell back
+  to the local BMC and acted on the Warewulf server itself. Such nodes are now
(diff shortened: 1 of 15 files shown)


mslacken (approved):
Kind of an edge case as this can only happen, if the ipmi address is only specified the node itself, but legit.

PR #2267, opened 2026-09-10
Allow version strings to capture more information

snehring:

We're running some kernels now where the version string of two installed kernels would be identical to each other with how warewulf currently parses versions. This lead to me being slightly confused as to why the older kernel was being chosen as the default kernel for an image instead of the newer one. The differentiating detail is further down the kernel version string (eg past the el9_6 in the string). This is a quick first pass at attempting to allow warewulf to understand more of the kernel version string as the version.

Since we're using go-version to do some version validation this currently replaces `_` with `-` since the former is not valid in semver.

This also updates tests that were broken by this change to expect the now potentially expanded results.

We don't have to do this if there's strong objections. It's easy enough to work around by removing the older kernel when updating an image so the only kernel in the image is the one that should be booted.

## Reviewer  checklist

- [ x] The test suite has been updated, if necessary

### DIFF
--- a/internal/pkg/util/version.go
+++ b/internal/pkg/util/version.go
@@ -9,16 +9,19 @@ import (
 
 var (
 	versionPattern *regexp.Regexp
+	imageSuffixPattern *regexp.Regexp
 )
 
 func init() {
-	versionPattern = regexp.MustCompile(`\d+\.\d+\.\d+(-[\d\.]+|)`)
+	versionPattern = regexp.MustCompile(`\d+\.\d+\.\d+(-[\w+\.-]+)?`)
+	imageSuffixPattern = regexp.MustCompile(`\.gz|\.bz2|\.xz|\.lz4|\.zst|\.lzma|\.lzo`)
 }
 
 func ParseVersion(versionString string) *version.Version {
 	matches := versionPattern.FindAllString(versionString, -1)
 	for _, match := range matches {
-		if version_, err := version.NewVersion(strings.TrimSuffix(match, ".")); err == nil {
+		match = imageSuffixPattern.ReplaceAllString(match, "")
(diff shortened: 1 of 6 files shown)


internal/pkg/util/version.go
```diff
@@ -8,17 +8,20 @@ import (
...
+	versionPattern = regexp.MustCompile(`\d+\.\d+\.\d+(-[\w+\.-]+)?`)
+	imageSuffixPattern = regexp.MustCompile(`\.gz|\.bz2|\.xz|\.lz4|\.zst|\.lzma|\.lzo`)
 }
 
 func ParseVersion(versionString string) *version.Version {
 	matches := versionPattern.FindAllString(versionString, -1)
 	for _, match := range matches {
-		if version_, err := version.NewVersion(strings.TrimSuffix(match, ".")); err == nil {
+		match = imageSuffixPattern.ReplaceAllString(match, "")
+		if version_, err := version.NewVersion(strings.ReplaceAll(strings.TrimSuffix(match, "."), "_", "-")); err == nil {
```
anderbubble, line 24: I'm concerned with going too far down the path of looking in non-standard locations for version strings. For example, in `6.12.0-211.16.1.el10-2.0.1.x86-64`, the `2.0.1` isn't really part of the kernel version, but is more natually associated with `el10`.

anderbubble (changes requested):
Comment attached inline.

I understand that's not what it's doing in your case, but site-local exceptions are why we also support specifying the desired kernel by explicit path.

internal/pkg/util/version.go
snehring: I think that's a valid concern. Even if we don't decide to capture the entire string, I do think the current regex of only allowing numbers for the kernel EXTRAVERSION is a bit restrictive.

